PR #135, opened 2014-07-03
Allow pushing config settings for the SourceInformationProvider

Tofudebeast:


onovotny:
@Tofudebeast What is the intent here? What problem is this trying to address/fix?

There's been a few updates to the `VisualStudioSourceInformationProvider` recently. As the changes appear small (aside from the formatting issues...), would you be able to create a new branch with just the key changes?

JeremyKuhne:
@onovotny Using a config file (say, to impact assembly loading) wasn't working because the domain that was being fired up didn't actually use it. I'll take a look at the current code and let you know.

PR #235, opened 2014-12-14
Added support for synchronous test discovery message reporting.

remcomulder:
This commit adds support for the use of a synchronous message bus when reporting discovered tests to a test runner tool.  NCrunch unfortunately requires test discovery to be a synchronous process as it is underpinned by a number of mechanisms (such a metadata discovery) that don't handle multithreading particularly well.


bradwilson:
Merged. Thanks!

PR #271, opened 2015-01-27
Do not write into StdErr in AppVeyor log visitor

FeodorFitsner:


bradwilson:
This removes all Console.Error, but it should really only remove the use of Console.Error when a test is skipped. Test failures should still be written to Console.Error.

I'm going to push a fix for this that also fixes the standard output visitor, so this PR shouldn't be necessary as my work will subsume it.

Thank you!

FeodorFitsner:
Well, fair enough, thanks!

FeodorFitsner:
Please let me know when there is a new build on nuget.org so I can updated it on build workers.

PR #273, opened 2015-01-28
Fixing issue with TeamCity9 where it does not properly recognize test co...

aweiker:
...unts when tests have similar names

### DIFF
--- a/src/xunit.runner.msbuild/Visitors/TeamCityDisplayNameFormatter.cs
+++ b/src/xunit.runner.msbuild/Visitors/TeamCityDisplayNameFormatter.cs
@@ -1,0 +1,27 @@
+using System;
+using System.Collections.Concurrent;
+using System.Threading;
+using Xunit.Abstractions;
+
+namespace Xunit.Runner.MSBuild
+{
+    public class TeamCityDisplayNameFormatter
+    {
+        readonly ConcurrentDictionary<string, int> assemblyMappings = new ConcurrentDictionary<string, int>();
+        int assemblyCount = 0;
+
+        public virtual string DisplayName(ITestCollection testCollection)
+        {
+            var id = assemblyMappings.GetOrAdd(
+                testCollection.TestAssembly.Assembly.Name,
+                key => Interlocked.Increment(ref assemblyCount));
+
+            return String.Concat(testCollection.DisplayName, " (", id, ")");
+        }
+
+        public virtual string DisplayName(ITest test)
+        {
+            return test.DisplayName;
+        }
+    }
+}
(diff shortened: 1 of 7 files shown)


bradwilson:
Merged, thanks!

valmaev:
Actually, there are still cases when TeamCity will show less tests than xUnit.net reports.
**TLDR**: if your test parameter is of enum type which have multiple members with same underlying value (e.g. [System.Net.HttpStatusCode](https://referencesource.microsoft.com/#System/net/System/Net/HttpStatusCode.cs,70)).

Originally reported to TeamCity issue tracker ([TW-49660](https://youtrack.jetbrains.com/issue/TW-49660#comment=27-2117365))

Consider following example:
```csharp
[Theory, MemberData(nameof(NonSuccessfulStatusCodes))]
public async Task AwesomeRestServiceClient_WithFailedResponse_ShouldThrow(
    HttpStatusCode responseStatusCode)
{
    // Test logic
}

public static IEnumerable<object[]> NonSuccessfulStatusCodes()
{
    return Enum.GetValues(typeof(HttpStatusCode))
        .Cast<int>()
        .Where(c => c < 200 || c > 299)
        .Select(c => new object[] {c});
}
```

xUnit.net runner will report about multiple tests with following names: 
```
AwesomeRestServiceClient_WithFailedResponse_ShouldThrow(responseStatusCode: 300)
```

*Important thing №1*: xUnit.net uses enum underlying value, but not enum member name.

OK, but so what? 

*Important thing №2*: If we will dig (not so) deep into .NET Framework, we will see that [System.Net.HttpStatusCode](https://referencesource.microsoft.com/#System/net/System/Net/HttpStatusCode.cs,70) enum have *different enum members with same underlying value*:
```csharp
public enum HttpStatusCode
{
    Ambiguous = 300,
    MultipleChoices = 300,

    Moved = 301,
    MovedPermanently = 301,

    Found = 302,
    Redirect = 302,

    RedirectMethod = 303,
    SeeOther = 303,
    
    RedirectKeepVerb = 307,
    TemporaryRedirect = 307
}
```
So, this particular issue can be solved by modifying xUnit.net runner so it will return:
```
AwesomeRestServiceClient_WithFailedResponse_ShouldThrow(responseStatusCode: Ambiguous)
AwesomeRestServiceClient_WithFailedResponse_ShouldThrow(responseStatusCode: MultipleChoices)
```
or 
```
AwesomeRestServiceClient_WithFailedResponse_ShouldThrow(responseStatusCode: Ambiguous(300))
AwesomeRestServiceClient_WithFailedResponse_ShouldThrow(responseStatusCode: MultipleChoices(300))
```
instead of
```
AwesomeRestServiceClient_WithFailedResponse_ShouldThrow(responseStatusCode: 300)
AwesomeRestServiceClient_WithFailedResponse_ShouldThrow(responseStatusCode: 300)
```

Or we can just fix the `MemberData` source by excluding duplication:
```csharp
public static IEnumerable<object[]> NonSuccessfulStatusCodes()
{
    return Enum.GetValues(typeof(HttpStatusCode))
        .Cast<int>()
        .Where(c => c < 200 || c > 299)
        .Distinct() // this will solve issue
        .Select(c => new object[] {c});
}
```

PR #288, opened 2015-02-05
Split base functionality of MemberDataAttribute into base class

wyand:
This change splits `MemberDataAttribute` into an abstract `MemberDataAttributeBase` that contains most functionality and a sealed `MemberDataAttribute` class implementing the existing `MemberDataAttribute` functionality.

The goal of this is to make it easier to add custom `MemberDataAttribute` extensions, that permit data-yielding members to return any `IEnumerable<object>` type instead of requiring `IEnumerable<object[]>`. The `MemberDataAttributeBase` class has a new `protected abstract ConvertDataItem` method that will be used in `GetData` to convert the resulting `object`s into `object[]`s.

Since this changeset involves a rename, it will be easier to review the changes by just looking at commit 6d73baeeadfdd7a393b2e110ad8482f74ef9dcf4 instead of the overall changes including the rename.

----------------

The idea that led me to this is implementing a custom `DataAttribute` that will map properties in data objects to method parameters by name, instead of requiring the array indices to match. For example:

```cs
public static IEnumerable<object> TheoryData()
{
    yield return new { a = "A1", b = "B1", c = "C1" };
    yield return new { a = "A2", b = "B2", c = "C2" };
}

[Theory, NamedMemberData("TheoryData")]
public void Theory1(string a, string b)
{ ... }

[Theory, NamedMemberData("TheoryData")]
public void Theory2(string a, string b, string c)
{ ... }
```

While this is certainly possible without this change, it requires duplicating a lot of the code in `[MemberData]`. With the change, only the `object` to `object[]` conversion needs to be written to implement `[NamedMemberData]`.

I'd also be happy to add another pull request with `[NamedMemberData]` implemented.


bradwilson:
We would like to accept this PR. Would you be okay with signing a contribution agreement with Outercurve?

wyand:
@bradwilson Happy to. Do I just fill in [this pdf](http://www.outercurve.org/assets/files/OutercurveFoundationContributionAgreement-editable.pdf), and is a reference to the pull request sufficient documentation for appendix A?

bradwilson:
Send me an e-mail at dotnetguy@gmail.com with your name and desired e-mail address, and I'll start the process with Outercurve.

bradwilson:
Merged, thanks!

wyand:
:thumbsup:

PR #297, opened 2015-02-13
Add InvariantCulture to xml output formatting.

alexperovich:
When tests are running asynchronously and some of them change the thread culture during the test this results in number format issues in the xml document. Some of the numbers are formatted with periods as the decimal point and others have commas.
The fix is to add CultureInfo.InvariantCulture to all the ToString calls.


bradwilson:
Awesome, thanks! We'll merge this before RC2.

PR #306, opened 2015-02-19
Restore NUnit XML output

coldacid:
There are some of us out here who still are stuck with CI platforms that expect/require NUnit or JUnit output for parsing test results. The original NUnit XSL transform was removed in the 1.9->2.0 transition, but these changes restore the ability to get NUnit style XML results including a minor update to the transform to work with the xUnit.net 2.0 XML.


bradwilson:
Merged, thanks!

PR #308, opened 2015-02-20
Compilation on monodevelop?

txdv:
I was trying to compile xunit with monodevelop (I'm on linux) and I am running into issues.

First issue was rather simple to solve, change the case of a the snk, but I run into errors which I don't 'get' immediately: https://gist.github.com/595bcafe5ec3b772e066


txdv:
Nevermind, https://github.com/akoeplinger/xunit/tree/xbuild is ahead, I will pressure him to make a pull request!

PR #310, opened 2015-02-20
Fix signing.snk file casing

akoeplinger:
The file is lower-cased on disk, but was used with upper-case S in some projects.


bradwilson:
Merged, thanks!

PR #324, opened 2015-02-23
Fixes #322: Can't generate report in nunit format when using multiple assemblies in xunit.console

### LINKED ISSUE #322: Can't generate report in nunit format when using multiple assemblies in xunit.console
Hi,

If I add more than one assembly to the xunit.console, the tests are executed properly, however if I would like to get the result in nunit format then I recieve an error: 

*System.InvalidOperationException: Token StartElement in state EndRootElement would result in an invalid XML document. Make sure that the ConformanceLevel setting is set to ConformanceLevel.Fragment or ConformanceLevel.Auto if you want to write an XML fragment.*

It seems that the provided NUnitXml.xslt would like to add two root (test-results) nodes, and the transformed would look something like this:

`<?xml version="1.0" encoding="UTF-8"?>
<test-results name="First.Test.dll" date="2015-02-23" time="17:12:12" total="1" failures="0" not-run="0">
  <test-suite name="First.Test.dll" success="True" time="0.290">
    <results>
      <test-suite name="xUnit.net v1 Tests for First.Test.dll" success="True" time="0.290">
        <results>
          <test-case name="First.Test.Test.TestMethod" executed="True" success="True" time="0.290"/>
        </results>
      </test-suite>
    </results>
  </test-suite>
</test-results>`

`<test-results name="Second.Test.dll" date="2015-02-23" time="17:12:12" total="1" failures="0" not-run="0">
  <test-suite name="Second.Test.dll" success="True" time="0.289">
    <results>
      <test-suite name="xUnit.net v1 Tests for Second.Test.dll" success="True" time="0.289">
        <results>
          <test-case name="Second.Test.dll" executed="True" success="True" time="0.289"/>
        </results>
      </test-suite>
    </results>
  </test-suite>
</test-results>`

Regards,
Attila

coldacid:
This change fixes the NUnit XML output when two or more assemblies are included in a test run. It's a bit more complicated than one would think at first due to the requirement of a single `<test-suite>` tag directly under `<test-results>`, requiring a bit more layering to ensure correct, valid output.

### DIFF
--- a/src/xunit.console/NUnitXml.xslt
+++ b/src/xunit.console/NUnitXml.xslt
@@ -6,42 +6,54 @@
     <xsl:apply-templates/>
   </xsl:template>
 
-  <xsl:template match="assembly">
-    <test-results>
-      <xsl:attribute name="name">
-        <xsl:value-of select="@name"/>
-      </xsl:attribute>
+  <xsl:template match="assemblies">
+    <test-results name="Test results">
       <xsl:attribute name="date">
-        <xsl:value-of select="@run-date"/>
+        <xsl:value-of select="assembly[1]/@run-date"/>
       </xsl:attribute>
       <xsl:attribute name="time">
-        <xsl:value-of select="@run-time"/>
+        <xsl:value-of select="assembly[1]/@run-time"/>
       </xsl:attribute>
       <xsl:attribute name="total">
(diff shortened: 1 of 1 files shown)


bradwilson:
Merged, thanks!

PR #335, opened 2015-03-03
Add tdnet project back to xunit.xbuild.sln

akoeplinger:
As discussed in https://github.com/xunit/xunit/commit/a78fe639f622643e749aaeca615d6203ca4d963c this allows to remove the `!__MonoCS__` hack and reduces a few warnings when building with xbuild (even though it makes little sense to build TD.NET on Mono).


bradwilson:
Merged.

PR #339, opened 2015-03-05
Changing System.XML to System.Xml so that it works xplat

sayedihashimi:
On OSX we are running into errors like ```Unable to resolve dependency framework/System.XML```. I think this is because the casing is not correct. The change here just update the .csproj files which had ```System.XML``` to ```System.Xml```. I'm not sure if there are other issues but wanted to get this PR to you for review.

### DIFF
--- a/src/xunit.console/xunit.console.csproj
+++ b/src/xunit.console/xunit.console.csproj
@@ -55,7 +55,7 @@
     <Reference Include="System.configuration" />
     <Reference Include="System.Runtime" />
     <Reference Include="System.Web.Extensions" />
-    <Reference Include="System.XML" />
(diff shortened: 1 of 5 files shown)


bradwilson:
Have you tried this fix on Linux with Mono? The existing casing works there, and I would like to ensure that fixing OS X doesn't, in turn, break Linux.

sayedihashimi:
@bradwilson I have not yet tried on linux

bradwilson:
I was able to verify that it continues to build on Ubuntu 14 with Mono 3.12.

bradwilson:
Merged, thanks!

PR #363, opened 2015-03-20
Since the class isn't sealed, it can be inherited, so make serialization virtual

kzu:
Custom frameworks, runners, etc. may want to create custom collections for their own parallelism (or whatever) purposes, and may need to participate in the serialization too. 

In particular, I would use this for my own VSSDK extension that runs in parallel VSX integration tests across all installed versions of VS by creating a custom test collection for each IDE (which is custom-run).


bradwilson:
Merged, thanks!

PR #383, opened 2015-04-05
Dispose Enumerators if possible

roji:
In some cases it is illegal to have multiple enumerators open on the same enumerable at the same time. This is the case with EntityFramework results which come from a database that doesn't support concurrent DbDataReaders (e.g. PostgreSQL).

When comparing enumerables, added a check if the enumerators are disposable, and if so, dispose them.


roji:
Additional note to describe where things went wrong for me: when comparing two enumerables, if there's a discrepancy an assertion exception is constructed, which attempts to format the enumerators as strings. Since the first enumerator (the one used for the comparison) is still open, the second enumerator cannot be created.

bradwilson:
Merged, thanks!

PR #390, opened 2015-04-10
Allow AspNet assemblies to load correctly when treated as platform

FiveTimesTheFun:
Previously, xunit assemblies were loaded by simple name, using
reflection with an AssemblyName without a version set.  This works
when the assembly is not treated as a trusted platform assembly.
When the assembly is treated as a trusted platform assembly, the bind
will fail with FUSION_E_REF_DEF_MISMATCH since binds to platform
assemblies must include a version.  By adding Version(0, 0) to the
AssemblyName, the bind will succeed because the assembly def version
will be greater than the assembly ref version.

See https://github.com/dotnet/coreclr/blob/master/src/binder/assemblybinder.cpp#L113 - prior to this change, this would succeed when fBeingBoundToPlatformAssembly was false, but fail when true.

### DIFF
--- a/src/common/SerializationHelper.cs
+++ b/src/common/SerializationHelper.cs
@@ -159,7 +159,7 @@ namespace Xunit.Sdk
             {
                 // Make sure we only use the short form for WPA81
                 var an = new AssemblyName(assemblyName);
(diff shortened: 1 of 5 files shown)


bradwilson:
Merged. Thanks!

PR #396, opened 2015-04-19
Workaround Xamarin linker issues

onovotny:
This PR ensures that the correct/matching execution libs are pulled into Xamarin.iOS & Android projects.

The runner app template itself (not the runner dll/nuget) has the following code `AddExecutionAssembly(typeof(ExtensibilityPointFactory).Assembly);` whose sole purpose is to ensure the Xamarin linker/packager includes the execution library.

Right now as a workaround the devices nuspec pulls in the `xunit.extensibility.execution` package for ios/android but that is incorrect as the version may not correctly match the version of xunit.core.


bradwilson:
Merged. Thanks!

PR #405, opened 2015-04-26
Ensure that win81 uses the universal versions

onovotny:


bradwilson:
Opened a discussion (#406) because I think we might just want to drop `win8` rather than perform more NuGet gymnastics.

Issue #366, opened 2015-03-23
xunit.console.exe fails on tests decorated with both [Fact] and [Theory] when traits filer is applied


ewgenym:
Steps:
1. Test class looks like this
```
    [Xunit.Trait("Category", "Smoke")]
    public class TestClass
    {
        [Xunit.Theory()]
        [Xunit.Fact()]
        public void should_do_the_magic(string value)
        {
            Xunit.Assert.True(true);
        }
    }
```
2. run xunit console like this;
```
xunit.console.exe TestXunit.dll -trait Category=Smoke
```

3. Actual result:
```
System.InvalidOperationException: Sequence contains more than one element
```

Expected is more meaningful error message that states that [Fact] and [Theory] are not allowed together on the same method with test name included.

PR #410, opened 2015-04-30
Add specific error for method with [Fact][Theory]

jhamm:
Fix for #366


jhamm:
An InvalidOperationException is thrown with this error message:

Only one [Fact] attribute or its derived type is allowed on a method. Method: TestClass.should_do_the_magic [Fact] Attributes: Xunit.FactAttribute, Xunit.TheoryAttribute

bradwilson:
Oops, I just did an implementation for this. Should've looked here first. :-p Thanks!

Issue #367, opened 2015-03-23
Add runner switch to treat skips as failures instead of warnings


bradwilson:
Skips should be something you do temporarily. It's fine to use them during development, but it would be nice to enforce "no skips in source control" by using runner switches for CI builds.

bradwilson:
Fixed by #411.

PR #411, opened 2015-05-03
add -noskips switch to console runner

jhamm:
When flag is set and skips exist, it will report the skips and add that
number to the failures.
For the all outputs, the failure reason will be FAIL_SKIP to indicate why
the test failed.

PR for #367


bradwilson:
Thanks for this PR.

There were a couple things I ended up not 100% sure on when I evaluated it; the biggest one is that tests are double counted (they are counted as a skip as well as a failure). I'm thinking the best solution might be an intercepting message bus that converts skips to failures so that the rest of the pipeline does not need to be changed. It's something I'd want to play around with to see if I like it, but I don't have time to do that today so this is going to have to wait for the next beta.

jhamm:
I understand your comment of the double counting. This was done on purpose because the test falls in both categories when the flag is set, but I can see how this could be misleading.

I can take a look at using the Message Bus approach to see if that is better.

jhamm:
I update the PR based on your suggestions. You are right the first commit didn't follow SRP. I think this one is better.

bradwilson:
Sorry it took me so long to get to this. I had to do some surgery on it to get it to compile because of the code changes that have happened in the intervening time.

I renamed the console switch to `-failskips` rather than `-noskips` because that seemed to better reflect its purpose. I also added the functionality into the MSBuild runner, triggered by adding `FailSkips="true"` to the `<xunit>` task.

Everything has been merged with commit 9bee623.

Thanks again!

Issue #425, opened 2015-05-27
Xunit is converting arrays to object[] in theories


Galad:
Hello,
I found an issue when using a Theory.
Those 2 tests fails :

    [Theory,
    InlineData(typeof(int[]), new int[]{ 1, 2, 3}),
    InlineData(typeof(string[]), new string[] { "a", "b", "c" })    
    ]
    public void TheoryShouldRetrieveCorrectType(Type type, object value)
    {
        Assert.IsType(type, value);
    }

What is happening is that the type of `value` is `object[]` instead of `int[]`.
I believe that we should retrieve the exact type that was passed to the InlideData attribute.

The issue does not occur with MemberData.

PR #427, opened 2015-05-28
Fixes issue #425

Galad:
This pull request fixes issue #425.

The issue came from the values of the `InlineDataAttribute` not being converted to the right type if the destination type defined in the test method is object.


bradwilson:
Merged, thanks!

PR #432, opened 2015-06-07
Use full version number to prevent issues runing on Wpa81

onovotny:
This has not been tested yet, but this should address the error we're seeing:

Could not load file or assembly 'v2_WPA81, Version=0.0.65535.65535, Culture=neutral, PublicKeyToken=null' 

The v2_WPA81.dll has a default version of 0.0.0.0 and the load is asking for a higher version.


bradwilson:
Merged, thanks!

PR #436, opened 2015-06-16
build: Enables FreeBSD support

jasonwilliams200OK:
Changed hardcoded bash shebang to env to support multiple directory
structures (required to build on FreeBSD).


bradwilson:
Merged, thanks!

Issue #319, opened 2015-02-22
Add support in runner.utility for saying "no app domains"


bradwilson:
With the Mono AppDomain bug garnering no comments, it's probably worth investigating a way to avoid the Mono problem entirely by just skipping AppDomains. They will have to accept the loss of functionality associated with that (no isolation, file locking on some file systems, etc.).

adamralph:
:+1: 

(For reference, this is a workaround for https://github.com/xunit/xunit/issues/158.)

bradwilson:
Shipping in 2.1 beta 4.

adamralph:
For reference https://github.com/xunit/xunit/blob/cb4e9fb25015e73957a586b5e7b869b710ea70e0/src/xunit.console/Program.cs#L165

PR #441, opened 2015-06-20
Improve xunit perf, in particular for Theory test case discovery

stephentoub:
Over at https://github.com/dotnet/corefx, we're starting to use [Theory]ies more aggressively, and we're contemplating some usage that could result in tens of thousands of entries for a given test suite.  Today, this results in some significant performance penalties.

For example, given this test with 40,000 entries:
```C#
public static IEnumerable<object[]> LotsOfData
{
    get
    {
        for (int i = 0; i < 40000; i++)
            yield return new object[] { i, "This is a some data.", "And this is some more data." };
    }
}

[Theory]
[MemberData("LotsOfData")]
public static void TheoryTestWithLotsOfData(int i, string data, string moreData)
{
}
```
on my machine it takes xunit ~30 seconds to discover and run the tests, spending ~21 seconds on discovery and then ~9 seconds on test execution.

I spent some time iteratively profiling for both CPU usage and memory allocations and eliminating the top offenders, mostly low-hanging fruit.  With the commits in this PR, the above example on my machine has discovery time drop from ~21 seconds to ~0.25 seconds, and test execution time drop from ~9 seconds to ~8 seconds, for a total reduction from ~30 seconds to ~8 seconds.

The individual commits have more detail on the exact changes made.

This PR makes a sizeable dent in allocations, but there's still lots of work that can be done to improve things further.  Just as an example, here's a table showing the top-allocated objects, and how many times per test they're being allocated for my earlier example (so multiply each number by 40,000 to get the appx number of allocations for all of the tests together).  I've only shown types of objects allocated 15 or more times per test:

| Type | Number Before | Number After |
| --- | --- | --- |
| String |  278  | 51 |
| Func`2 | 220 | 0  |
| RuntimeMethodInfoStub | 68 | 8  |
| IntPtr | 61 |  1 |
| String[] | 54 |  8 |
| Char[] | 43 | 0  |
| List`1 |  39  | 22 |
| Object[] |  39 | 15 |
| Int[] | 36 | 2  |
| Stack`1 | 23 | 1 |
| CustomAttributeRecord[] | 20  | 14 |
| XunitSerializationTriple | 20 |  0 |
| Attribute[] | 19  |  9 |
| Byte[] | 18 |  0 |
| StringBuilder | 16 | 0  |
| SZGenericArrayEnumerator | 15 |  5 |

There's a very long tail of object types that get allocated twice or once per test as well, but the list was too long to type out here :smile:.

### DIFF
--- a/src/xunit.execution/Sdk/Utility/CommonTasks.cs
+++ b/src/xunit.execution/Sdk/Utility/CommonTasks.cs
@@ -1,0 +1,9 @@
+﻿using System.Threading.Tasks;
+
+namespace Xunit.Sdk
+{
+    internal static class CommonTasks
+    {
+        internal static readonly Task Completed = Task.FromResult(0);
+    }
+}
--- a/src/common/SerializationHelper.cs
+++ b/src/common/SerializationHelper.cs
@@ -64,6 +64,14 @@ namespace Xunit.Sdk
             return string.Format("{0}:{1}", GetTypeNameForSerialization(value.GetType()), serializationInfo.ToSerializedString());
         }
 
+        /// <summary>Gets whether the specified <paramref name="value"/> is serializable with <see cref="Serialize"/>.</summary>
+        /// <param name="value">The object to test for serializability.</param>
+        /// <returns>true if the object can be serialized; otherwise, false.</returns>
+        internal static bool IsSerializable(object value)
+        {
+            return XunitSerializationInfo.CanSerializeObject(value);
+        }
+
         /// <summary>
         /// Converts an assembly qualified type name into a <see cref="Type"/> object.
         /// </summary>
--- a/src/common/SerializationHelper.cs
+++ b/src/common/SerializationHelper.cs
@@ -107,8 +107,8 @@ namespace Xunit.Sdk
                             return null;
 
                         var genericArgument = assemblyQualifiedTypeName.Substring(firstOpenSquare + 1, lastOpenSquare - firstOpenSquare - 2);  // Strip surrounding [ and ]
-                        var innerTypeNames = SplitAtOuterCommas(genericArgument).Select(x => x.Substring(1, x.Length - 2)).ToArray();  // Strip surrounding [ and ] from each type name
-                        var innerTypes = innerTypeNames.Select(GetType).ToArray();
+                        var innerTypeNames = SplitAtOuterCommas(genericArgument).Select(x => x.Substring(1, x.Length - 2));  // Strip surrounding [ and ] from each type name
+                        var innerTypes = innerTypeNames.Select(s => GetType(s)).ToArray();
                         if (innerTypes.Any(t => t == null))
                             return null;
 
@@ -140,14 +140,11 @@ namespace Xunit.Sdk
                 }
             }
 
-            var parts = SplitAtOuterCommas(assemblyQualifiedTypeName).Select(x => x.Trim()).ToList();
-            if (parts.Count == 0)
-                return null;
-
-            if (parts.Count == 1)
-                return Type.GetType(parts[0]);
-
-            return GetType(parts[1], parts[0]);
+            IList<string> parts = SplitAtOuterCommas(assemblyQualifiedTypeName, trimWhitespace: true);
+            return 
+                parts.Count == 0 ? null :
+                parts.Count == 1 ? Type.GetType(parts[0]) :
+                GetType(parts[1], parts[0]);
         }
 
         /// <summary>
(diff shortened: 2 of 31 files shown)


stephentoub:
ps I should mention that this was all done using the *.dnx assemblies, where AppDomains aren't used.  When looking at perf using the *.desktop assemblies, the biggest remaining cost by far appears to come from the use of .NET Remoting, resulting in huge numbers of allocations as well.

sharwell:
I've been just starting to look into this due to growing performance concerns for the StyleCopAnalyzers project.

:question: Can you compare the before/after of this on the discovery and execution of the StyleCopAnalyzers project?

:question: Have you tested the performance inside the IDE or the command line? Does this change result in similar performance benefits for both cases?

src/xunit.execution/Sdk/Utility/CommonTasks.cs
```diff
@@ -0,0 +1,9 @@
+﻿using System.Threading.Tasks;
+
+namespace Xunit.Sdk
+{
+    internal static class CommonTasks
+    {
+        internal static readonly Task Completed = Task.FromResult(0);
```
sharwell: :bulb: Consider using `VoidResult` (declaring `VoidResult` as a private nested type).

https://github.com/rackerlabs/dotnet-threading/blob/master/Rackspace.Threading/VoidResult.cs

```csharp
internal static readonly Task Completed = Task.FromResult(default(VoidResult));
```

bradwilson:
Thanks for the PR. We'll evaluate it when we get together next.

stephentoub:
> Thanks for the PR. We'll evaluate it when we get together next.

Thanks, Brad.

> Can you compare the before/after of this on the discovery and execution of the StyleCopAnalyzers project?

I ran the tests with the changes, and it didn't really make a measurable difference, maybe a few seconds (out of the two minutes it took to run the tests).  But, as I mentioned in my "p.s." above, with the desktop bits a lot of time is being spent in .NET Remoting-related efforts, and these StyleCopAnalyzers tests are using the desktop bits.  I just profiled the tests briefly, and during the discovery phase at least 80% of the time and 50% of the allocations are spent related to the .NET Remoting stack.

bradwilson:
@stephentoub @sharwell Theoretically, once this is available, it should give better performance in desktop at the expense of locking the DLLs while the tests run (i.e., appropriate for command-line execution, but not for in-editor execution): https://github.com/xunit/xunit/issues/319

sharwell:
@stephentoub Thank you. That confirms the results of the testing I performed locally as well. Near the top of my list to experiment with is using a WCF service instead of remoting for the communication layer. For me it's very *very* early stage though.

bradwilson:
@stephentoub BTW, I feel like there's something else I should mention.

If you're really in a situation where you are returning 40k data rows, then pre-enumeration really doesn't make sense. Aside from being stupidly expensive, it wouldn't be realistic to display them individually to let you run them individually. You can use the `DisableDiscoveryEnumeration` property of `[MemberData]` to turn off pre-enumeration in that scenario.

https://github.com/xunit/xunit/blob/3084c9b60a1207be6c30606441956bb7a99757f6/src/xunit.core/MemberDataAttributeBase.cs#L34

stephentoub:
Thanks, @bradwilson.  I think the changes in this PR are goodness regardless, but it definitely sounds like we should also be turning off pre-enumeration in general.  Is there a flag that can be specified to do so for all [Theory]s, rather than needing to do it for individual uses of [MemberData]?

bradwilson:
@stephentoub I agree that the PR is generally useful. We'll definitely be looking at it soon. :)

Today, there is a global flag to turn off pre-enumeration via configuration (see http://xunit.github.io/docs/configuration-files.html); however, you're using DNX, where configuration files aren't yet supported (since the initial version was built on `App.config` files, a concept which DNX does not have).

I've been contemplating how to come up with sensible defaults for this stuff. I have a few notions:

* The most obvious: get a configuration story for DNX. :)
* Pre-enumeration is a UI-centric feature, so runners which just run tests without letting users choose them from a UI should probably never pre-enumerate.
* Pre-enumeration should probably stop once a particular test hits a cap of data items. (25? 50? 100?)

Thoughts on these or other strategies?

stephentoub:
Thanks, Brad.  Those strategies make sense to me, in particular only pre-enumerating when using a relevant runner (in which case it seems like the default should really be to not pre-enumerate and particular runners can opt to turn it on?)  I don't have a good sense for the cap... seems like you'll probably hit cases where no matter what value you choose, someone will want a higher cap, so you'd probably need it to be configurable.

bradwilson:
Thanks.

I squashed and rebased in 82e719db5dfa7e31ce31b4b169065e147dea118d. Also, I turned off theory pre-enumeration for the Console and MSBuild runners in 07178782a6ea455ae42f1fa6f4b3bfec935cd587, which should help a little bit more. :)

stephentoub:
Terrific! Thanks, Brad.

PR #577, opened 2015-09-03
Added namespace filtering to console runner

Jarga:
I was hoping to have the ability to filter runnable scripts by namespace through the console runner.  This will provide an easy way of building a script hierarchy without the need to add multiple traits to scripts. Provides a similar functionality to the "File Path" filter that is available in the Visual Studio Test Explorer.

### DIFF
--- a/src/xunit.runner.utility/Project/XunitFilters.cs
+++ b/src/xunit.runner.utility/Project/XunitFilters.cs
@@ -1,5 +1,6 @@
 ﻿using System;
 using System.Collections.Generic;
+using System.Linq;
 using Xunit.Abstractions;
 
 namespace Xunit
@@ -18,6 +19,7 @@ namespace Xunit
             IncludedTraits = new Dictionary<string, List<string>>(StringComparer.OrdinalIgnoreCase);
             IncludedClasses = new HashSet<string>(StringComparer.OrdinalIgnoreCase);
             IncludedMethods = new HashSet<string>(StringComparer.OrdinalIgnoreCase);
+            IncludedNameSpaces = new HashSet<string>(StringComparer.OrdinalIgnoreCase);
         }
 
         /// <summary>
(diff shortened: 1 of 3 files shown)


src/xunit.runner.utility/Project/XunitFilters.cs
```diff
@@ -53,9 +60,22 @@ public bool Filter(ITestCase testCase)
...
 
             return true;
         }
+        bool FilterIncludedNameSpaces(ITestCase testCase)
+        {
+            // No assemblies in the filter == everything is okay
+            if (IncludedNameSpaces.Count == 0)
+                return true;
+
+            if (IncludedNameSpaces.Count != 0 && IncludedNameSpaces.Any(a => testCase.TestMethod.TestClass.Class.Name.Contains(a)))
```
bradwilson: I'm not quite sure how I feel about this as an implementation for "namespace" filter. It's essentially a "partial name" filter.

Jarga: That's true, its not really limited to "namespaces" since it's just a "name contains". Would it be better as "partialname" parameter and have "IncludedPartialNames" property in the XunitFilters?

bradwilson: I think appending a `.` and using `StartsWith` instead of `Contains` would come close enough to a "namespace" filter (which I think is probably the right feature).

Jarga: Done. I think using contains gives a bit more flexibility on what you can filter by but this solves my use case. (Re-fixed the title of the pull request as well, i got a little ahead of myself on the changes)

bradwilson:
Merged, thanks!

Jarga:
Awesome, thank you!

PR #580, opened 2015-09-04
Add strong name to xunit.execution.dotnet.dll

dsplaisted:
As described in dotnet/coreclr#1477, there's a bug in CoreCLR where loading an unsigned assembly without specifying a version number can fail.  this PR works around this by adding a strong name to xunit.execution.dotnet.dll.


NuGet.Config
```diff
@@ -4,7 +4,8 @@
     <add key="All" value="(Aggregate source)" />
   </activePackageSource>
   <packageSources>
-    <add key="Public Feed" value="https://api.nuget.org/v3/index.json" protocolVersion="3" />
+    <!--<add key="Public Feed" value="https://api.nuget.org/v3/index.json" protocolVersion="3" />-->
+    <add key="Public Feed" value="https://api.nuget.org/v2/ />
```
bradwilson: Why is this change here?

dsplaisted: Oops.  I was getting an error restoring gitlink when I tried to build, and this change seemed to fix it.  I'll rebase it out.

bradwilson:
Merged, thanks!

Issue #605, opened 2015-09-26
MSBuild runner: Add an IgnoreFailures parameter.


tstewart65:
Add an IgnoreFailures parameter to the MSBuild test runner so that the runner can ignore test failures and not fail the build.

PR #606, opened 2015-09-26
Fix for #605 - MSBuild Runner add an IgnoreFailures parameter.

tstewart65:
MSBuild runner modification - Added an IgnoreFailures parameter.  If IgnoreFailures is set true then the Execute function will return true if there are test failures.  This allows for better integration with TeamCity since it will allow you to use the Mute Test feature to Mute failing tests.


bradwilson:
Thanks!

Issue #254, opened 2015-01-09
Assert.Collection should include the collection in the failure when counts don't match


bradwilson:
Right now, when the counts don't match, it just shows the counts. It should show the collection as well, so the user can easily inspect what's in it.

jongleur1983:
fixes https://github.com/xunit/xunit/pull/617

PR #617, opened 2015-10-07
solving #254

jongleur1983:
## How I solved the issue
added an additional property to CollectionException that holds the actual elements. add that as an optional constructor parameter
add the actual values to the CollectionException where thrown.

## Tests:
- MismatchedElementsWithElementList is similar to MismatchedElements, but compares the ActualElements list instead of count for assertion,
- MismatchedElementCountWithElementList is similar to MismatchElementCount, but compares the ActualElements list instead of count for assertion.

### Not implemented: Actual Elements as part of the exception message
Adding the actual elements to the exception message would have been dangerous as collections can be quite long and IMHO the message itself is not the right way to print that. Thus I  used a property instead. 
Developers can use the ActualElements-Property of the CollectionException by code inspection while debugging or when handling the exception in code instead.

### DIFF
--- a/src/xunit.assert/Asserts/CollectionAsserts.cs
+++ b/src/xunit.assert/Asserts/CollectionAsserts.cs
@@ -271,7 +271,7 @@ namespace Xunit
         /// exactly one element.</exception>
         public static void Single(IEnumerable collection, object expected)
         {
-            Single(collection.Cast<object>(), item => Object.Equals(item, expected));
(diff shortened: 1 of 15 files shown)


bradwilson:
I would like to see this use ArgumentFormatter to include the collection in the message; if it's not there, then value is minimized (I'm forced to debug to find the information).

jongleur1983:
didn't know ArgumentFormatter before, but yes, that's a good solution for it. Added it now.

Issue #623, opened 2015-10-14
Coding Style Conventions


jongleur1983:
Hi,
I tried to dive deeper into the xunit code to understand more about how it works and stumbled over several things I personally don't like regarding the style of source code.

I am aware that something of it may be a matter of taste and my own opinion, but I think, some things could be done by convention to make code easier to read and better to maintain.

To name a few:

* when to use string and when to use String? Both are equivalent as string is an alias in the c# language for System.String. most often string is used, but there are a lot of uses of String as well in the xunit code base.
* Using code blocks: If statements that only contain a single statement often don't have a curly braces block in the code base. This is more concise, but error prone as it's easy to overlook what is part of the if statement and what is not.

I didn't find any statement about stuff like that and how it should be used at xUnit. 
Even if it's not enforced: is there a best practise where code cleanup work would be apprechiated, or are pull requests dealing with that considered nonsense noise?

bradwilson:
We don't have a formal style guide, primarily because we review and edit every PR.

> when to use string and when to use String? Both are equivalent as string is an alias in the c# language for System.String. most often string is used, but there are a lot of uses of String as well in the xunit code base.

We should be using `string` everywhere, but we undoubtedly don't have this covered 100%.

> Using code blocks: If statements that only contain a single statement often don't have a curly braces block in the code base. This is more concise, but error prone as it's easy to overlook what is part of the if statement and what is not.

This is a personal style thing, and generally speaking, we don't add unnecessary curly braces.

> Even if it's not enforced: is there a best practise where code cleanup work would be apprechiated, or are pull requests dealing with that considered nonsense noise?

I suppose a (theoretical) PR to create code consistency would be fine, so long as it was in the style that we prefer. Given my opening statement, it would be quite difficult for someone to get this right 100% of the time. :)

PR #624, opened 2015-10-14
code cleanup: replace Object by object and String by string.

jongleur1983:
as discussed in https://github.com/xunit/xunit/issues/623 this PR replaces occurrences of String by string and Object by object where appropriate.
Does not change any functionality, nor does it change any other code style or something like that.


bradwilson:
Thanks!

PR #625, opened 2015-10-15
Fix of tests failing depending on current culture

yakimovim:
Two tests were failing on my machine because my default culture is different from English. Here comma is used instead of dot as a decimal separator. So I had "Time: 1,235s" instead of "Time: 1.235s".

I wrote a simple method that executes given delegate with given culture. It sets default culture before execution of the delegate and returns it back after that. Then I rewrote failing tests using this method. Now they pass on my machine.

### DIFF
--- a/src/common/ExecutionHelper.cs
+++ b/src/common/ExecutionHelper.cs
@@ -1,4 +1,5 @@
 using System;
+using System.Globalization;
 using System.IO;
 using System.Linq;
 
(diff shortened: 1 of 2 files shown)


src/common/ExecutionHelper.cs
```diff
@@ -55,5 +56,26 @@ public static string PlatformSuffix
         /// a generalized reference to the platform-specific assembly.
         /// </summary>
         public static readonly string SubstitutionToken = ".{Platform}";
+
+        /// <summary>
+        /// Executes given <paramref name="action"/> with given <paramref name="culture"/>
+        /// </summary>
+        /// <param name="culture">Culture of thread to execute the <paramref name="action"/></param>
+        /// <param name="action">Action to execute with <paramref name="culture"/></param>
+        public static void ExecuteWithCulture(CultureInfo culture, Action action)
```
bradwilson: Why not use [CulturedFact] instead? https://github.com/xunit/xunit/blob/master/test/test.utility/CulturedFactAttribute.cs

yakimovim:
Thank you. I'll rewrite it using CulturedFact attribute.

GaTechThomas:
I am not able to pick up a reference to CulturedFact using the Xunit.net assemblies. Can we use CulturedFact out of the box, or do we have to copy/paste code the way that it was done in xamarin?

PR #631, opened 2015-10-20
Fix of tests failing depending of culture using CulturedFact attribute

yakimovim:
Two tests were failing on my machine because my default culture is different from English (in class OnMessage_ITestExecutionSummary). Here comma is used instead of dot as a decimal separator. So I had "Time: 1,235s" instead of "Time: 1.235s".

I applied CulturedFact attribute to these tests. Now they pass on my machine.


bradwilson:
Thanks!

Issue #211, opened 2014-11-17
Apply TraitAttribute to whole assembly


krwq:
Hello,

I'm using many .csproj files in one project and using runner to run tests on all of them as part of the build.
It would be nice if xUnit supported applying Trait attribute on whole assembly as it would be easier to run tests related to some concrete feature.

I.e:
FeatureA.Tests.dll contains few sets of tests for FeatureA
FeatureB.Tests.dll contains few tests which test FeatureB but also use FeatureA in few of them.

It would be much easier to apply trait related to FeatureA to whole assembly than spamming this trait for every single class in FeatureA tests..

bradwilson:
Added with #711.

krwq:
Thanks Brad! :smile:

PR #637, opened 2015-10-22
Trait discoverers requiring method information. Assembly and namespace traits

yakimovim:
I think it would be nice to be able to put traits not only on classes and methods, but also on assemblies and namespaces. For example my solution in Visual Studio can contain several projects with unit tests and several projects with integration tests. I would like to put in projects with unit tests trait "TestType" = "Unit" and in projects with integration tests trait "TestType" = "Integration". So I implemented AssemblyTrait attribute and made XUnit to check assembly attributes for traits as well.

Here is another example. My solution may have one project with unit tests and one project with integration tests. I may want to place trait "SubSystem" = "Data" on namespaces in both projects to be able to run all tests for given sub-system. It is more difficult to solve this problem. C# does not allow to actually place attributes on namespaces. So NamespaceAttribute is also applied to assembly.  ITraitDiscoverer does not have any information about namespace of method for which it should provide traits. So I introduced new interface IMethodedTraitDiscoverer that accepts both IAttributeInfo and IMethodInfo into the GetTraits method. From IMethodInfo one can extract information about namespace of current method.

NamespaceAttribute has two properties. NamespaceCaseInsensitive specifies if namespace of current method should be compared with namespace of attribute disregarding case of letters. Default value is false as C# is case sensitive language. NotApplicableToNestedNamespaces allows not to apply this trait to "nested" namespaces. For example if trait is applied to "OuterNamespace" namespace then by default it also applied to "OuterNamespace.InnerNamespace" namespace. Default value of this property is false.

### DIFF
--- a/src/xunit.core/AssemblyTraitAttribute.cs
+++ b/src/xunit.core/AssemblyTraitAttribute.cs
@@ -1,0 +1,20 @@
+﻿using System;
+using Xunit.Sdk;
+
+namespace Xunit
+{
+    /// <summary>
+    /// Attribute used to decorate an assembly with arbitrary name/value pairs ("traits").
+    /// </summary>
+    [TraitDiscoverer("Xunit.Sdk.AssemblyTraitDiscoverer", "xunit.core")]
+    [AttributeUsage(AttributeTargets.Assembly, AllowMultiple = true)]
+    public sealed class AssemblyTraitAttribute : Attribute, ITraitAttribute
+    {
+        /// <summary>
+        /// Creates a new instance of the <see cref="AssemblyTraitAttribute"/> class.
+        /// </summary>
+        /// <param name="name">The trait name</param>
+        /// <param name="value">The trait value</param>
+        public AssemblyTraitAttribute(string name, string value) { }
+    }
+}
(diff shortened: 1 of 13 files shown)


bradwilson:
@yakimovim Thanks for the submission. I will discuss with Jim about adding this feature.

In the meantime, the size of this feature would require that we get a contribution agreement from you. Here is the sample: http://www.outercurve.org/assets/files/OutercurveFoundationContributionAgreement-editable.pdf

Please take a look at this and let us know if you'd be okay signing it, and we'll get back to you.

Thanks. :)

yakimovim:
Hello,

I'm okay with signing the agreement.

Ivan.

2015-11-01 22:25 GMT+03:00 Brad Wilson <notifications@github.com>:

bradwilson:
@yakimovim Please e-mail me at dotnetguy at gmail and we'll get the process started. Thanks!

bradwilson:
Jim and I discussed this, and we're in favor of doing this for the assembly (https://github.com/xunit/xunit/issues/211), but we don't really like the namespace annotation. Could you please make the changes to do just assembly annotations?

Thanks,
Brad & Jim

yakimovim:
I have created a pull request for issue #211 with just assembly annotations.


Ivan.

2015-12-06 0:52 GMT+03:00 Brad Wilson <notifications@github.com>:

bradwilson:
Covered by #711

Issue #640, opened 2015-10-24
Use custom IEqualityComparer for items in collections


odedw:
It would be nice to be able to use a custom comparer for objects in a collection when asserting on collections. Here's an example:

``` cs
[Fact]
public static void CollectionWithCustomComparer()
{
    var expected = new List<Foo> { new Foo() { Number = 1 }, new Foo() { Number = 2 } };
    var actual = new List<Foo> { new Foo() { Number = 1 }, new Foo() { Number = 2 } };

    Assert.Equal(expected, actual, new FooComparer());
}

class Foo
{
    public int Number { get; set; }
}

class FooComparer : IEqualityComparer<Foo>
{
    public bool Equals(Foo x, Foo y)
    {
        return x.Number == y.Number;
    }

    public int GetHashCode(Foo obj)
    {
        return obj.Number.GetHashCode();
    }
}
```

Issue #240, opened 2014-12-19
Create Assert.Equal for Sets


arthurward:
Assert.Equal has the same issue with the ISet interface as it did for IDictionary, as reported in issue #17.

odedw:
I'd like to take a look at this.

bradwilson:
Go for it! :)

bradwilson:
Implemented with #641.

PR #641, opened 2015-10-24
Refactor AssertEqualityComparer to handle sets

odedw:
Issue #240 

It was a little tricky checking whether the type implements `ISet` as there is no non generic `ISet` (as opposed to `IDictionary`).

I added the `ObjectEqualityComparerAdapter` class to be able to use the innerComparer if there is one and I couldn't use the `DefaultInnerComparer` since `AssertEqualityComparerAdapter` doesn't implement `GetHashCode`. Although, I couldn't find a way to pass a custom inner comparer (opened a suggestion - issue #640).

### DIFF
--- a/src/xunit.assert/Asserts/Sdk/AssertEqualityComparer.cs
+++ b/src/xunit.assert/Asserts/Sdk/AssertEqualityComparer.cs
@@ -71,6 +71,11 @@ namespace Xunit.Sdk
             if (dictionariesEqual.HasValue)
                 return dictionariesEqual.GetValueOrDefault();
 
+            // Sets?
+            var setsEqual = CheckIfSetsAreEqual(x, y, typeInfo);
+            if (setsEqual.HasValue)
+                return setsEqual.GetValueOrDefault();
+
             // Enumerable?
             var enumerablesEqual = CheckIfEnumerablesAreEqual(x, y);
             if (enumerablesEqual.HasValue)
@@ -147,6 +152,38 @@ namespace Xunit.Sdk
             }
 
             return dictionaryYKeys.Count == 0;
+        }
+
+        bool? CheckIfSetsAreEqual(T x, T y, TypeInfo typeInfo)
+        {
+            if (!IsSet(typeInfo))
+                return null;
+
+            var enumX = x as IEnumerable;
+            var enumY = y as IEnumerable;
+            if (enumX == null || enumY == null)
+                return null;
+
+            var elementType = typeof(T).GenericTypeArguments[0];
+            MethodInfo method = GetType().GetTypeInfo().GetDeclaredMethod("CompareTypedSets");
+            method = method.MakeGenericMethod(new Type[] { elementType });
+            return (bool)method.Invoke(this, new object[] { enumX, enumY });
+        }
+
+        bool CompareTypedSets<R>(IEnumerable enumX, IEnumerable enumY)
+        {
+            var setX = new HashSet<R>(enumX.Cast<R>());
+            var setY = new HashSet<R>(enumY.Cast<R>());
+            return setX.SetEquals(setY);
+        }
+
+        bool IsSet(TypeInfo typeInfo)
+        {
+            return typeInfo.ImplementedInterfaces
+                .Select(i => i.GetTypeInfo())
+                .Where(ti => ti.IsGenericType)
+                .Select(ti => ti.GetGenericTypeDefinition())
+                .Contains(typeof(ISet<>).GetGenericTypeDefinition());
         }
 
         /// <inheritdoc/>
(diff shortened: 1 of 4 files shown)


src/xunit.assert/Asserts/Sdk/AssertEqualityComparer.cs
```diff
@@ -149,6 +158,36 @@ public bool Equals(T x, T y)
...
+        {
+            if (!IsSet(typeInfo))
+                return null;
+
+            var enumX = x as IEnumerable;
+            var enumY = y as IEnumerable;
+            if (enumX == null || enumY == null)
+                return null;
+
+            var objectEnumX = enumX.Cast<object>();
```
bradwilson: Should sets that aren't of the same type be considered equal?

src/xunit.assert/Asserts/Sdk/AssertEqualityComparer.cs
```diff
@@ -149,6 +158,36 @@ public bool Equals(T x, T y)
...
+            var objectEnumY = enumY.Cast<object>();
+            if (objectEnumX.Count() != objectEnumY.Count())
+                return false;
+
+            var setX = new HashSet<object>(objectEnumX, objectComparer);
```
bradwilson: Why use the object comparer, rather than getting a comparer for the T of `ISet<T>`?

odedw: T here is not the object type, it's ISet<M>. I couldn't find a way to cast the the sets to ISet<M> and compare them using ISet's comparer, that's why I used the object comparer (or the inner comparer if there is one). The result should be the same. Plus, I wanted to allow using a custom comparer in the future.

src/xunit.assert/Asserts/Sdk/AssertEqualityComparer.cs
odedw: That's a good question. if they are null I would say that they are equal, if not the object comparer will take care of it. What do you think?

src/xunit.assert/Asserts/Sdk/AssertEqualityComparer.cs
bradwilson: No, I _do actually mean_ the T in `ISet<T>` here (not the T in `Assert.Equal<T>`).

odedw: Ok I see what you mean. 
If I can somehow get the Type of T I can maybe do something like this:
```
var comparerType = typeof(Comparer<>).MakeGenericType(typeof(T));
IComparer comparer = (IComparer)Activator.CreateInstance(comparerType);
```
However the hashset c'tor only accepts IEqualityComparer not IComparer. Is that what you had in mind?

bradwilson: You can use `AssertEqualityComparer<>` instead of `Comparer<>`. :)

odedw: [Good idea](http://blogs.telegraph.co.uk/news/files/2013/07/homer-simpson-doh-400x288.jpg)! I'll try to make it work.

odedw: So I'm probably missing something, but I can't find a way to get the underlying element type as GetGenericArguments is not available in xunit.assert and xunit.core, and also I don't see a way to use it after creating an instance of AssertEqualityComparer of that type since I can't get a reference to the Comparer (of the correct type). 
```cs
var comparerType = typeof(AssertEqualityComparer<>).MakeGenericType(typeof(T).GetGenericArguments()/*unavailable*/);

object/*Of what type?*/ comparer = Activator.CreateInstance(comparerType);
```
Any ideas?

bradwilson: Try using .GetTypeInfo() on the Type to see if you get something more useful. The PCL reflection library functionality is split between Type and TypeInfo (for reasons I don't understand).

odedw: Found a way!
```cs
var elementType = typeof(T).GenericTypeArguments[0];
MethodInfo method = GetType().GetTypeInfo().GetDeclaredMethod("CompareTypedSets");
method = method.MakeGenericMethod(new Type[] { elementType });
return (bool)method.Invoke(this, new object[] { enumX, enumY });
```
where CompareTypedSets is:
```cs
bool CompareTypedSets<R>(IEnumerable enumX, IEnumerable enumY)
{
    var setX = new HashSet<R>(enumX.Cast<R>());
    var setY = new HashSet<R>(enumY.Cast<R>());
    return setX.SetEquals(setY);
}
```
I'll rewrite the commit history to include this. Thanks for the help!

odedw:
I removed the ObjectEqualityComparerAdapter and instead just created a Hashset from the underlying element type using reflection.

bradwilson:
Merged, thanks!

Issue #634, opened 2015-10-20
Can't use generic parameter type in Theory -- FormatTypeName null refs on generic parameters


nguerrera:
Given the following, the 'Repro' test will fail due to a NullReferenceException in v2.1.0.

``` C#
public static readonly object[][] ReproData = new[] {
    new object[] { typeof(List<>).GetTypeInfo().GenericTypeParameters.First() },
};

[Theory]
[MemberData("ReproData")]
public static void Repro(Type type) {}
```

The issue appears to be here: https://github.com/xunit/xunit/blob/6c62c646ef19ae1fb938447805afd132fa90d966/src/xunit.assert/Asserts/Sdk/ArgumentFormatter.cs#L174

Regrettably, TypeInfo.FullName is null for type parameters.

jongleur1983:
as far as I can see the problem is that `typeof(List<>).GetTypeInfo().GenericTypeParameters.First()`does return something like a Type placeholder, where TypeName is T, but FullName is null, as long as T is not set to a specific type.

Using `typeof(List<int>.GetTypeInfo().GenericTypeParameters.First()` works as expected of course.

I traced it down to an exception silently handled in TheoryDiscoverer.Discover, digging further now.

jongleur1983:
MSDN agrees with the behaviour of https://github.com/xunit/xunit/blob/6c62c646ef19ae1fb938447805afd132fa90d966/src/xunit.assert/Asserts/Sdk/ArgumentFormatter.cs#L174 here: https://msdn.microsoft.com/de-de/library/system.type.fullname%28v=vs.110%29.aspx , describing the value of `TypeInfo.FullName` as
> The fully qualified name of the Type, [...]; or null if the current instance represents [...] a generic type that is not a generic type definition but contains unresolved type parameters.
This seems to be the case here, as `List<>` basically is `List<T>` with unresolved `T`.

Therefore the bug is in fact in the ArgumentFormatter that currently assumes the FullName to be set on any type it get's.

So how to deal with it?
I think `ArgumentFormatter.FormatTypeName(Type type)` should at least deal with this special case.

jongleur1983:
locally I implemented tests covering that issue by directly invoking the ArgumentFormatter for unbound generic type arguments;
It would be easy to return the generic argument name in that case from `FormatTypeName()`; but as a result it would then return `typeof(T)` for a generic type argument that is named T, which seems not very useful at all.
[...]

PR #642, opened 2015-10-24
catch special case of ArgumentFormatter.Format(object) where the para…

jongleur1983:
solving bug in `ArgumentFormatter.Format(object)` when parameter is a generic type parameter not bound to a specific type, as explained in #634.


bradwilson:
Thanks!

Issue #287, opened 2015-02-05
Provide an option for outputting JSON status objects similar to /teamcity


hach-que:
The "-teamcity" option outputs the status of running tests as it occurs.  It'd be nice to have an option that outputs JSON objects separated by newlines as the status of running tests change, as this would make it easier to integrate progress indicators into other tools (because most languages can parse JSON).

Basically the output might look something like:

```
{ "type": "testSuiteStarted", "name": "Test collection for ...", "flowId": "..." }
{ "type": "testStarted", "name": "Test collection for ...", "flowId": "..." }
{ "type": "testFinished", "name": "Test collection for ...", "flowId": "...", "duration": 123 }
{ "type": "testFailed", "name": "Test collection for ...", "flowId": "...", "details": "..." }
```

odedw:
I'd like to take a look at this.

bradwilson:
Implemented with #647.

PR #647, opened 2015-10-27
Add JsonReporter with tests

odedw:
(issue #287)
Added a JsonReporter that can be used with the flag '-json'.
I added the messages and fields I thought would be useful, but I built it in a way that would make it easy to add more fields.
The tests verify that all of the messages are valid JSONs.

Here is an output example (running on the reporters test assembly):
https://gist.github.com/odedw/01cdda96f1ae5143406f#file-gistfile1-txt

### DIFF
--- a/src/xunit.runner.reporters/JsonExtentions.cs
+++ b/src/xunit.runner.reporters/JsonExtentions.cs
@@ -1,0 +1,199 @@
+using System;
+using System.Collections.Generic;
+using Xunit.Abstractions;
+#if PLATFORM_DOTNET
+using Newtonsoft.Json;
+#else
+using System.Web.Script.Serialization;
+#endif
+
+namespace Xunit.Runner.Reporters
+{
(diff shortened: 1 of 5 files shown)


bradwilson:
@odedw Thanks for this. Given the size of the commit, we'll need to get a contribution agreement from you. Can you please e-mail me at dotnetguy at gmail and we'll get the process started? If you want to read it before contacting us, this is the sample agreement: http://www.outercurve.org/assets/files/OutercurveFoundationContributionAgreement-editable.pdf

Thanks. :)

bradwilson:
CA received from Outercurve.

odedw:
Hey, just wanted to be sure, do I need to do something else?

bradwilson:
Nope. I was hoping to get to this this past weekend, but didn't.

odedw:
OK cool, no worries.

bradwilson:
Merged, thanks!

PR #657, opened 2015-11-01
Use FileShare.Read when opening transform files

### LINKED ISSUE #656: Transforms should open files with FileShare.Read
The transform code today doesn't specify a `FileShare` mode when opening files and hence inherits the default of no sharing (even though the file is open with `FileMode.Read`).  This means two xunit processes can race for the ability to read this file and in such a case the loser will exit with the following error:

```
  System.IO.IOException: The process cannot access the file 'C:\Users\jaredpar\.nuget\packages\xunit.runner.console\2.1.0\tools\HTML.xslt' because it is being used by an
  other process.
     at System.IO.__Error.WinIOError(Int32 errorCode, String maybeFullPath)
     at System.IO.FileStream.Init(String path, FileMode mode, FileAccess access, Int32 rights, Boolean useRights, FileShare share, Int32 bufferSize, FileOptions options,
   SECURITY_ATTRIBUTES secAttrs, String msgPath, Boolean bFromProxy, Boolean useLongPath, Boolean checkHost)
     at System.IO.FileStream..ctor(String path, FileMode mode, FileAccess access, FileShare share)
     at Xunit.ConsoleClient.TransformFactory.Handler_XslTransform(String xslPath, XElement xml, String outputFileName)
     at Xunit.ConsoleClient.TransformFactory.<>c__DisplayClass2_1.<.ctor>b__1(XElement xml, String outputFileName)
     at Xunit.ConsoleClient.TransformFactory.<>c__DisplayClass5_0.<GetXmlTransformers>b__1(XElement xml)
     at Xunit.ConsoleClient.Program.<>c__DisplayClass10_0.<RunProject>b__5(Action`1 transformer)
     at System.Collections.Generic.List`1.ForEach(Action`1 action)
     at Xunit.ConsoleClient.Program.RunProject(XunitProject project, Boolean serialize, Nullable`1 parallelizeAssemblies, Nullable`1 parallelizeTestCollections, Nullable
```

jaredpar:
This allows xunit process to run in parallel without contending for access to the transform file.

closes #656

### DIFF
--- a/src/xunit.console/Utility/TransformFactory.cs
+++ b/src/xunit.console/Utility/TransformFactory.cs
@@ -60,7 +60,7 @@ namespace Xunit.ConsoleClient
             var xmlTransform = new XslCompiledTransform();
 
             using (var writer = XmlWriter.Create(outputFileName, new XmlWriterSettings { Indent = true }))
-            using (var xsltStream = File.Open(xslPath, FileMode.Open, FileAccess.Read))
+            using (var xsltStream = File.Open(xslPath, FileMode.Open, FileAccess.Read, FileShare.Read))
             using (var xsltReader = XmlReader.Create(xsltStream))
             using (var xmlReader = xml.CreateReader())
             {


bradwilson:
Thanks!

PR #688, opened 2015-11-19
Allow asserting that a specific piece of code raises a specific event.

vegardlarsen:
A pattern I've encountered quite a few times when unit testing recently is to assure an event is raised as the direct cause of some action. This can be quite cumbersome:

```cs
var completedRaised = false;
var x = new Foo();
x.Completed += (sender, args) => completedRaised = true;
x.Start();
Assert.True(completedRaised);
```

This pull request makes this a little easier to do, especially when you have multiple events to test:

```cs
var x = new Foo();
var raisedEvent = Assert.Raises<EventArgs>(
    h => x.Completed += h, 
    h => x.Completed -= h,
    () => x.Start());
```

There are several advantages:
- much shorter
- ensures proper setup and teardown of the event handlers
- easier to inspect the sender and arguments
- easily fail the test if you get event arguments derived from what you expect
- supports `async` methods (`RaisesAsync` and `RaisesAnyAsync`)
- supports accepting derived event args (`RaisesAny` and `RaisesAnyAsync`)

I've tried my best to stick to the coding style in the project, but please let me know if I've missed anything, and I'll be happy to fix it.

### DIFF
--- a/test/test.xunit.assert/Asserts/EventAssertsTests.cs
+++ b/test/test.xunit.assert/Asserts/EventAssertsTests.cs
@@ -1,0 +1,214 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+using Xunit;
+using Xunit.Sdk;
+
+public class EventAssertsTests
+{
+    public class Raises_Generic
+    {
+        [Fact]
+        public static void ExpectedEventButCodeDoesNotRaise()
+        {
+            try
+            {
+                var x = new RaisingClass();
+                Assert.Raises<EventArgs>(h => x.Completed += h, h => x.Completed -= h, () => { });
+            }
+            catch (RaisesException exception)
+            {
+                Assert.Equal("(No event was raised)", exception.Actual);
+                Assert.Equal("EventArgs", exception.Expected);
+            }
+        }
+
+        [Fact]
+        public static void ExpectedEventButRaisesEventWithDerivedArgs()
+        {
+            try
+            {
+                var x = new RaisingClass();
+                Assert.Raises<EventArgs>(
+                    h => x.Completed += h,
+                    h => x.Completed -= h,
+                    () => x.RaiseWithArgs(new DerivedEventArgs()));
+            }
+            catch (RaisesException exception)
+            {
+                Assert.Equal("DerivedEventArgs", exception.Actual);
+                Assert.Equal("EventArgs", exception.Expected);
+                Assert.Equal("(Raised event did not match expected event)", exception.UserMessage);
+            }
+        }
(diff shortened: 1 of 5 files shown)


bradwilson:
Thanks for this! It's a nice feature. :)

Substantial contributions require a Contribution Agreement with Outercurve (you can read an example here: http://www.outercurve.org/assets/files/OutercurveFoundationContributionAgreement-editable.pdf). Would you be comfortable signing one? If so, please e-mail me at dotnetguy at gmail dot com and I'll get the process started.

vegardlarsen:
The contribution agreement has now been signed. :)

bradwilson:
We'll take a look when we get together again.

In the meantime, it appears that you have 13 failing tests, according to the build check.

vegardlarsen:
@bradwilson I noticed, but I assumed you were having trouble with your build server. They are all failing because of problems loading `Moq` (`System.BadImageFormatException : Could not load file or assembly 'Moq, Version=4.1.1308.2321, Culture=neutral, PublicKeyToken=69f491c39445e920'`). They do not occur on my machine, and I haven't changed any references or package dependencies.

I do however have 7 failing unit tests in the solution when it runs on my machine, but those were the same 7 tests that failed on a clean checkout from `master`, so I am assuming that should be OK.

test/test.xunit.assert/Asserts/EventAssertsTests.cs
```diff
@@ -0,0 +1,214 @@
...
+{
+    public class Raises_Generic
+    {
+        [Fact]
+        public static void ExpectedEventButCodeDoesNotRaise()
+        {
+            try
+            {
+                var x = new RaisingClass();
+                Assert.Raises<EventArgs>(h => x.Completed += h, h => x.Completed -= h, () => { });
```
bradwilson: I would prefer to see these written with `Assert.Throws` (or `Record.Exception` + `Assert.IsType`) rather than try/catch, because as written, it's easier for them to pass false positives.

bradwilson: @vegardlarsen Since there aren't too many of them, just consider this a note for the future. I'll fix them myself. :)

bradwilson:
Merged, thanks!

vegardlarsen:
Great! Note taken! :)

RehanSaeed:
This only seems to support the `EventHandler<T>`. What about `EventHandler` without the `<T>` or indeed events that don't use the standard `EventHandler<T>` handler like the `PropertyChangedEventHandler`?

Reactive Extensions has methods very similar to this and provides overloads for all of the above.

bradwilson:
@RehanSaeed Please open a new issue.

PR #694, opened 2015-11-25
Fix governance link

bdukes:
The governance link was linking relative to the GitHub repo, rather than to the (separate) `xunit.github.io` domain.


bradwilson:
Oops. :)

Issue #695, opened 2015-11-26
MemberDataAttribute works counter-intuitive


jongleur1983:
Today I stumbled over a case that sounded strange in the first place.

I wrote a xunit Theory that get's it's input from a PropertyData-Attribute (now replaced by MemberDataAttribute I think, but I'm pretty sure the issue stays the same as I didn't found anything about a change here).

This attribute requires the name of a property of type `IEnumerable<object[]>`, located in the test class.

So far everything sounds fine, but using an IEnumerable I assumed that any object that is enumerated by `yield` is used to call the test.
Instead, the test runner enumerates the whole property, similar to calling `ToList()` before executing the first test run.

In my case the reason to use this attribute is a quite complex data structure that serves the test input. As the objects are mutable it seemed to be easier to yield something, mutate it and yield it again with the now different state, but as the enumerable is completely enumerated before, my test in this case receives the first element not in the state I expected (that's the one when it has been yielded), but in the state it has after complete enumeration, something entirely different.

I therefore propose to do two things to clarify it:
* clearly state how it behaves in the documentation of the attribute
* change behaviour or api:
 * either, the enumerable should not be enumerated completely beforehand, or
 * the property should preferably be of type `ICollection<object[]>`instead, which may be basically the same implementation, but it's more clear that it's a list already, and all objects are inside from start on (I know that that's technically not true, but it's not that misleading).

I personally would prefer the first solution, as that would allow to mutate objects and call the same test routine in different states of the same object.

jongleur1983:
As a workaround I used the following for now:
```C#
public static IEnumerable<object[]> WorkAroundInputSource
{
    get
    {
        var count = MutableTestInputSource.Count();
        for (int i = 0; i < count; i++)
        {
            yield return MutableTestInputSource.Skip(i).First();
        }
        yield break;
    }
}

public static IEnumerable<object[]> MutableTestInputSource 
{
    get
    {
        var result = this.GenerateFirstInput();
        yield return result;
        this.Mutate(result); // result is in a different state afterwards
        yield return result;
        // ...
    }
}

[Theory]
[PropertyData("WorkAroundInputSource")]
public void MyTest(params...) {
    // do the test
}
```

Unfortunately this iterates through the enumerable several times which may have a substantial impact on the test performance, especially for long lists and complex input data setups.

bradwilson:
> Instead, the test runner enumerates the whole property, similar to calling ToList() before executing the first test run.

It's not similar-to, it is exactly that.

[...]

PR #700, opened 2015-11-29
better documentation to avoid issue #695

jongleur1983:
make clear in the documentation that MemberDataAttribute should not return the same objects twice.

### DIFF
--- a/src/xunit.core/ClassDataAttribute.cs
+++ b/src/xunit.core/ClassDataAttribute.cs
@@ -8,6 +8,7 @@ namespace Xunit
     /// <summary>
     /// Provides a data source for a data theory, with the data coming from a class
     /// which must implement IEnumerable&lt;object[]&gt;.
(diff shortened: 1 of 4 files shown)


bradwilson:
This documentation is incomplete, because the rule in question applies to _**all**_ data attributes, which means it should probably be placed on DataAttribute itself as well as all derived attributes.

jongleur1983:
good point, will look into it later.

jongleur1983:
? How can that commit that added only comments lead to a failure of teamcity? Sorry, don't understand what's wrong there, sorry in case my commits are causing the trouble, but it would be great if someone could tell me if I did something wrong or how to solve it.

onovotny:
Your commit didn't break the build; VS 2015 Update 1 on the build box is causing the RefactoringEssentials analyzer package to throw and is thus breaking the build.

jongleur1983:
good to know, thanks

bradwilson:
Thanks!

Issue #607, opened 2015-09-26
build.sh is broken


ermshiperete:
Trying to build by running ./build.sh on Linux results in `Directory '$HOME/Develop/xunit/xunit.xbuild.sln' not found.`

bradwilson:
I just uploaded a fix for this.

However, `build.sh` is still not functional, because neither NuGet 3.x package restore nor C# 6 features will work on Mono right now.

kostrse:
Faced the similar problem, thanks for information.

OS X with Mono 4.0.4:
```
~/Dev/xunit $ ./build.sh 
Restoring NuGet packages...
MSBuild auto-detection: using msbuild version '4.0' from '/Library/Frameworks/Mono.framework/Versions/4.0.4/lib/mono/4.5'.
Error parsing solution file at /Users/user1/Dev/xunit/xunit.vs2015.sln: Could not load type 'Microsoft.Build.Construction.SolutionParser'.
```

BTW, maybe switch to Paket?
It won't solve the issue with C# 6, but at least a much better (and cross-platform) package manager.

bradwilson:
> BTW, maybe switch to Paket?

No.

ErikSchierboom:
I had the same problem and was able to fix it by referencing Mono 3.3.0 in the `build.sh` file. Should I send a PR?

bradwilson:
Sure. Thanks!

akoeplinger:
@ErikSchierboom do you mean NuGet 3.3? Cause Mono 3.3 didn't have any C#6 features.

I am able to compile xunit with a recent Mono (i.e. from source) with these changes: https://github.com/akoeplinger/xunit/commit/cc0c8743d47b377fbf3d3bcb850cea8d652c037d.

onovotny:
I wouldn't recommend mono 3.x anyway as there are major issues with deadlocks. Mono 4 does support C#6 syntax:
http://www.mono-project.com/docs/about-mono/releases/4.0.0/

onovotny:
As far as those specific changes in https://github.com/akoeplinger/xunit/commit/cc0c8743d47b377fbf3d3bcb850cea8d652c037d, I can suggest two things.

1.  the download for nuget should be `https://dist.nuget.org/win-x86-commandline/latest/nuget.exe`. That will always get the latest version.
[...]

PR #706, opened 2015-12-05
Use latest NuGet.exe file for Linux/Mac builds. References #607

ErikSchierboom:
The `build.sh` file is updated to use the latest NuGet.exe, which fixes the build.

For the relevant NuGet issue, see: https://github.com/NuGet/Home/issues/1543.


bradwilson:
Thanks!

PR #708, opened 2015-12-05
Adds support for marking DataAttributes as skipped which resolves #532

### LINKED ISSUE #532: Mark a InlineData as 'Skip'
I know we can mark a fact as 'skip', but I can't seem to find a way to do it for theories and inline data.  I would like to be able to do this so that I don't have failures, but can still see that there are tests not working.

jageall:
I wasn't sure whether to make the XunitSkippedDataRowTestCase internal or not. I went with public as I don't believe this is a breaking change and figured it would be needed if anyone derived from TheoryDiscoverer.


bradwilson:
Thanks!

PR #711, opened 2015-12-09
Issue#211. Apply TraitAttribute to whole assembly

yakimovim:
This is the implementation of Issue #211. Another possible solution: do not create separate AssemblyTrait attribute, but allow existing Trait attribute to be set on assembly.

### DIFF
--- a/src/xunit.core/AssemblyTraitAttribute.cs
+++ b/src/xunit.core/AssemblyTraitAttribute.cs
@@ -1,0 +1,20 @@
+﻿using System;
+using Xunit.Sdk;
+
+namespace Xunit
+{
+    /// <summary>
+    /// Attribute used to decorate an assembly with arbitrary name/value pairs ("traits").
+    /// </summary>
+    [TraitDiscoverer("Xunit.Sdk.AssemblyTraitDiscoverer", "xunit.core")]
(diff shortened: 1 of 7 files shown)


bradwilson:
Thanks!

krwq:
Thanks!

julealgon:
> Another possible solution: do not create separate AssemblyTrait attribute, but allow existing Trait attribute to be set on assembly.

I'll share my two cents about this. My very first intuition when the question came to me, was to immediately check the `TraitAttribute`'s `AttributeUsage` values and look for assembly there.

As soon as I noticed it could only be applied to classes and methods, I thought "hmm maybe this is just not possible? I'll search for it..." and then I found this issue. When I found out the issue was implemented (but before checking _how_ it was implemented), I then immediately assumed "oh, so its supposed to work! I must be using an outdated version of the XUnit package, let me try to update..." only to find it was already at the latest version.

Only after being confused as to why it wasn't working as I expected, was when I actually looked into _how_ you did it and found out it was a dedicated attribute.

All of this is to say that, yes, having the same attribute controlling this would feel infinitely more natural and discoverable in my opinion, as someone who just went through this a few minutes ago.

I think you guys should reconsider the design here for a future version, as it would make the library more intuitive overall without one necessarily needing to go through documentation or in search of a different attribute "that means the same thing". That just feels like a design smell to me to be honest.

PR #712, opened 2015-12-11
Add Assert.Imply to assert implications and consequences

Reza1024:
Assert.Implies(foo, bar) would be a nice and more readable replacement for Assert.True(bar || !foo)

### DIFF
--- a/src/xunit.assert/Asserts/BooleanAsserts.cs
+++ b/src/xunit.assert/Asserts/BooleanAsserts.cs
@@ -89,5 +89,29 @@ namespace Xunit
             if (!condition.HasValue || !condition.GetValueOrDefault())
                 throw new TrueException(userMessage, condition);
         }
+
+         /// <summary>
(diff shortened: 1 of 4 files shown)


bradwilson:
@Reza1024 Thanks for the PR. I will discuss this with @jnewkirk, and if we decide to add this feature, we will need to get a contribution agreement signed from you. You can see an example here: http://www.outercurve.org/assets/files/OutercurveFoundationContributionAgreement-editable.pdf

Reza1024:
Hi @bradwilson I am OK to sign the agreement but need to go through the legal process with my company too. Please let me know when you made your decision.

jnewkirk:
Thanks for the submission. Brad and I have discussed this and we are not going to add this feature. The reasoning is that in our opinion it would not have broad usage and there are mechanisms for you to extend the Assert class if you wanted to include it in your code.

Issue #627, opened 2015-10-16
Premature PropertyChanged event cannot be tested against using Assert.PropertyChanged


robert-j-engdahl:
The following code raises the PropertyChanged event prematurely, so that listeners will only have access to the old value. 

``` c#
public string SomeProperty
{
    get { return _someProperty; }
    set 
    {
        RaisePropertyChanged()
        _someProperty = value;
    }
}
```

In its current form `Assert.PropertyChanged` is redundant at best, because the above code needs to be tested like so:

``` c#
[Theory, AutoData]
public void SomeProperty_Is_Observable1(string expectedValue)
{
    string actualValue = null;
    _target.PropertyChanged += (s, e) =>
    {
        if (e.PropertyName == nameof(_target.SomeProperty))
        {
            actualValue = _target.SomeProperty;
        }
    };

    _target.SomeProperty = expectedValue;

    Assert.Equal(expectedValue, actualValue);
}
```

The above test will catch the bug.  Unlike the following, which will hapily pass:

``` c#
[Theory, AutoData]
public void SomeProperty_Is_Observable2(string expectedValue)
{
    Assert.PropertyChanged(_target, nameof(_target.SomeProperty), () => 
        {
            _target.SomeProperty = expectedValue;
        });
}
```

A simple solution would be to add a templated version of `Assert.PropertyChanged()`, say `Assert.PropertyChanged<T>()`, which returns the property value as it was when `PropertyChanged` was received:

``` c#
[Theory, AutoData]
public void SomeProperty_Is_Observable3(string expectedValue)
{
    var actualValue = Assert.PropertyChanged<string>(_target, nameof(_target.SomeProperty), () => 
        {
            _target.SomeProperty = expectedValue;
        });
    Assert.Equal(expectedValue, actualValue);
}
```

bradwilson:
If the intention is that the property always sets the provided value, then why not do that for the user?

```csharp
Assert.PropertyChanged(_target, name(_target.SomeProperty), expectedValue);
// Sets the property, ensures that PropertyChanged was called, and gets the value to ensure the set value is returned
```

(I can see the value of both, just thinking that this signature would be more compact and hits the 90% case.)

robert-j-engdahl:
That is also one awesome method signature.  It lets an observable property, which is implemented in three statements, be tested in a unit test containing just the assertion statement.

[...]

PR #757, opened 2016-02-02
Issue #627: Let Assert.PropertyChanged detect premature PropertyChanged events

robert-j-engdahl:
The original issue is found here: https://github.com/xunit/xunit/issues/627

I have added optional expectedValue parameter to Assert.PropertyChanged and Assert.PropertyChangedAsync.  When not null, it will be checked (in addition to existing checks) that the actual value of the property being changed equals the expected value.

If it matters, this was made using TDD.

Existing facts covering the updated methods have been changed to theories that run with and without the expectedValue arguments (though it is given explicitly as null).


bradwilson:

bradwilson:
Closed for age.

PR #765, opened 2016-02-07
Adding support for wildcard in method filter ('method' command-line p…

jcouv:
Specifying method filters with the fully-specified method name is rather unfriendly and tedious.
By adding wildcard support to the 'method' parameter on the console runner, the user can run a single method with `*.MyMethod` or `*.MyClass.MyMethod`, which is more convenient.

One thing gave me some trouble with this change: the `IncludedMethods` member of the filters class. Because it is public, I left it as-is, to avoid breaking code that depends on it. But ideally, I wanted to make it private and to make adding method filters use the new public method `AddIncludedMethod` instead.

Regards,
Julien

### DIFF
--- a/src/xunit.runner.utility/Project/XunitFilters.cs
+++ b/src/xunit.runner.utility/Project/XunitFilters.cs
@@ -1,6 +1,7 @@
 ﻿using System;
 using System.Collections.Generic;
 using System.Linq;
+using System.Text.RegularExpressions;
 using Xunit.Abstractions;
 
 namespace Xunit
@@ -10,6 +11,8 @@ namespace Xunit
     /// </summary>
     public class XunitFilters
     {
+        List<Regex> includedMethodRegexes;
+
         /// <summary>
         /// Initializes a new instance of the <see cref="XunitFilters"/> class.
         /// </summary>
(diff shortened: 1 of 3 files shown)


src/xunit.runner.utility/Project/XunitFilters.cs
```diff
@@ -43,6 +47,43 @@ public XunitFilters()
         public HashSet<string> IncludedMethods { get; }
 
         /// <summary>
+        ///
```
robert-j-engdahl: you forgot this?

jcouv: Thanks for catching that. I'll fix.

jcouv:
I've made a new rebased commit, but I'm not sure how to update the Pull Request to pick it up. I'll create a new one.

https://github.com/jcouv/xunit/commit/4ef26bcfaee8c0beb61132a2d5ada11f18eda731

onovotny:
@jcouv in general, it's best to create a topic branch to work on instead of adding code to your master branch. Then, when you rebase the branch, you can do a force push. GitHub will automatically do the right thing if that branch is tied to a PR.

PR #766, opened 2016-02-08
Adding support for wildcard in method filter ('method' command-line p…

jcouv:
Specifying method filters with the fully-specified method name is rather unfriendly and tedious.
By adding wildcard support to the 'method' parameter on the console runner, the user can run a single method with `*.MyMethod` or `*.MyClass.MyMethod`, which is more convenient.

One thing gave me some trouble with this change: the `IncludedMethods` member of the filters class. Because it is public, I left it as-is, to avoid breaking code that depends on it. But ideally, I wanted to make it private and to make adding method filters use the new public method `AddIncludedMethod` instead.

Re-submitting with missing comment.

Regards,
Julien

### DIFF
--- a/src/xunit.console/CommandLine.cs
+++ b/src/xunit.console/CommandLine.cs
@@ -271,7 +271,7 @@ namespace Xunit.ConsoleClient
                     if (option.Value == null)
                         throw new ArgumentException("missing argument for -method");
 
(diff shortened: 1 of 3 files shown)


bradwilson:

bradwilson:
Thanks for the legwork on this! I'm going to run with it and add support to the rest of the runners.

Issue #547, opened 2015-08-19
VisualStudio Runner does not properly display tests with member data


dealproc:
I am unsure if this is an issue or not, so am reporting it to inquire.

When providing a MemberData attribute to a theory, if the member emits more than one item, it seems as if the Visual Studio runner is considering all items as a single test, instead of breaking them up into multiple tests.

Should a mock of this issue be created and posted for reference, or is it enough to just report this?

bradwilson:
In Visual Studio, it can only show separate tests per data row when the data in the data row is serializable.

Serializable here is defined as the handful of intrinsics that we have built-in support for. You can see the actual list of things here: https://github.com/xunit/xunit/blob/4e39f2c784fae4e864c67743f825c5f811c04d42/src/common/XunitSerializationInfo.cs#L104-L176

MemberData is not on this list. When you have non-serializable data, then you get a single test case for the theory, and running it runs all the data. You are no longer able to run any individual data row.

dealproc:
Ok.  Was mainly a curiosity.  So an anonymous type of intrinsic values cannot be split up.  No worries.  it just gets annoying because the test explorer is saying one thing, but the output in TeamCity is saying another.  I may be observing wrong, but will play a bit to verify.

bradwilson:
@dealproc For your own custom types, you can implement [IXunitSerializable](https://github.com/xunit/xunit/blob/master/src/abstractions/xunit.abstractions/Serialization/IXunitSerializable.cs).

Ausm:
@bradwilson
Does that mean, that there is no way to let xUnit call a theory method like
_public void TestSimpleExpression(string name, Expression<Func<MyClass, bool>> expression, int expectedCount)_

I tried it with a wrapper which implements [IXunitSerializable](https://github.com/xunit/xunit/blob/master/src/abstractions/xunit.abstractions/Serialization/IXunitSerializable.cs), and it worked as long as I used the wrapper class in the argument as well.
This one _public void TestSimpleExpression(string name, Wrapper<Expression<Func<MyClass, bool>>> expression, int expectedCount)_ worked, but when I remove the Wrapper<> type in the argument, a System.ArgumentException is thrown, and the implicit/explicite operator of the Wrapper<> was ignored.

[...]

PR #769, opened 2016-02-11
Cast test method arguments before calling test method.

Ausm:
Add code to cast all given test method arguments to the types get from the MethodInfo.
See discussion [here](https://github.com/xunit/xunit/issues/547#issuecomment-182578848).


Ausm:
I see, I need to improve that.
Forget this request, will come up with a better one.

PR #780, opened 2016-02-20
Adding timestamp for test results

jcouv:
This change records the time when the tests completed and displays that timestamp in the HTML results produced by the console runner.
Let me know if you'd like me to add this to other runners and/or transforms


bradwilson:
This is a good idea. We should put this value in from all the runners.

bradwilson:
Merged, thanks!

PR #804, opened 2016-03-19
Fix: Console runner crashes when writing invalid characters to XML file

jamesqo:
Hi! First off I'd like to thank you guys for making xUnit. It's a pretty nice framework, and when I first got introduced to it via CoreFX I had a way better time using it than with NUnit.

Anyway, I recently ran into a problem while using the console runner over at dotnet/corefx#7042 where it would crash while writing invalid surrogate characters to the XML file. For example, this would fail:

```csharp
public static IEnumerable<object[]> UrlEncode_TestData()
{
    yield return new object[] { "\uD800", "%EF%BF%BD" };
}

[Theory]
[MemberData(nameof(UrlEncode_TestData))]
public static void UrlEncode(string value, string expected)
{
    Assert.Equal(expected, WebUtility.UrlEncode(value));
}

// error: The surrogate pair (0xD800, 0x22) is invalid. A high surrogate character (0xD800 - 0xDBFF) must always be paired with a low surrogate character (0xDC00 - 0xDFFF).
```

This pull request fixes that by encoding surrogate characters as `\x----` in the XML (where `----` is the hex value of the char). I've also added tests to enforce this, and make sure that valid surrogate sequences are outputted as normal.

### DIFF
--- a/src/common/XmlTestExecutionVisitor.cs
+++ b/src/common/XmlTestExecutionVisitor.cs
@@ -275,6 +275,13 @@ namespace Xunit
             for (var idx = 0; idx < value.Length; ++idx)
                 if (value[idx] < 32)
                     escapedValue.Append($"\\x{((byte)value[idx]).ToString("x2")}");
+                else if (char.IsSurrogatePair(value, idx))
+                {
+                    escapedValue.Append(value[idx++]); // Append valid surrogate chars like normal
+                    escapedValue.Append(value[idx]);
+                }
+                else if (char.IsSurrogate(value, idx)) // Translate invalid ones to \x----
+                    escapedValue.Append($"\\x{((int)value[idx]).ToString("x4")}");
                 else
                     escapedValue.Append(value[idx]);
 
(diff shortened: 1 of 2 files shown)


bradwilson:
Thanks for the PR!

You've inadvertently uncovered a second failure (as you can see with the CI failure). I was able to recreate the problem by setting the environment variable `TEAMCITY_PROJECT_NAME` (to any value) before running `build.ps1` (to trigger the TeamCity output, which prints the badly formatted argument).

Build error

This makes me wonder whether we shouldn't move all of this logic out of the XML writer and into the [argument formatter](https://github.com/xunit/assert.xunit/blob/5393af8846737eea79db9ac3b6732e316377ea69/Sdk/ArgumentFormatter.cs#L75-L82), so that it's solved in one place instead of many... (You can see that we have similar logic for char parameters already there)

jamesqo:
@bradwilson Sounds great. I'll close this PR then, take a look into the `ArgumentFormatter` code and submit a PR there. Thanks for responding so quickly :smile:

jamesqo:
@bradwilson: I've added the tests for `ArgumentFormatter` I've promised in the assert.xunit PR. When I run the tests introduced here, though, it seems that they still fail- in other words, I don't think any of the code here is actually calling `ArgumentFormatter.Format`. What would you think of reopening this PR/making the `XmlTestExecutionVisitor` code call the `Format` function, so the original problem at hand gets solved?

Issue #759, opened 2016-02-04
[Proposal] Support Automatic "Pretty" Display Name for Test Cases


commonsensesoftware:
## Overview

The **DisplayName** property for facts and theories are great for making tests really easy to read, but after explicitly defining them for a couple of years now, it feels like Xunit could do better at providing an automatic, pretty display using _Convention over Configuration (CoC)_.

For example, many developers write tests as:

``` C#
public class MyTestClass
{
  [Fact]
  public void verify_that_something_does_the_right_thing() {  }
}
```

What the author intended the test to say is "verify that something does the right thing".  This is a limitation in the source language as opposed to an issue with Xunit.  Some Xunit extensions, like that for F#, allow you to write tests in exactly this way.
## Proposal

I propose supporting the automatic pretty display name of classes and methods that follow a well-defined convention. In doing so, display names become easier and more natural to read.

The following rules and conventions would apply:
- Automatic display formatting is enabled by default (IMHO few developers would want to opt out)
- The **DisplayName** property still overrides the default, formatted display name
- The convention rules must be applicable across programming languages
- The underscore ('_') character is always replaced with a space
- The following substitution tokens are supported if they have an underscore ('_') to both sides, appear as the first token, or appear as the last token. Token matches are case-insensitive. These are the same tokens supported in Windows PowerShell:
  - **eq** '='
  - **ne** '!='
  - **lt** '<'
  - **le** '<='
  - **gt** '>'
  - **ge** '>='
- An ASCII escape sequence can be specified using **X[HH]**, where the sequence is uppercase and **H** is a valid hexadecimal character
- A Unicode escape sequence can be specified using **U[HHHH]**, where the sequence is uppercase and **H** is a valid hexadecimal character
## Examples

The following outlines a few basic usage examples:

``` C#
public class Example
{
  [Fact]
  public void unit_tests_are_awesomeX21_U263A(){ }

  [Fact]
  public void api_version_1_should_be_lt_2(){ }

  [Fact]
  public void X3DX3D_operator_overload_should_be_equivalent_to_Equals_method(){ }

  [Fact]
  public void X27stuffedX27_should_not_be_ambiguous_X27stUFFEDX27(){ }

  [Fact]
  public void X27maxed_outX27_should_not_be_ambiguous_with_X27maXED_outX27(){ }

  [Fact]
  public void TestNameShouldRemainUnchanged(){ }

  [Fact( DisplayName = "custom test name" )]
  public void CustomTestName(){ }

[...]

PR #828, opened 2016-04-16
"Pretty" Display Name Implementation for Test Cases

commonsensesoftware:
## Summary

This PR contains the implementation for "pretty" test case display names as described in issue #759.

The default behavior for test case display names is unchanged.  Developers must still opt into changing the display name settings via the **xunit.methodDisplay** configuration setting, but now with additional options.

The following new **TestMethodDisplay** options can be used in conjunction with the existing options to control test case display names:
- **ReplaceUnderscoreWithSpace** - replaces each '_' with a ' '
- **AllowOperatorMonikers** - recognizes and replaces supported operator monikers (ex: 'eq' to "=")
- **AllowEscapeSequences** - recognizes and replaces supported escape sequences (ex: 'X2C' to ',')

Developers can choose which, if any, of these options to turn on.  I realize not everyone may agree with using all of these options, but for those that do, I believe it's significantly more productive than re-entering special text via the **DisplayName** of facts or theories for scenarios that only involve a few characters.  The **DisplayName** continues to trump all of these settings and should be used for complex test case names to prevent the corresponding method names from becoming unintelligible.
## Notes

I chose to change the **TestMethodDisplay** to be a bit vector rather than introduce a new enumeration and configuration option.  I feel these new settings are too close in meaning with the existing options to warrant creating completely new configuration options. Since the values are backward-compatible, this seemed more than reasonable.

Changing the **TestMethodDisplay** to support flags means that it should _technically_ be renamed to indicate this behavior (ex: "TestMethodDisplayOptions"). Since this would introduce a breaking change, I decided against it.

For reasons I cannot explain, the value specified for **xunit.methodDisplay** will not parse when the new options are specified unless the value is supplied as an integer.  For example, the value **"Method,AllFormatExtensions"** should work, but it doesn't; however, using **"30"** does.  I've gone through all of the parsing code, unit tests, etc. and I can't explain this behavior.  Parsing bit vector enumerations from strings this way has been supported since .NET 1.0.  The exact cause and solution escapes me. For now, I'm accepting this limitation.  If you know why this is happening, I'm happy to make the appropriate update.

I haven't signed the contribution agreement yet, but I've read it, agree with it, and I'm happy to return a signed copy to you once it's provided. I've reviewed all of the contribution guidelines and believe that I have followed them accordingly.
## Screenshots

Here are some example screenshots from a demo application using a local build:
basic-example
gwt-example

### DIFF
--- a/src/common/TestMethodDisplayOptions.cs
+++ b/src/common/TestMethodDisplayOptions.cs
@@ -1,0 +1,63 @@
+#if XUNIT_FRAMEWORK
+namespace Xunit.Sdk
+#else
+namespace Xunit
+#endif
+{
+    using System;
+
+    /// <summary>
+    /// Indicates the method display options for test methods.
+    /// </summary>
+    [Flags]
+    public enum TestMethodDisplayOptions
+    {
+        /// <summary>
+        /// Indicates no additional method display options.
+        /// </summary>
+        /// <remarks>This is the default configuration option.</remarks>
+        None = 0x00,
+
+        /// <summary>
+        /// Replace underscores in display names with a space.
+        /// </summary>
+        ReplaceUnderscoreWithSpace = 0x01,
+
+        /// <summary>
+        /// Replace well-known monikers with their equivalent operator.
+        /// </summary>
+        /// <list type="bullet">
+        /// <item><description>lt : &lt;</description></item>
+        /// <item><description>le : &lt;=</description></item>
+        /// <item><description>eq : =</description></item>
+        /// <item><description>ne : !=</description></item>
+        /// <item><description>gt : &gt;</description></item>
+        /// <item><description>ge : &gt;=</description></item>
+        /// </list>
+        UseOperatorMonikers = 0x02,
+
+        /// <summary>
+        /// Replace supported escape sequences with their equivalent character.
+        /// <list type="table">
+        /// <listheader>
+        ///  <term>Encoding</term>
+        ///  <description>Format</description>
+        /// </listheader>
+        /// <item><term>ASCII</term><description>X hex-digit hex-digit (ex: X2C)</description></item>
+        /// <item><term>Unicode</term><description>U hex-digit hex-digit hex-digit hex-digit (ex: U00A9)</description></item>
+        /// </list>
+        /// </summary>
+        UseEscapeSequences = 0x04,
+
+        /// <summary>
+        /// Replaces the period delimiter used in namespace and type references with a comma.
+        /// </summary>
+        /// <remarks>This option is only honored if the <see cref="TestMethodDisplay.ClassAndMethod"/> setting is also enabled.</remarks>
+        ReplacePeriodWithComma = 0x08,
+
+        /// <summary>
+        /// Enables all method display options.
+        /// </summary>
+        All = ReplaceUnderscoreWithSpace | UseOperatorMonikers | UseEscapeSequences | ReplacePeriodWithComma
+    }
+}
--- a/src/common/TestOptionsNames.cs
+++ b/src/common/TestOptionsNames.cs
@@ -5,6 +5,7 @@ static class TestOptionsNames
         public static readonly string DiagnosticMessages = "xunit.discovery.DiagnosticMessages";
         public static readonly string InternalDiagnosticMessages = "xunit.discovery.InternalDiagnosticMessages";
         public static readonly string MethodDisplay = "xunit.discovery.MethodDisplay";
+        public static readonly string MethodDisplayOptions = "xunit.discovery.MethodDisplayOptions";
         public static readonly string PreEnumerateTheories = "xunit.discovery.PreEnumerateTheories";
         public static readonly string SynchronousMessageReporting = "xunit.discovery.SynchronousMessageReporting";
     }
--- a/src/xunit.core/xunit.core.csproj
+++ b/src/xunit.core/xunit.core.csproj
@@ -10,6 +10,7 @@
     <Compile Include="..\common\ExceptionExtensions.cs" LinkBase="Common" />
     <Compile Include="..\common\Guard.cs" LinkBase="Common" />
     <Compile Include="..\common\TestMethodDisplay.cs" LinkBase="Common" />
+    <Compile Include="..\common\TestMethodDisplayOptions.cs" LinkBase="Common" />
     <Compile Include="..\xunit.assert\Asserts\Sdk\ArgumentFormatter.cs" LinkBase="Common\Asserts" />
     <Compile Include="..\xunit.assert\Asserts\Sdk\AssertEqualityComparer.cs" LinkBase="Common\Asserts" />
     <Compile Include="..\xunit.assert\Asserts\Sdk\AssertEqualityComparerAdapter.cs" LinkBase="Common\Asserts" />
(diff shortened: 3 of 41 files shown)


bradwilson:
I don't think reusing the `methodDisplay` value is actually appropriate because it's not a flags enum (it's one or the other). I think using separate formatting flags makes more sense. I think what I'd want to see is something like:

`methodDisplay` = choose one of `Method`, `ClassAndMethod`, `Pretty`
`prettyMethodFlags` = flag enum of pretty print options, including a value `All` which is the default

commonsensesoftware:
I can definitely see your viewpoint. I thought about it for a quite a while before deciding to switch the enumeration to support flags. Luck would have it that changing the enumeration to support flags had no change to the existing values or their meaning.

Splitting out the _pretty_ display options makes sense, but it still poses a problem unless we were to add two options because ```ClassAndMethod | Pretty``` and ```Method | Pretty``` produce two distinctly different combinations. The enumeration _could_ be changed to:

```c#
public enum TestMethodDisplay
{
    ClassAndMethod = 1,
    Method,
    PrettyClassAndMethod,
    PrettyMethod
}
```

The _pretty_ method display options can be then be broken out into a separate enumeration. Selecting one of the new _pretty_ options enables the feature, including the existing support the full class name or just method name.

I also had the goal of trying to avoid introducing a new configuration value. If an xUnit user opts into _pretty_ display options, is it even worth having configuration options? I wanted users to have some level of control on what _could_ be used, but it may honestly be over-configuration. In using the fork I created, my usage pattern has been to always turn on all the pretty display features and then only use the ones I want.

What's your option? Do you think it would be better to just ditch the _pretty_ display enumeration, enable all the features, and users just decide when or if to use particular subfeatures after they opt-in with one of the **PrettyXXX** TestDisplayMethod options?

bradwilson:
Hmm, I'd assumed that pretty meant you only used the method name, but if it's the case that pretty applies to both Method and ClassAndMethod, then you don't really need the flag at all. We just set up the transforms as flags (on a separate config value), with the default being None instead of All. That way, whether you're using ClassAndMethod or Method, you enable the name transformations by picking the flags.

Thoughts?

commonsensesoftware:
I think that's the right behavior. To avoid surprises to users, the current formatting behavior should be the same as it is today until users choose to enable them.

The name _Pretty_ is a misnomer. I didn't have better name when I was originally describing the feature. Since the feature is really just additional display formatting options, I've added a new enumeration with the **Options** suffix that is a bit vector. I thought this name would have better symmetry with the existing **TestMethodDisplay**.  What do you think?

```c#
[Flags]
public enum TestMethodDisplayOptions
{
    None = 0x00,
    ReplaceUnderscoreWithSpace = 0x01,
    UseOperatorMonikers = 0x02,
    UseEscapeSequences = 0x04,
    ReplacePeriodWithComma = 0x08,
    All = ReplaceUnderscoreWithSpace | UseOperatorMonikers |
          UseEscapeSequences | ReplacePeriodWithComma
}
```

The **ReplacePeriodWithComma** is a new option from the last iteration to support the period replace in test class names. This option will only be honored when combined with ```TestMethodDisplay.ClassAndMethod``` because there is nothing to format with ```TestMethodDisplay.Method``` since method names cannot have periods in them.

This will enable a user to write:

```c#
namespace given_a_placed_order
{
	public class when_the_order_is_fulfilled
	{
		[Fact]
		public void then_the_ordered_items_should_be_delivered() { }
	}
}
```

Which will output:

```given a placed_order, when the order is fulfilled, then the ordered items should be delivered```

In terms of the configuration settings, I've added the following:

```xml
<configuration>
 <appSettings>
   <add key="xunit.methodDisplay" value="method" />
   <add key="xunit.methodDisplayOptions" value="all" />
 </appSettings>
</configuration
```

```json
{
 "methodDisplay": "method",
 "methodDisplayOptions": "all"
}
```

I am a little concerned about the number of changes caused by introducing a new configuration setting. Effectively, all the places where **TestMethodDisplay** was used, parsed, or passed around, I needed to update it with the companion support for **TestMethodDisplayOptions**. I supplied default values to follow the current implementation for **TestMethodDisplay**, but it still changes the signature of several constructors and methods. Isn't that a breaking change for users that write their own SDK extensions?

I've taken all the points we've discussed thus far, incorporated them, updated the corresponding tests, and verified all the tests still pass.  Once you're satisfied with the names, behaviors, and direction, I can send an updated PR.

bradwilson:

commonsensesoftware:
I've updated the PR based on previous feedback. I assume the runner would have to also be updated to take advantage of the new configuration setting.  I'm sure you're up to your eyeballs with other stuff. Let me know any additional thoughts you have when you eventually have time to review and reply.

commonsensesoftware:
Any additional thoughts? I'm sure you've been very busy with far more pressing features and I can only imagine that there has been a lot of pressure to get 2.2.0 out. Let's not forget that you still have day job too! Congrats!

I did eventually figure out why the enumeration only parses correctly using the original approach. The runner doesn't have the updated reference to the utility library. So while the configuration setting exists and can be read, it will only be processed when the value is numeric because the string names aren't present in the version of TestMethodDisplay referenced in the utility library.

That said, I was really hoping to make the cut for 2.2.0 since the direction was to add a configuration setting (TestMethodDisplayOptions) and the runner has to have a reference to the version of the utility library that understands how to read the new setting.

I didn't think there was anything else you were waiting on from me. If there is, please let me know and I'm happy to address those items straight away.

gozineha:
آزمون آنلاین فنی حرفه ای گزینه ها شامل تمامی نمونه سوالات فنی و حرفه ای می باشد. این آزمون ها به گونه ای طراحی شده اند که دقیقا مشابه با آزمون اصلی فنی و حرفه ای هستند. پس از شرکت در این آزمون ها شما با نمونه سوالات بسیاری آشنا خواهید شد که به شما در امتحانات اصلی فنی حرفه ای بسیار کمک می کند.
**[آزمون آنلاین فنی حرفه ای](http://gozineha.ir/school/secondary/technical/online-test-vocational.html)**

damianh:
Would be nice to see this land.

commonsensesoftware:
I thought I had largely followed the checklist:

- [ ] Ensured it was an accepted _'help wanted'_ issue
- [x] Read and followed the project governance
- [x] Signed the CLA agreement
- [x] Followed the existing coding styles
- [x] Wrote unit tests

The original issue never had any traction. Rather than wait indefinitely, I created this PR given how simple and small the code was. That eventually got some traction, but it's still not an accepted feature. On the positive side though, @bradwilson could have flat out said _"No"_ a **long** time ago. I suspect he's just been too busy with other aspects of the framework and hasn't gotten around to fully evaluating whether he wants this to be formally part of xUnit; I can appreciate that.

In the meantime, I've reluctantly been maintaining a fork which enables this feature. You can find the packages at:

* [more.xunit](https://www.nuget.org/packages/more.xunit/)
* [more.xunit.runner.visualstudio](https://www.nuget.org/packages/more.xunit.runner.visualstudio/)

The **only** difference in those packages is the support for the new configuration setting and test case display formatting. If the feature is ever formally added to xUnit, I'll shut down the fork.

@bradwilson, there's been about 1K downloads of this variant package over the past 6 weeks alone (I can send you the NuGet stats if necessary). It seems the community does want it, should you still be up in the air about it. Until then, I'll keep this PR up-to-date in hopes that it will one day be accepted.

bradwilson:
Sorry, this never raised high enough up on my radar for 2.3. I'll make sure to take another look for 2.4.

bradwilson:
I had to make a few changes, and this PR pre-dates the Github feature that lets me push directly to your branch, so I did them locally and pushed up the result. Everything has been merged.

Thanks again! This is an excellent job.

Issue #24, opened 2014-03-22
MSBuild runner: Add support for stopping on first test failure


bradwilson:

bradwilson:
Implemented.

Issue #23, opened 2014-03-22
Console runner: Add support for stopping on first test failure


bradwilson:

bradwilson:
Implemented

chadly:
How does one use this feature?

chadly:
`dotnet xunit -stoponfail`

PR #831, opened 2016-04-22
Add support for stopping runners after first test failure

EMarshal:
Uses the message bus to add support to the console runner (issue #23) and the MSBuild runner (issue #24) for stopping on the first test failure. Assemblies with test failures will stop immediately, running assemblies will continue until complete, and no further assemblies will be executed.

### DIFF
--- a/src/common/TestOptionsNames.cs
+++ b/src/common/TestOptionsNames.cs
@@ -11,6 +11,7 @@ static class TestOptionsNames
     internal static class Execution
     {
         public static readonly string DiagnosticMessages = "xunit.execution.DiagnosticMessages";
+        public static readonly string StopOnFail = "xunit.execution.StopOnFail";
         public static readonly string DisableParallelization = "xunit.execution.DisableParallelization";
         public static readonly string MaxParallelThreads = "xunit.execution.MaxParallelThreads";
         public static readonly string SynchronousMessageReporting = "xunit.execution.SynchronousMessageReporting";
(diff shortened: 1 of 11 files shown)


tgiphil:
+1

I'd like to see an option to stop after a configurable number of failures.

bradwilson:
Is the intention of this feature to:

1. Stop running tests in an assembly when it has a failure, but continue running the other assemblies
2. Stop running tests in all assemblies whenever a single failure it hit

It appears that this code does option 1, but frankly I'd expected option 2.

bradwilson:

ErikApption:
Any update on this issue? This option would save a significant built time when there is an early failure

EricJFisher:
Since this has been stuck at `cla-required` for over 7 months what is the appropriate action to see this through? I am in need of Option 2, should I reimplement this with Option 2 and submit a new PR or?

marcind:
Given how there's merge conflicts now in this PR, a new PR with option 2 would be the fastest way to go.

EMarshal:
I kinda forgot about this - I could take another look to clean it up, probably be a week or two before I get to it though. If you want to look at it before then, go for it.

EricJFisher:
I'll try to give it a look over this weekend. I'll also take the time to setup Option 2, thanks for responding :)

EMarshal:
Reworked and merged as part of #1320.

Issue #841, opened 2016-05-15
Huge amount of exceptions thrown when xunit is processing custom attributes.


CyrusNajmabadi:
In https://github.com/dotnet/roslyn we make extensive use of custom attributes to flexibly control which tests should run.  For example, in https://github.com/dotnet/roslyn/blob/master/src/VisualStudio/Core/Test/CodeModel/CSharp/CodeFunctionTests.vb you can see attributes like the following:

```vb
<ConditionalWpfFact(GetType(x86)), Trait(Traits.Feature, Traits.Features.CodeModel)>
Public Async Function TestAccess1() As Task
```

These attributes cause an enormous amount of exceptions to be thrown in the unit discovery phase.  Specifically with this callstack:

```
 	mscorlib.dll!System.Convert.ChangeType(value, conversionType, provider)	Unknown
 	mscorlib.dll!System.Convert.ChangeType(value, conversionType)	Unknown
>	xunit.execution.desktop.dll!Xunit.Sdk.Reflector.ConvertArgument(arg, type)	C#
 	xunit.execution.desktop.dll!Xunit.Sdk.Reflector.ConvertArguments(args, types)	C#
 	xunit.execution.desktop.dll!Xunit.Sdk.ReflectionAttributeInfo.Instantiate(attributeData)	C#
 	xunit.execution.desktop.dll!Xunit.Sdk.ReflectionAttributeInfo.ReflectionAttributeInfo(attribute)	C#
 	xunit.execution.desktop.dll!Xunit.Sdk.ReflectionMethodInfo.GetCustomAttributes(method, attributeType, attributeUsage)	C#
 	xunit.execution.desktop.dll!Xunit.Sdk.ReflectionMethodInfo.GetCustomAttributes(method, assemblyQualifiedAttributeTypeName)	C#
 	xunit.execution.desktop.dll!Xunit.Sdk.ReflectionMethodInfo.GetCustomAttributes(assemblyQualifiedAttributeTypeName)	C#
 	xunit.execution.desktop.dll!Xunit.Sdk.XunitTestFrameworkDiscoverer.FindTestsForMethod(testMethod, includeSourceInformation, messageBus, discoveryOptions)	C#
 	xunit.execution.desktop.dll!Xunit.Sdk.XunitTestFrameworkDiscoverer.FindTestsForType(testClass, includeSourceInformation, messageBus, discoveryOptions)	C#
[...]

PR #842, opened 2016-05-15
Do not attempt to unnecessarily convert types which are compatible.

CyrusNajmabadi:
Fixes https://github.com/xunit/xunit/issues/841

In cases where there are assignable types passed in, this could would still attempt to convert them by calling Convert.ChangeType.  This method throws when 'arg' is not IConvertible.

### DIFF
--- a/src/xunit.execution/Sdk/Reflection/Reflector.cs
+++ b/src/xunit.execution/Sdk/Reflection/Reflector.cs
@@ -44,7 +44,7 @@ namespace Xunit.Sdk
 
         internal static object ConvertArgument(object arg, Type type)
         {
-            if (arg != null && arg.GetType() != type)
+            if (arg != null && !type.IsAssignableFrom(arg.GetType()))
             {
                 try
                 {


CyrusNajmabadi:
An alternative approach would be to change the code to be like this:

```c#
                    if (type.IsArray)
                    {
                        var elementType = type.GetElementType();
                        var enumerable = (IEnumerable<object>)arg;
                        var castMethod = EnumerableCast.MakeGenericMethod(elementType);
                        var toArrayMethod = EnumerableToArray.MakeGenericMethod(elementType);
                        return toArrayMethod.Invoke(null, new object[] { castMethod.Invoke(null, new object[] { enumerable }) });
                    }
                    else if (arg is IConvertible)  //<-- Test before calling something that will immediately throw.
                    {
                        return Convert.ChangeType(arg, type);
                    }
```

CyrusNajmabadi:
@bradwilson Is this something you can take a look at?  Thanks!

bradwilson:
Merged, thanks!

PR #870, opened 2016-06-04
Fix async fixture issues in #869 

### LINKED ISSUE #869: IAsyncLifetime fixtures are potentially initialised more than once
If a test class includes more than one IAsyncLifetime implementing class fixtures, all existing class fixtures will have their ```InitializeAsync()``` calls re-invoked, and on test clean-up the number of calls to ```InitializeAsync()``` won't match the number of calls to ```DisposeAsync()```.

For example, if a class includes three class fixtures ```AlphaFixture```, ```BetaFixture```, and ```GammaFixture```. ```AlphaFixture``` will be invoked 3 times, ```BetaFixture``` twice, and ```GammaFixture``` once.

Collection fixtures are also affected.

Here is a test reproducing the bug: 

```cs
using System;
using System.Threading.Tasks;
using Xunit;

namespace Xunit.AsyncFixture.Bug
{
    public class AsyncFixture<T> : IAsyncLifetime
    {
        public int TimesInitialised { get; set; } = 0;

        public Task InitializeAsync()
        {
            ++TimesInitialised;
            return Task.CompletedTask;
        }

        public Task DisposeAsync()
        {
            return Task.CompletedTask;
        }
    }

    public class Alpha { }
    public class Beta { }
    public class Gamma { }

    /// <remarks>
    /// This test asserts that each class fixture should only have its InitializeAsync() function called once.
    /// </remarks>
    public class AsyncBugTests : IClassFixture<AsyncFixture<Alpha>>, IClassFixture<AsyncFixture<Beta>>, IClassFixture<AsyncFixture<Gamma>>
    {
        public AsyncBugTests(AsyncFixture<Alpha> alpha, AsyncFixture<Beta> beta, AsyncFixture<Gamma> gamma)
        {
            Assert.Equal(1, alpha.TimesInitialised);
            Assert.Equal(1, beta.TimesInitialised);
            Assert.Equal(1, gamma.TimesInitialised);
        }

        [Fact]
        public void Test()
        {
        }
    }
}
```

A workaround for this issue is to create a composite test fixture that initialises each fixture only once. Another workaround would be to ignore repeated calls to ```InitializeAsync()``` in the fixture.

A second potential bug is that async test fixtures are initialised serially and not in parallel, reducing the benefit of supporting async fixtures.

I would expect following test to only take 2 seconds to execute but due to the combination of both bugs takes 5 seconds.

```cs
using System;
using System.Threading.Tasks;
using Xunit;

namespace Xunit.Async.TimingBug
{
    public class AsyncFixture<T> : IAsyncLifetime
    {
        public async Task InitializeAsync()
        {
            await Task.Delay(1000);
        }

        public async Task DisposeAsync()
        {
            await Task.Delay(1000);
        }
    }

    public class Alpha { }
    public class Beta { }

    public class AsyncBugTests : IClassFixture<AsyncFixture<Alpha>>, IClassFixture<AsyncFixture<Beta>>
    {
        [Fact]
        public void Test()
        {
        }
    }
}
```

onovotny:
PR to fix #869 

- [x]  Do init/dispose in parallel
- [ ]  Prevent multiple invocations of init

### DIFF
--- a/src/xunit.execution/Sdk/Frameworks/Runners/XunitTestClassRunner.cs
+++ b/src/xunit.execution/Sdk/Frameworks/Runners/XunitTestClassRunner.cs
@@ -88,9 +88,11 @@ namespace Xunit.Sdk
         async Task CreateClassFixtureAsync(Type fixtureType)
         {
             CreateClassFixture(fixtureType);
-            foreach(var uninitializedFixture in ClassFixtureMappings.Values.OfType<IAsyncLifetime>())
-                await Aggregator.RunAsync(uninitializedFixture.InitializeAsync);
-            
+
+            await Task.WhenAll(
+                ClassFixtureMappings.Values
+                .OfType<IAsyncLifetime>()
+                .Select(uninitializedFixture => Aggregator.RunAsync(uninitializedFixture.InitializeAsync)));
         }
 
         /// <inheritdoc/>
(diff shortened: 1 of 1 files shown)


Sumo-MBryant:
Looks like you beat me to it, feel free to take any of the changes from my PR. They include test updates, a fix for the multiple invocations issue, and fixes to both class fixtures and collections.

onovotny:
Closing this in favor of #871 since it's more complete

PR #871, opened 2016-06-04
Stop repeated initialization of async fixtures

### LINKED ISSUE #869: IAsyncLifetime fixtures are potentially initialised more than once

Sumo-MBryant:
Fixes #869

IAsyncLifetime fixtures both class fixtures and in collections are only
initialized once.

All async initialization and disposal is now executed concurrently.

### DIFF
--- a/src/xunit.execution/Sdk/Frameworks/Runners/XunitTestClassRunner.cs
+++ b/src/xunit.execution/Sdk/Frameworks/Runners/XunitTestClassRunner.cs
@@ -47,6 +47,11 @@ namespace Xunit.Sdk
         protected Dictionary<Type, object> ClassFixtureMappings { get; set; } = new Dictionary<Type, object>();
 
         /// <summary>
+        /// Gets the already initialized async fixtures <see cref="CreateClassFixtureAsync"/>.
+        /// </summary>
+        protected HashSet<IAsyncLifetime> InitializedAsyncFixtures { get; set; } = new HashSet<IAsyncLifetime>();
+
+        /// <summary>
         /// Creates the instance of a class fixture type to be used by the test class. If the fixture can be created,
         /// it should be placed into the <see cref="ClassFixtureMappings"/> dictionary; if it cannot, then the method
         /// should record the error by calling <code>Aggregator.Add</code>.
@@ -88,9 +93,12 @@ namespace Xunit.Sdk
         async Task CreateClassFixtureAsync(Type fixtureType)
         {
             CreateClassFixture(fixtureType);
-            foreach(var uninitializedFixture in ClassFixtureMappings.Values.OfType<IAsyncLifetime>())
-                await Aggregator.RunAsync(uninitializedFixture.InitializeAsync);
-            
+            foreach (var uninitializedFixture in ClassFixtureMappings.Values.OfType<IAsyncLifetime>().ToList())
+                if (!InitializedAsyncFixtures.Contains(uninitializedFixture))
+                {
+                    InitializedAsyncFixtures.Add(uninitializedFixture);
+                    await Aggregator.RunAsync(uninitializedFixture.InitializeAsync);
+                }
         }
 
         /// <inheritdoc/>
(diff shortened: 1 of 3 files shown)


src/xunit.execution/Sdk/Frameworks/Runners/XunitTestClassRunner.cs
```diff
@@ -88,9 +93,12 @@ protected virtual void CreateClassFixture(Type fixtureType)
         async Task CreateClassFixtureAsync(Type fixtureType)
         {
             CreateClassFixture(fixtureType);
-            foreach(var uninitializedFixture in ClassFixtureMappings.Values.OfType<IAsyncLifetime>())
-                await Aggregator.RunAsync(uninitializedFixture.InitializeAsync);
-            
+            foreach (var uninitializedFixture in ClassFixtureMappings.Values.OfType<IAsyncLifetime>().ToList())
+                if (!InitializedAsyncFixtures.Contains(uninitializedFixture))
```
onovotny: Please add braces around this since it's more than one line.

src/xunit.execution/Sdk/Frameworks/Runners/XunitTestClassRunner.cs
```diff
@@ -88,9 +93,12 @@ protected virtual void CreateClassFixture(Type fixtureType)
...
+                {
+                    InitializedAsyncFixtures.Add(uninitializedFixture);
+                    await Aggregator.RunAsync(uninitializedFixture.InitializeAsync);
```
onovotny: Should we be awaiting here or using a `Task.WhenAll` outside of the loop? Seems like things may still happen serially here?

onovotny:
LGTM outside of the two notes I added. @bradwilson?

As a more general question, should the `await`'s in this file use `ConfigureAwait(false)`? Doesn't seem like any of them do currently.

src/xunit.execution/Sdk/Frameworks/Runners/XunitTestClassRunner.cs
Sumo-MBryant: Thank you for the feedback. I have rewritten the function to adopt the changes from #870 and include the fix for repeated initialisation.

bradwilson:
Thanks!

Issue #472, opened 2015-07-19
Get rid of TestMessageVisitor because the perf sucks


bradwilson:
Across app domains, the roughly 40 casts for dispatching can be pretty punitive. This is especially true for discovery, which fundamentally only has 2 message types.

bradwilson:
Resolved with #930.

PR #881, opened 2016-06-16
Microoptimize TestDiscoveryVisitor and DiagnosticMessageVisitor

TimLovellSmith:
...by overriding implementation in DiagnosticMessageVisitor.OnMessage, and not using TestMessageVisitor<T> base base class for TestDiscoveryVisitor.

This seems to shave half a second off some of my test runs, but YMMV - and I find measuring performance hard.

Partially addresses #472.

### DIFF
--- a/src/common/TestDiscoveryVisitor.cs
+++ b/src/common/TestDiscoveryVisitor.cs
@@ -1,4 +1,6 @@
-﻿using System.Collections.Generic;
+﻿using System;
+using System.Collections.Generic;
+using System.Threading;
 using Xunit.Abstractions;
 
 #if XUNIT_FRAMEWORK
@@ -7,19 +9,55 @@ namespace Xunit.Sdk
 namespace Xunit
 #endif
 {
-    class TestDiscoveryVisitor : TestMessageVisitor<IDiscoveryCompleteMessage>
+    class TestDiscoveryVisitor : LongLivedMarshalByRefObject, IMessageSink, IDisposable
     {
         public TestDiscoveryVisitor()
         {
+            Finished = new ManualResetEvent(false);
             TestCases = new List<ITestCase>();
+        }
+
+        /// <summary>
+        /// This event is triggered when the completion message has been seen.
+        /// </summary>
+        public ManualResetEvent Finished { get; private set; }
+
+        /// <inheritdoc/>
+        public void Dispose()
+        {
+            ((IDisposable)Finished).Dispose();
         }
 
         public List<ITestCase> TestCases { get; private set; }
 
-        protected override bool Visit(ITestCaseDiscoveryMessage testCaseDiscovered)
+        /// <inheritdoc/>
+        public virtual bool OnMessage(IMessageSinkMessage message)
+        {
+            var discoveryMessage = message as ITestCaseDiscoveryMessage;
+            if (discoveryMessage != null)
+            {
+                return Visit(discoveryMessage);
+            }
+
+            var completeMessage = message as IDiscoveryCompleteMessage;
+            if (message is IDiscoveryCompleteMessage)
+            {
+                var result = Visit(completeMessage);
+                Finished.Set();
+                return result;
+            }
+
+            return true;
+        }
+
+        protected virtual bool Visit(IDiscoveryCompleteMessage completeMessage)
+        {
+            return true;
+        }
+
+        protected virtual bool Visit(ITestCaseDiscoveryMessage testCaseDiscovered)
         {
             TestCases.Add(testCaseDiscovered.TestCase);
-
             return true;
         }
     }
--- a/src/xunit.console/Visitors/DiagnosticMessageVisitor.cs
+++ b/src/xunit.console/Visitors/DiagnosticMessageVisitor.cs
@@ -18,21 +18,35 @@ namespace Xunit.ConsoleClient
             this.noColor = noColor;
         }
 
-        protected override bool Visit(IDiagnosticMessage diagnosticMessage)
+        /// <inheritdoc/>
+        public override bool OnMessage(IMessageSinkMessage message)
         {
             if (showDiagnostics)
-                lock (consoleLock)
+            {
+                var diagnosticMessage = message as IDiagnosticMessage;
+                if (diagnosticMessage != null)
                 {
-                    if (!noColor)
-                        Console.ForegroundColor = ConsoleColor.Yellow;
-
-                    Console.WriteLine($"   {assemblyDisplayName}: {diagnosticMessage.Message}");
-
-                    if (!noColor)
-                        Console.ResetColor();
+                    return Visit(diagnosticMessage);
                 }
+            }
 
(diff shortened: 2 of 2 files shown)


src/xunit.console/Visitors/DiagnosticMessageVisitor.cs
```diff
@@ -18,21 +18,35 @@ public DiagnosticMessageVisitor(object consoleLock, string assemblyDisplayName,
             this.noColor = noColor;
         }
 
-        protected override bool Visit(IDiagnosticMessage diagnosticMessage)
+        /// <inheritdoc/>
+        public override bool OnMessage(IMessageSinkMessage message)
         {
             if (showDiagnostics)
```
TimLovellSmith: Note: I moved casting and visitation inside of the if (showDiagnostics) test. This might change observed behavior/bugs for someone... let alone from the fact that subclasses will no longer see any messages except IDiagnosticMessages.

src/common/TestDiscoveryVisitor.cs
```diff
@@ -7,20 +9,56 @@ namespace Xunit.Sdk
 namespace Xunit
 #endif
 {
-    class TestDiscoveryVisitor : TestMessageVisitor<IDiscoveryCompleteMessage>
+    class TestDiscoveryVisitor : LongLivedMarshalByRefObject, IMessageSink, IDisposable
```
TimLovellSmith: Could break someone: no longer inherits TestMessageVisitor & TestMessageVisitor<T>.

bradwilson:
I'm of a split personality on how I want to fix this.

Part of me wants to make a 3.0 branch that just keeps everything from xunit.abstractions on the far side of the app domain, and go back to use remoted strings like in 1.x (though different; not XML).

Also, have you looked at [AssemblyRunner](https://github.com/xunit/xunit/blob/master/src/xunit.runner.utility/Runners/AssemblyRunner.cs)? This is an alternative that I have been toying with that has significantly better performance characteristics.

bradwilson:
Build this sample and use it to run your tests, and see how the perf is: https://github.com/xunit/samples.xunit/tree/master/TestRunner

TimLovellSmith:
@bradwilson TBH I think going back to remoted strings [for at least most events] would be an improvement for 3.0 - sending every event as a remoting proxy seems pretty wild. I haven't looked at AssemblyRunner yet. Will try the TestRunner thing...

TimLovellSmith:
@bradwilson Just by using TestRunner/XUnit without any changes, I found that TestRunner runs my favorite test assembly in 22.2 seconds, MSBuild 23.9 seconds. That seems like a probably significant difference even after just 1 sample, but of course they are not necessarily equivalent in features/configuration.

The CPU sampling profiles look quite different too. I think something changed about my profiling, so I see "NullableTypeOfByRef" instead that other JIT helper method. Anyway, with TestRunner a good chunk of time (7-8% excl) was spent in 'NullableTypeOfByRef' which is being called by ... `AssemblyRunner.DispatchMessage`. So it has a perf hit from its dynamic dispatching of message sink messages too - just not as bad as TestMessageVisitor.

Also noticing that only about half as many samples (5%) go to TransparentProxyStub_CrossContext, which is another place where things are improved, but maybe could be even better.

bradwilson:
Yeah, nothing compares to going back to strings (or, better yet, turning off app domains), so that's probably where I will focus to see how much better the results can be. I'm hoping to be able to achieve it without breaking the extensibility (just moving some of the running bits over into the other app domain, only changing/updating the contract of runner.utility).

ilmax:
What about implementing a fully flagged visitor pattern? The performance should be better because you can get rid of casts, but we loose the ability of implement two message type with a single class.

bradwilson:
I'm not sure how you do this without casts. Can you give a quick example?

ilmax:
@bradwilson please note that I have not much knowledge of the xunit code, so my suggestion can be totally wrong here.

pseudo code

```csharp
class TestDiscoveryVisitor : ITestDiscoveryVisitor
{
	protected virtual bool Visit(IDiscoveryCompleteMessage completeMessage) {...}

	protected virtual bool Visit(ITestCaseDiscoveryMessage testCaseDiscovered) {...}
	
	....
}

public interface ITestCaseDiscoveryMessage
{
	....
	bool Accept(ITestDiscoveryVisitor visitor);
}

public interface IDiscoveryCompleteMessage
{
	....
	bool Accept(ITestDiscoveryVisitor visitor);
}

class TestCaseDiscoveryMessage : ITestCaseDiscoveryMessage
{
	public bool ITestCaseDiscoveryMessage(ITestDiscoveryVisitor visitor)
	{
		return visitor.Visit(this);
	}
}

class DiscoveryCompleteMessage : IDiscoveryCompleteMessage
{
	public bool ITestCaseDiscoveryMessage(ITestDiscoveryVisitor visitor)
	{
		return visitor.Visit(this);
	}
}
```

Essentially every "visited" class should accept a visitor in the `Accept` method and call visit passing itself as the method parameter.

This way we could avoid the cast, but as pointed out in the prev message, we loose the ability to implement two message type in a single class.

bradwilson:
Most of those message interfaces are defined in `xunit.abstractions`. 😞 

https://github.com/xunit/xunit/blob/master/src/abstractions/xunit.abstractions/Messages/ITestCaseDiscoveryMessage.cs

TimLovellSmith:
Am I right in thinking the goal for xunit.abstractions was - no changes til 3.0? Or is it no changes ever?

TimLovellSmith:
@bradwilson In general I'm curious how you see this kind of change fitting into the roadmap.

Suppose there may possible be clever ways to make this a nonbreaking change... like define ITestCaseDiscoveryMessage2... IMessageSink2? (Or is it IMessageVisitor?) etc and have the runtime choose the best one to use.

Is there any upside to doing an xunit 2.5 with this kind of change versus starting xunit 3.0?
Or would it make sense to try to do that kind of back compat as part of 3.0 also, so that xunit 3 stuff can work with xunit 2 stuff?

bradwilson:
Ideally, we would never change the abstractions again. Once we do, we create a breaking forward compatibility problem (where test runners linked against abstractions v2 cannot run tests that use abstractions v3); the point of freezing abstractions is to preserve that forward compatibility.

Regardless of the version number, I think it's best so stop using the abstractions across the app domain (even though that's what they were theoretically there for). Until I start digging in, though, I'm not sure how big a job it will be. At the moment I'm focused on getting a new beta out on the 27th to coincide with .NET Core 1.0. :)

bradwilson:
Closing this because of the recent move from visitors => sinks for macro performance improvements. :)

TimLovellSmith:
@bradwilson Woohoo!! Can't wait to try out macro performance improvements.

onovotny:
@TimLovellSmith they're nice...and include the new long-running test/deadlock detection if enabled via config. You can use them today by using the xUnit MyGet feed with the latest versions.

PR #911, opened 2016-07-16
Fix TeamCity Suite grouping (test count)

ianbattersby:
**This PR excludes the assembly id suffix and escapes as per TC reqs.**

It appears TeamCity uses the test-suite name to not only group test results but to derive the count. Unfortunately in TC 9.x it seems to take exception to the suffixed assembly attribute (issue [TW-46190](https://youtrack.jetbrains.com/issue/TW-46190)).

For example;

```
##teamcity[testSuiteStarted name='Test collection for MyTestAssembly.Handlers.RecordCredentials.CommandHandlerTests (2)' …]
```

vs;

```
##teamcity[testSuiteStarted name='Test collection for MyTestAssembly.Handlers.RecordCredentials.CommandHandlerTests' …]
```

In addition the output DisplayName was not escaped as per: https://confluence.jetbrains.com/display/TCD8/Build+Script+Interaction+with+TeamCity

### DIFF
--- a/src/xunit.runner.reporters/Utility/TeamCityDisplayNameFormatter.cs
+++ b/src/xunit.runner.reporters/Utility/TeamCityDisplayNameFormatter.cs
@@ -22,12 +22,24 @@ namespace Xunit.Runner.Reporters
                 }
             }
 
-            return $"{testCollection.DisplayName} ({id})";
+            return EscapeValue($"{testCollection.DisplayName}");
         }
 
         public virtual string DisplayName(ITest test)
         {
             return test.DisplayName;
+        }
+
+        static string EscapeValue(string value)
+        {
+            value = value.Replace("|", "||");
+            value = value.Replace("'", "|'");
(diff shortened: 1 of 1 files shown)


bradwilson:
The problem here is that TC9 broke test counts when test suites are not uniquely named (they grouped the tests together). This change would undo that work.

I'm reluctant to accept this fix without further understanding of how it affects TeamCity test counts.

bradwilson:

bradwilson:
@ianbattersby Wondering if you had some feedback on my previous comment.

ianbattersby:
Hi @bradwilson, apologies - I'd missed this. The project concerned where we saw this isn't in my immediate reach at the moment but I can see on GHE it's still using a custom build of xUnit for this reason. Let me close this for now and then re-open if it's still a problem once lifting xUnit 2.3 release and when running on TC 2017.1.3 :+1:

PR #915, opened 2016-07-16
Alter NUnitXml.xslt for TC compatibility

ianbattersby:
Importing NUnit generated output into TeamCity using the XML processing build feature can cause tests to be included twice if the hierarchy is different from than output during the build (using ##teamcity prefix).

** This commit simplifies the output of the NUnit xml report so that it matches the build output.**

_NB: Alternatively we could (a) have a different version just for TC or change the build-log output to have the same hierarchy.

### DIFF
--- a/src/xunit.console/NUnitXml.xslt
+++ b/src/xunit.console/NUnitXml.xslt
@@ -32,45 +32,12 @@
         </xsl:attribute>
       </environment>
       <culture-info current-culture="unknown" current-uiculture="unknown" />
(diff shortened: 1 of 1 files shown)


bradwilson:
The intention of this file was never to provide TeamCity compatibility, as we already provide that out of the box.

ianbattersby:
Noted. I noticed ##[stdout...] covered the test output but didn't know if the TC parser would take it all. Will close this and if someone else wants to import into TC they can find this version here :)

Thanks!

Issue #920, opened 2016-07-22
Update xunit.abstractions to Dotnet Core 1.0 RTM.


jp7677:
Hi there,

I'm using the following packages in my Dotnet Core 1.0 RTM  tests project 
```
"xunit": "2.2.0-beta2-build3300",
"dotnet-test-xunit": "2.2.0-preview2-build1029",
```
It works perfectly, but unfortunately nuget still pulls some rc2 packages. The reason is imho that xunit.abstraction still refers to rc2 (https://github.com/xunit/xunit/blob/master/src/abstractions/xunit.abstractions/project.json). Could you please update these to RTM packages? Would be nice to get rid of all pre-RTM stuff ;)

bradwilson:
Change pushed here: 2671cc4

jp7677:
Thanks a lot!

PR #923, opened 2016-07-27
Update xunit.abstractions to Dotnet Core 1.0 RTM.

jp7677:
Though not sure if the buildsystem applies a proper version to xunit.abstractions. Please let me know if I need to modify some more lines to get rid of RC2 packages. This pull request should resolve https://github.com/xunit/xunit/issues/920.

### DIFF
--- a/src/abstractions/xunit.abstractions.nuspec
+++ b/src/abstractions/xunit.abstractions.nuspec
@@ -17,10 +17,10 @@
       <group targetFramework="net35" />
       <group targetFramework="portable-net45+netcore45+wp8+wpa81" />
       <group targetFramework="netstandard1.0">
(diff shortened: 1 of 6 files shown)


jp7677:
Ok, seems ilke TeamCity doesn't yet applies a version number to xunit.abstractions :(

jp7677:
mmh, guess this is not the correct approach, should I close this PR?

bradwilson:
We don't auto-build abstractions, because we're continuing to use the 2.0 DLL. When I ship the 2.0.1 "RTM", I will update the references to use the single `<dependency id="NETStandard.Library" version="1.6.0" />` as the .NET team recommends now.

Thanks!

bradwilson:
Change pushed here: 2671cc4

PR #924, opened 2016-07-27
Provide source information for referenced assemblies, fixing test categorisation

gulbanana:
In the Visual Studio runner, test source information is provided using `VisualStudioSourceInformationProvider`. VS uses this information to group tests by project, to jump to the source of tests and to run tests from a context menu. However, the provider is returning null for inherited test cases in referenced projects. These tests show up as category External" and IDE features don't work for them. The problem can be seen in the following solution:
https://github.com/gulbanana/xunit-repro

This PR updates the DiaSession wrapper to inspect the assembly associated with individual test class types instead of assuming all test cases are in the same assembly. It shouldn't have performance implications as it will only produce one extra session wrapper per assembly that contains discovered tests.

It wasn't possible to write unit tests for this change, here's a companion PR with integration testing: [#6](https://github.com/xunit/xunit.integration/pull/6).

### DIFF
--- a/src/xunit.runner.utility/Utility/DiaSession.cs
+++ b/src/xunit.runner.utility/Utility/DiaSession.cs
@@ -1,6 +1,7 @@
 ﻿#if !PLATFORM_DOTNET
 
 using System;
+using System.Collections.Generic;
 using System.Reflection;
 
 namespace Xunit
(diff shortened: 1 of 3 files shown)


bradwilson:
Thanks!

PR #929, opened 2016-08-03
Added lock around TestOutputHelper getter.

tdupont750:
Tests with async logging can cause a IndexOutOfBoundsException because the TestOutputHelper.Output getter does not lock the internal StringBuilder. This commit locks the getter with the pre-existing lockObject.


bradwilson:
Thanks! :)

PR #930, opened 2016-08-04
Optimize performance of message dispatching

marcind:


jcansdale:
I've head good things about this PR. Curious where the perf gains mostly came from? :smile:

marcind:
1. Attempting to cast a marshalled object to a specific interface is expensive. The old implementation received a message and then tried a cast using 30ish different interfaces. New code passes a list of implemented interfaces into the visitor, which it can use to do a "type check" before attempting the cast. This has the largest benefit when AppDomains are used, but even without them the perf gets better.
2. The old code implemented default no-op handlers for every type of message. New code switches to events, thus being able to detect if the visitor cares about a particular type of message. The cast in 1) is only attempted if there's a handler that will react to the message.

marcind:
Also, this addresses #472

jcansdale:
Interesting, thanks for the explanation! :rocket:

bradwilson:
Merged, thanks!

PR #941, opened 2016-08-16
Long Running Test detection for running tests

onovotny:


bradwilson:
Updated and merged, thanks!

There will be further work that stems off of this, because it reveals a current limitation with the sink system (the assumption that all messages are valid in all contexts), as illustrated by my need to feed the diagnostic message sink entirely through the class hierarchy. This makes me believe that long running test detection is actually a Russian Doll style sink that exists to the side, rather than in the mainline.

PR #949, opened 2016-08-27
Performance tweaks for test discovery

marcind:
Using a profiler I found a number of inefficient reflection look-ups which could benefit from some caching.
In addition make `TestFrameworkOptions` serializable to avoid the cost of repeated value lookups across AppDomain boundaries.


bradwilson:
Merged, thanks!

PR #965, opened 2016-09-05
Add support for embedded generic types, generic arrays and ref types in test methods

### LINKED ISSUE #805: Cannot use a Theory accepting a generic array
Consider this test case:
````CSharp
[Theory]
[InlineData(new int[] { 1, 2, 3 })]
public static void Test<T>(T[] array)
{
    foreach (var i in array)
    {
        Console.WriteLine(i);
    }
}
````
An exception occurs when the test case is run. Here is the stack trace:
````
ClassLibrary1.Class1.Test<Object>(array: [1, 2, 3]) [FAIL]
      System.ArgumentException : Object of type 'System.Int32[]' cannot be converted to type 'System.Object[]'.
      Stack Trace:
           at System.RuntimeType.TryChangeType(Object value, Binder binder, CultureInfo culture, Boolean needsSpecialCast)
           at System.RuntimeType.CheckValue(Object value, Binder binder, CultureInfo culture, BindingFlags invokeAttr)
           at System.Reflection.MethodBase.CheckArguments(Object[] parameters, Binder binder, BindingFlags invokeAttr, CultureInfo culture, Signature sig)
           at System.Reflection.RuntimeMethodInfo.InvokeArgumentsCheck(Object obj, BindingFlags invokeAttr, Binder binder, Object[] parameters, CultureInfo culture)
           at System.Reflection.RuntimeMethodInfo.Invoke(Object obj, BindingFlags invokeAttr, Binder binder, Object[] parameters, CultureInfo culture)
        --- End of stack trace from previous location where exception was thrown ---
           at System.Runtime.CompilerServices.TaskAwaiter.ThrowForNonSuccess(Task task)
           at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
        --- End of stack trace from previous location where exception was thrown ---
           at System.Runtime.CompilerServices.TaskAwaiter.ThrowForNonSuccess(Task task)
           at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
````

Based on the above, it seems that the wrong generic method is being created for the test method, i.e. `Test<Object>()` rather than `Test<Int32>()`.

Another thing to note: if a generic constrant is applied to the test method, i.e.:

````
public static void Test<T>(T[] array) where T : struct
````

this also fails, with a different exception: 

````
ClassLibrary1.Class1.Test [FAIL]
      System.ArgumentException : GenericArguments[0], 'System.Object', on 'Void Test[T](T[])' violates the constraint of type 'T'.
      ---- System.Security.VerificationException : Method ClassLibrary1.Class1.Test: type argument 'System.Object' violates the constraint of type parameter 'T'.
      Stack Trace:
           at System.RuntimeType.ValidateGenericArguments(MemberInfo definition, RuntimeType[] genericArguments, Exception e)
           at System.Reflection.RuntimeMethodInfo.MakeGenericMethod(Type[] methodInstantiation)
        ----- Inner Stack Trace -----
           at System.RuntimeMethodHandle.GetStubIfNeeded(RuntimeMethodHandleInternal method, RuntimeType declaringType, RuntimeType[] methodInstantiation)
           at System.Reflection.RuntimeMethodInfo.MakeGenericMethod(Type[] methodInstantiation
````
I believe this exception occurs when the test case is being generated. Presumably calling `MakeGenericMethod` with T=System.Object fails because it does not meet the constraints.

As another side note, the test case with the generic constraint cause the VS test runner to "run continuously" trying to discover the tests and seems to get stuck in a loop.

### LINKED ISSUE #946: [bug] Generic nullable type inference does not work correctly
## Repro

```cs
public class Class1
{
    [Theory]
    [MemberData(nameof(FooData))]
    public void Foo(int? nullable)
    {
    }

    [Theory]
    [MemberData(nameof(FooData))]
    public void GenericFoo<T>(T? nullable) where T : struct
    {
        Console.WriteLine(typeof(T));
    }

    public static IEnumerable<object[]> FooData()
    {
        yield return new object[] { new int?(3) };
    }
}
```

## Expected

`System.Int32` is printed to the console.

## Actual

xUnit recognizes `T` as `System.Object`, and since it does not match the constraint of `T :  struct` an exception is thrown.

```
System.ArgumentException : GenericArguments[0], 'System.Object', on 'Void GenericFoo[T](System.Nullable`1[T])' violates the constraint of type 'T'.
 ---- System.Security.VerificationException : Method BugLib2.Class1.GenericFoo: type argument 'System.Object' violates the constraint of type parameter 'T'.
 Stack Trace:
      at System.RuntimeType.ValidateGenericArguments(MemberInfo definition, RuntimeType[] genericArguments, Exception e)
      at System.Reflection.RuntimeMethodInfo.MakeGenericMethod(Type[] methodInstantiation)
   ----- Inner Stack Trace -----
      at System.RuntimeMethodHandle.GetStubIfNeeded(RuntimeMethodHandleInternal method, RuntimeType declaringType, RuntimeType[] methodInstantiation)
      at System.Reflection.RuntimeMethodInfo.MakeGenericMethod(Type[] methodInstantiation)
```

I believe this is due to the special way Nullable is handled when it is boxed/unboxed; when you box a `Nullable<T>` it actually becomes a boxed `T`, or null if it was null. For example:

```cs
int? nullable = 3;
object boxed = nullable;
Console.WriteLine(boxed is int); // True
object boxedNull = default(int?);
Console.WriteLine(boxedNull == null); // True
```

The changed type of the boxed item is probably causing some issues with the type inference algorithms for generic theories.

hughbe:
## Summary
XUnit can now infer the following tests and more (see tests):
```
// e.g. int?
public void GenericArrayTest<T>(T? value) where T : struct { }

// e.g. int[]
public void GenericArrayTest<T>(T[] value) { }

// e.g. int?[]
public void GenericArrayTest<T>(T?[] value) { }

// e.g. new List<int>()
public void EmbeddedGenericClass<T>(List<T> value) {  }

// e.g. new Dictionary<string, int>()
public void EmbeddedGenericClass<T, U>(Dictionary<T, U> value) {  }

// e..g new List<List<int>>()
public void EmbeddedEmbeddedGenericClass<T>(List<List<T>> value) {  }

// e..g new List<List<int?>()
public void EmbeddedEmbeddedGenericClass<T>(List<T?> value) where T : struct {  }

// e..g new List<int[]>()
public void EmbeddedArrayGenericClass<T>(List<T[]> value) {  }

// e.g. 5
public void GenericRefTest<T>(ref T value) { }

// e.g. new GenericClass3<GenericClass<bool>, GenericClass2<GenericClass3<ulong, long, int>, string>, uint>()
public void CrazyGenericMethod<T, U, V, W, X>(GenericClass3<GenericClass<T>, GenericClass2<GenericClass3<U, V, int>, string>, X> gen) { }
```

## Limitations
`HasElementType` and `GetElementType` are not defined on `ITypeInfo`.
So there were some options for the implementation to add the new behaviour.

- Expose a `HasElementType` property and `GetElementType` method on ITypeInfo. This would require adding public API, and would be a source breaking change, as this is an interface.
- Directly check for ReflectionTypeInfo. Unfortunately this breaks the abstraction that ITypeInfo provides, but provides direct access to a reflected type without exposing public API. Also, it means that this functionality is limited to reflection-based test discovery.

/cc @bradwilson 
fixes #946 
fixes #805

### DIFF
--- a/src/xunit.execution/Sdk/TypeUtility.cs
+++ b/src/xunit.execution/Sdk/TypeUtility.cs
@@ -141,7 +141,7 @@ namespace Xunit.Sdk
                 {
                     // Drill down through the generic arguments of the parameter provided,
                     // and find one matching the genericType
-                    // This allows us to resolve complex generics of with any number of parameters
+                    // This allows us to resolve complex generics with any number of parameters
                     // and at any level deep
                     // e.g. Dictionary<T, U>, Dictionary<string, U>, Dictionary<T, int> etc.
                     // e.g. Dictionary<Dictionary<T, U>, List<Dictionary<V, W>>
@@ -251,15 +251,15 @@ namespace Xunit.Sdk
         /// the generic parameters and their values.
         /// </summary>
         /// <param name="genericType">The generic type to be resolved</param>
-        /// <param name="passedParameters">The parameter values being passed to the test method</param>
-        /// <param name="methodParameters">The parameter infos for the test method</param>
+        /// <param name="parameters">The parameter values being passed to the test method</param>
+        /// <param name="parameterInfos">The parameter infos for the test method</param>
         /// <returns>The best matching generic type</returns>
-        public static ITypeInfo ResolveGenericType(this ITypeInfo genericType, object[] passedParameters, IParameterInfo[] methodParameters)
+        public static ITypeInfo ResolveGenericType(this ITypeInfo genericType, object[] parameters, IParameterInfo[] parameterInfos)
         {
-            for (var idx = 0; idx < methodParameters.Length; ++idx)
+            for (var idx = 0; idx < parameterInfos.Length; ++idx)
             {
-                ITypeInfo methodParameterType = methodParameters[idx].ParameterType;
-                Type passedParameterType = passedParameters[idx]?.GetType();
+                ITypeInfo methodParameterType = parameterInfos[idx].ParameterType;
+                Type passedParameterType = parameters[idx]?.GetType();
 
                 Type matchedType = null;
                 if (ResolveGenericParameter(genericType, methodParameterType, passedParameterType, ref matchedType))
--- a/src/xunit.execution/Sdk/TypeUtility.cs
+++ b/src/xunit.execution/Sdk/TypeUtility.cs
@@ -2,6 +2,7 @@
 using System.Collections.Generic;
 using System.Globalization;
 using System.Linq;
+using System.Reflection;
 using Xunit.Abstractions;
 
 namespace Xunit.Sdk
@@ -90,6 +91,14 @@ namespace Xunit.Sdk
             return $"<{string.Join(", ", typeNames)}>";
         }
 
+        /// <summary>
+        /// Resolves an individual generic type given an intended generic parameter type and the type of an object passed to that type.
+        /// </summary>
+        /// <param name="genericType">The generic type, e.g. T, to resolve.</param>
+        /// <param name="methodParameterType">The non-generic or open generic type, e.g. T, to try to match with the type of the objecct passed to that type.</param>
+        /// <param name="passedParameterType">The non-generic or closed generic type, e.g. string, used to resolve the method parameter.</param>
+        /// <param name="resultType">The resolved type, e.g. the parameters (T, T, string, typeof(object)) -> (T, T, string, typeof(string)).</param>
+        /// <returns>True if resolving was successful, else false.</returns>
         private static bool ResolveGenericParameter(this ITypeInfo genericType, ITypeInfo methodParameterType, Type passedParameterType, ref Type resultType)
         {
             // We can never infer the type parameter from a null type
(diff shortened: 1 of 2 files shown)


hughbe:
Just opened and closed so the CLA bot shows up

hughbe:
I've updated this PR to add a bunch more tests to verify the new feature causes no breaking changes.

N.B: 
fixes #946 
fixes #805

bradwilson:
> Expose a HasElementType property and GetElementType method on ITypeInfo.

Unfortunately, everything in `xunit.abstractions.dll` must be considered set it stone, because it is the mechanism that enables version independence with runners, with both backward- and forward-compatibility. This would basically be the cause of a major version revision (i.e., 3.0).

I think adding this for reflection-only is a reasonable strategy for today.

hughbe:
> Unfortunately, everything in xunit.abstractions.dll must be considered set it stone
I didn't go for that approach for that reason. I don't think having reflection sets an awful precedent, as I recall seeing other examples in the codebase

bradwilson:
Merged, thanks!

PR #966, opened 2016-09-06
Improve perf and allocations of GetParameters

hughbe:
- Don't use linq: 7x faster according to my benchmarks
- Cache the result

This saves two array allocations and an Array.Copy call in `MethodInfo.GetParameters()`

/cc @bradwilson


bradwilson:
Thanks!

PR #983, opened 2016-09-21
Cache objects which might have to be retrieved across remoting boundaries

TimLovellSmith:
in DelegatingXmlCreationSink.CreateTestResultElement().

Rationale:
I did profiling of latest xunit bits from the myget feed, with the msbuild runner, and found that 4.3% of CPU samples were used in CreateTestResultElement, and most of that is used up on TransparentProxyStub_CrossContext.

I think this change should cut down on the number of cross-context calls.


bradwilson:
Thanks!

TimLovellSmith:
Just to check, I tested it out with latest CI build and it got the expected small saving in CPU usage, CreateTestResultElement() went down to just 2.9% of CPU samples inclusive.

PR #997, opened 2016-10-13
Add support for optional parameters in a test method

hughbe:
I seem to remember seeing an issue in the repo requesting something like this but can't find it now!

Also: http://stackoverflow.com/questions/14818439/xunit-test-engines-inlinedataattribute-optional-method-parameters

/cc @bradwilson

### DIFF
--- a/test/test.xunit.execution/Acceptance/Xunit2TheoryAcceptanceTests.cs
+++ b/test/test.xunit.execution/Acceptance/Xunit2TheoryAcceptanceTests.cs
@@ -4,7 +4,6 @@ using System.Collections.Generic;
 using System.Globalization;
 using System.Linq;
 using System.Reflection;
-using System.Threading;
 using Xunit;
 using Xunit.Abstractions;
 using Xunit.Sdk;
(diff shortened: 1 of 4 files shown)


hughbe:
Hmm, how odd that the tests are failing: using TestDriven.NET to run tests (as per the recommended contributor guidelines), I get all passes
tests

@bradwilson is there some sort of weird CI stuff going on? Could you take a look when you have time :)

hughbe:
Okay, so it turns out xunit has two different runners of theory tests! I had only managed to test one of them from TestDriven.net, but the CI tests a different one.

I've fixed this and removed some code that isn't necessary at all

hughbe:
@bradwilson can you take a look if you have time - I keep coming up with places I wanna use this in my tests, thanks!

test/test.xunit.execution/Acceptance/Xunit2TheoryAcceptanceTests.cs
```diff
@@ -14,6 +13,102 @@ public class Xunit2TheoryAcceptanceTests
...
+            [Theory]
+            [InlineData("abc", 6)]
+            public void OneOptional_TwoParameters_TwoPassed(string s, int i = 5)
+            {
+                Assert.Equal("abc", s);
+                Assert.Equal(6, i);
+            }
+
+            [Theory]
+            [InlineData()]
```
bradwilson: Never thought I'd see this. :-p

bradwilson:
Merged, thanks!

PR #1002, opened 2016-10-24
Add support for params arrays in test methods

### LINKED ISSUE #763: Why can't I use params for Theory method parameters
Hi,

I'm a new guy working with xUnit and I was wondering why can't I do something like this:

```
[Theory]
[InlineData("test", "", "thx bai")]
[InlineData("", "test", "thx bai")]
[InlineData("abc", "def", "ghi")]
[InlineData("abc", "ABC", "abc", "ABC")]
public void Testing_String_Array(params string[] strings)
{
        var commonPrefix = MyClass.DoMyChores(strings);

        commonPrefix.Should().Equals(string.Empty);
}

```
I aim at test readability. I tried to the same with MemberData, but that lead me to...

```
public static IEnumerable<object[]> ValidStrings
{
        get
        {
                return new[]
                {
                        new object[] { new string[] { "test", string.Empty, "thx bai" } },
                        new object[] { new string[] { string.Empty, "test", "thx bai" } },
                        new object[] { new string[] { "abc", "def", "ghi" } },
                        new object[] { new string[] { "abc", "ABC", "abc", "ABC" } }
                };
        }
}

```

Which works, but looks overly complicated.

Is there a chance that this will be fixed?

### LINKED ISSUE #945: params object[] as test argument
What about adding `params object[]` to supported test arguments?
Now it is impossible.
Eg When I try to use it with `[InlineData(-1)]`, I'm getting this:

> System.ArgumentException : Object of type 'System.Int32' cannot be converted to type 'System.Object[]'.

hughbe:
Fixes #945
Fixes #763

I also changed the TypeUtility method (I presume you made it public, right?) to be an extension method, to be consistent with the other methods in the class.

Allows the following tests (see the tests added for more details):

```
[Theory]
[InlineData()]
[InlineData(null)]
[InlineData(1)]
[InlineData(new object[] { new object[] { 1 } })]
[InlineData(new int[] { 1 })]
[InlineData(1, 2, 3, 4, 5, 6)]
public void ParamsObjectTest(params object[] array) {}
```

/cc @bradwilson
@klimcio and @Serg046 FYI

### DIFF
--- a/src/xunit.execution/Sdk/Frameworks/Runners/XunitTestCaseRunner.cs
+++ b/src/xunit.execution/Sdk/Frameworks/Runners/XunitTestCaseRunner.cs
@@ -47,8 +47,7 @@ namespace Xunit.Sdk
             var parameterTypes = new Type[parameters.Length];
             for (int i = 0; i < parameters.Length; i++)
                 parameterTypes[i] = parameters[i].ParameterType;
-
-            testMethodArguments = TypeUtility.ResolveMethodArguments(TestMethod, testMethodArguments);
+            
             TestMethodArguments = Reflector.ConvertArguments(testMethodArguments, parameterTypes);
 
             IEnumerable<Attribute> beforeAfterTestCollectionAttributes;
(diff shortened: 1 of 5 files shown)


bradwilson:
Thanks!

jamesqo:
Nice! I just ran into this issue, glad to see it's fixed upstream.

PR #1005, opened 2016-10-24
Fix NullReferenceException if IDataDiscoverer.GetData returns null

### LINKED ISSUE #928: MemberData source initialization in constructor
Hello,

Is it ok that the following code doesn't work?

``` csharp
namespace XUnitIssueReporducing
{
    using System.Collections.Generic;
    using Xunit;

    public class Tests
    {
        public static IEnumerable<object[]> InitializedInConstructor;

        public Tests()
        {
             InitializedInConstructor = new List<object[]>
            {
                new object[] { "1", "2"}
            };
        }

        [Theory, MemberData(nameof(InitializedInConstructor))]
        public void DoesntWork(string str1, string str2)
        {
            Assert.False(str1 == str2);
        }
    }
}
```

I mean, test failed with exception:

```
System.NullReferenceException
Object reference not set to an instance of an object.
   at Xunit.Sdk.XunitTheoryTestCaseRunner.<AfterTestCaseStartingAsync>d__7.MoveNext() in C:\BuildAgent\work\cb37e9acf085d108\src\xunit.execution\Sdk\Frameworks\Runners\XunitTheoryTestCaseRunner.cs:line 62
```

Thanks.

hughbe:
Fixes #928

/cc @bradwilson
@yahor-stalnik FYI

### DIFF
--- a/src/xunit.execution/Sdk/Frameworks/Runners/XunitTheoryTestCaseRunner.cs
+++ b/src/xunit.execution/Sdk/Frameworks/Runners/XunitTheoryTestCaseRunner.cs
@@ -61,7 +61,13 @@ namespace Xunit.Sdk
                     var discovererType = SerializationHelper.GetType(args[1], args[0]);
                     var discoverer = ExtensibilityPointFactory.GetDataDiscoverer(diagnosticMessageSink, discovererType);
 
-                    foreach (var dataRow in discoverer.GetData(dataAttribute, TestCase.TestMethod.Method))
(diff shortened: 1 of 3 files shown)


hughbe:
@bradwilson in a lot of my PRs I'm (hopefully) improving how theory works. Weirdly, I have to duplicate much of my code in two places - is this really the best way to go? I think its by design, due to pre-enumeration/serialization etc. etc.

bradwilson:
@hughbe I often find the best way to remove duplication is to just let it happen, and when things have settled down, look for the best pattern to extract the common logic. So, once you're done with your flurry of PRs, let's reconsider how to make the code cleaner. :-D

Thanks!

PR #1007, opened 2016-10-24
Fix TestOutputHelper crashing on invalid xml chars

### LINKED ISSUE #876: Failed to save xunit test result xml due to invalid xml char
The following test fail under dotnet core with xunit build 2.2.0-beta1-build3239.

``` csharp
    public class Tests
    {
        private readonly ITestOutputHelper _output;

        public Tests(ITestOutputHelper output)
        {
            _output = output;
        }

        [Fact]
        public void Test()
        {
            _output.WriteLine(Encoding.UTF8.GetString(new byte[] { 0x1B }));
        }
    }
```

run: dotnet test -xml results.xml

The process will fail with error code 3 and the following message:
error: '', hexadecimal value 0x1B, is an invalid character.

If you run dotnet test without the -xml option, the test is successful.

hughbe:
Fixes #876

/cc @bradwilson

@gigi81 and @edumunoz FYI

### DIFF
--- a/src/xunit.execution/Sdk/Frameworks/TestOutputHelper.cs
+++ b/src/xunit.execution/Sdk/Frameworks/TestOutputHelper.cs
@@ -57,10 +57,37 @@ namespace Xunit.Sdk
             {
                 GuardInitialized();
 
-                buffer.Append(output);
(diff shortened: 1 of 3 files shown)


bradwilson:
Thanks!

PR #1008, opened 2016-10-24
Fix serialization and deserialization of negative zero doubles/floats

### LINKED ISSUE #787: Negative zero is not serialized correctly in Theory data
```
[Theory]
[InlineData(0.0, false)]
[InlineData(-0.0, true)]
public void Test(double value, bool isNegativeZero)
{
    double expected = isNegativeZero ? -0.0 : 0.0; 
    long expectedBits = BitConverter.DoubleToInt64Bits(expected);

    long actualBits = BitConverter.DoubleToInt64Bits(value);

    Assert.Equal(expectedBits, actualBits);
}
```

Test passes in `xunit.console.exe`, but fails when run from Visual Studio.

hughbe:
Fixes #787

/cc @bradwilson

@EmperorXLII FYI


bradwilson:
Thanks!

PR #1017, opened 2016-11-05
Updated AllException test to expect the updated format.

Jameak:
This update goes with my change in xunit/assert.xunit#7


bradwilson:
Merged, thanks!

PR #1020, opened 2016-11-06
Add support for implicit and explicit conversions in theories

### LINKED ISSUE #1013: Please use the implicit or explicit type conversion operators on theory arguments
I have types that have a `[DebuggerDisplay]` attribute so debugging is quite a lot easier. However, they cannot be arguments to a theory with `[InlineData]` because they cannot be provided as constants. Some of them convert nicely (and trivially) to a blittable type however, which I can use with `[InlineData]`.

However if I change the theories method signature, all tests fail with 
`Object of type 'A' cannot be converted to type 'B'.`
at
`RuntimeType.TryChangeType(object, Binder, CultureInfo, Boolean)`
(...)
`TestInvoker'1.CallTestMethod(object)`.

If I do this at runtime, there's no problem - first of all it obviously compiles without explicit type cast.

Truth be told, one reason why I would especially like this is to make the names of tests more relevant in Test Explorer in VS.

hughbe:
Fixes #1013

### DIFF
--- a/test/test.xunit.execution/Acceptance/Xunit2TheoryAcceptanceTests.cs
+++ b/test/test.xunit.execution/Acceptance/Xunit2TheoryAcceptanceTests.cs
@@ -250,7 +250,9 @@ public class Xunit2TheoryAcceptanceTests
 
             Assert.Collection(results.Cast<ITestPassed>().OrderBy(r => r.Test.DisplayName),
                 result => Assert.Equal(@"Xunit2TheoryAcceptanceTests+TheoryTests+ClassWithOperatorConversions.ExplicitConversion(e: Explicit { Value = ""abc"" })", result.Test.DisplayName),
-                result => Assert.Equal(@"Xunit2TheoryAcceptanceTests+TheoryTests+ClassWithOperatorConversions.ImplicitConversion(i: Implicit { Value = ""abc"" })", result.Test.DisplayName)
+                result => Assert.Equal(@"Xunit2TheoryAcceptanceTests+TheoryTests+ClassWithOperatorConversions.ImplicitConversion(i: Implicit { Value = ""abc"" })", result.Test.DisplayName),
+                result => Assert.Equal(@"Xunit2TheoryAcceptanceTests+TheoryTests+ClassWithOperatorConversions.IntToLong(i: 1)", result.Test.DisplayName),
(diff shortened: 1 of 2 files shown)


Ausm:
This still doesn't work for int and long:
``` C#
[Theory]
[InlineData(5)]
public void TestViaIntAndLong(long x) { }
```
because 
```typeof(long).IsAssignableFrom(typeof(int))``` is false and there is no op_Implicit or op_Explicit function.
bool, byte, short, int and long are all the same in the IL, and the conversion works via the runtime.
There is no implicit or explicit operator for those types.
See #1025, this solution should work for all types.

**Comment by hughbe** [comment]
It does still work - we return the original value here: https://github.com/xunit/xunit/pull/1020/files#diff-60e08670daa222f4aa773b0c4f4030c9R145

**Comment by hughbe** [comment]
I've updated my PR with tests demonstrating primitive conversions

**Comment by Ausm** [comment]
Aha, yes, I see, you are null checking the MethodInfo's, and RuntimeType.TryChangeType does the implicit conversion for primitives later when calling the theory method.
But what's with explicit conversions between primitives, they fail, because there is no op_Explicit.
I don't think that it's a good idea to query and invoke the op_... methods via reflection. Converting an object from one type to an other can be complex, and you cannot cover all special cases and you don't know how the .Net framework is changing in the future. And MethodInfo.Invoke is a strong performance killer, and you call it for each parameter of each dataRow given to each theory.

Whether it should be in the ResolveMethodArguments or in the TestInvoker, don't know.
I put it in the TestInvoker, because the Exception was thrown there, and I tried to fix it.

File: test/test.xunit.execution/Acceptance/Xunit2TheoryAcceptanceTests.cs [hunk 1]
```diff
@@ -244,6 +244,70 @@ public void OptionalParameters_ManyPassed(string s = "abc", int i = 1, params ob
...
+            public void UIntToULong(ulong i)
+            {
+                Assert.Equal(1UL, i);
+            }
+
+            public class Explicit
+            {
+                public string Value { get; set; }
+
+                public static implicit operator Explicit(string value)
```
jamesqo: Was this meant to be `explicit`?

bradwilson:
Updated and merged by hand. Thanks!

PR #1023, opened 2016-11-07
Add tests for included collection in Assert.Collection failure

hughbe:
See https://github.com/xunit/assert.xunit/pull/9


bradwilson:
Thanks!

PR #1024, opened 2016-11-07
Don't allow Fact tests to be generic

### LINKED ISSUE #969: Minor Diagnostic: xunit fails to execute generic test method
As far as I can see, this is not really a bug in xunit, but a clearer error message might help other people who stumble over this issue.

If I compile this F# test
image

and try to run it, then I get this error
image
The error says something like 

> Late-bound operations can't be executed for fields with types or methods, for which ContainsGenericParameters was set to "true"

The real issue is that the code was compiled to this C# equivalent
image

The real fix is to change the F# code to this 
image
so that it does not get compiled to a generic method.

In general, this issue is not restricted to F#, someone might also define a generic test method in C#, but there it would be more obvious what is wrong

Version: I have tested it with both 2.1.0 (latest stable) and 2.2.0-beta2-build1149 (latest prerelease)

hughbe:
Fixes #969

### DIFF
--- a/src/xunit.execution/Sdk/Frameworks/FactDiscoverer.cs
+++ b/src/xunit.execution/Sdk/Frameworks/FactDiscoverer.cs
@@ -42,10 +42,20 @@ namespace Xunit.Sdk
         /// <returns>Returns zero or more test cases represented by the test method.</returns>
         public virtual IEnumerable<IXunitTestCase> Discover(ITestFrameworkDiscoveryOptions discoveryOptions, ITestMethod testMethod, IAttributeInfo factAttribute)
         {
-            var testCase =
-                testMethod.Method.GetParameters().Any()
-                    ? new ExecutionErrorTestCase(diagnosticMessageSink, discoveryOptions.MethodDisplayOrDefault(), testMethod, "[Fact] methods are not allowed to have parameters. Did you mean to use [Theory]?")
-                    : CreateTestCase(discoveryOptions, testMethod, factAttribute);
+            IXunitTestCase testCase;
+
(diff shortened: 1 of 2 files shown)


src/xunit.execution/Sdk/Frameworks/FactDiscoverer.cs
```diff
@@ -42,10 +42,20 @@ protected virtual IXunitTestCase CreateTestCase(ITestFrameworkDiscoveryOptions d
...
+            IXunitTestCase testCase;
+
+            if (testMethod.Method.GetParameters().Any())
+            {
+                testCase = new ExecutionErrorTestCase(diagnosticMessageSink, discoveryOptions.MethodDisplayOrDefault(), testMethod, "[Fact] methods are not allowed to have parameters. Did you mean to use [Theory]?");
+            }
+            else if (testMethod.Method.IsGenericMethodDefinition)
+            {
+                testCase = new ExecutionErrorTestCase(diagnosticMessageSink, discoveryOptions.MethodDisplayOrDefault(), testMethod, "[Fact] methods are not allowed to be generic.");
+            }
```
jamesqo: Nit: Maybe the XML doc for this method should be updated as well.

```xml
        /// <summary>
        /// Discover test cases from a test method. By default, inspects the test method's argument list
        /// to ensure it's empty, and if not, returns a single <see cref="ExecutionErrorTestCase"/>;
        /// otherwise, it returns the result of calling <see cref="CreateTestCase"/>.
        /// </summary>
```

bradwilson:
Updated and merged by hand. Thanks!

PR #771, opened 2016-02-11
Fixing: Wrong implementation of the ImplicitlyConvertibleDataPasses Test (Issue #770)

Ausm:

PR #1025, opened 2016-11-07
Fixing: Wrong implementation of the ImplicitlyConvertibleDataPasses Test

Ausm:
Reopened for CLA bot. Old PR: #771

### DIFF
--- a/src/xunit.execution/Sdk/Frameworks/Runners/TestInvoker.cs
+++ b/src/xunit.execution/Sdk/Frameworks/Runners/TestInvoker.cs
@@ -1,5 +1,6 @@
 ﻿using System;
 using System.Linq;
+using System.Linq.Expressions;
 using System.Reflection;
 using System.Security;
 using System.Threading;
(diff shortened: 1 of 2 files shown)


Ausm:
@bradwilson:
I'm not satisfied with that solution actually.
It solves the problem, but I think it would be better, to dynamically create a function via Lambda Expressions, which calls the theory method.
And we can do all the casting, and array accessing etc there, it will be handled like the Framework handles it and it would give us a much better Performance.
What do you think?

bradwilson:
Sorry that I'm just getting to look at this now.

I'm not 100% clear what problem it is you're trying to solve. I'm a little leary about changing the signature of extensibility classes without understanding the benefit that we're trying to achieve.

Thanks...

Ausm:
Nothing you need to be sorry for...
It was an alternative fix for #1020, or #1020 was the alternative for this.
Main difference is that I use a lambda expression for conversion here, while the solution in the other PR uses `RuntimeType.TryChangeType` for primitives and a reflection call for types with implicit/explicit cast operators.
A performance killer on one side, since reflection calls are fast as a snail.
And on the other side it uses reflection for types with according op_... methods, and for primitives, which do not have op_... methods even if they explicitly or implicitly convertible, it uses `RuntimeType.TryChangeType`, any third case would fail.

But you already merged the other PR, so I think you can close this.

PR #1026, opened 2016-11-07
Fix obfuscated ArgumentNullException thrown when a DataDiscoverer does not exist

hughbe:
Fixes xunit#839

- DataDiscoverer doesn't exist
- DataDiscoverer is not an IDataDiscoverer

@citizenmatt FYI


hughbe:
Pushed a fix for the CI error (had to update 2 different locations).

I also added an extra check making sure the DataDiscoverer specified is actually an IDataDiscoverer - previously an InvalidCastException was thrown

bradwilson:
Thanks!

PR #1055, opened 2016-12-20
Enable a NETStandard 1.5 version that includes support for DiaSession

### LINKED ISSUE #1054: Go to source in VS not working for xunit .net core tests
## Steps to reproduce
* Install VS 2017 RC with .net and docker preview workload
* Create a new xunit test project
* Build the project
* After tests are discovered in the Test Explorer, double click on a Test

## Expected
User should be able to navigate to source code

## Actual
Nothing happens since source code information is not available in Test Explorer.

onovotny:
Fixes #1054 by providing a version of the runner utility that supports DiaSession:
img

After this is merged, the visual studio runner needs to use this updated version of the utility. In the interim, manually specifying the package works.

### DIFF
--- a/src/xunit.runner.utility/Configuration/ConfigReader_Json.cs
+++ b/src/xunit.runner.utility/Configuration/ConfigReader_Json.cs
@@ -1,7 +1,7 @@
 using System;
 using System.IO;
 
-#if PLATFORM_DOTNET && NETSTANDARD1_1
+#if NETSTANDARD1_1
 using System.Reflection;
 #endif
 
@@ -129,7 +129,7 @@ namespace Xunit
             return null;
         }
 
-#if PLATFORM_DOTNET && NETSTANDARD1_1
+#if NETSTANDARD1_1
         static Lazy<MethodInfo> fileOpenReadMethod = new Lazy<MethodInfo>(GetFileOpenReadMethod);
 
         static MethodInfo GetFileOpenReadMethod()
--- a/src/xunit.runner.utility/Configuration/ConfigReader_Json.cs
+++ b/src/xunit.runner.utility/Configuration/ConfigReader_Json.cs
@@ -1,7 +1,7 @@
 using System;
 using System.IO;
 
-#if PLATFORM_DOTNET
+#if PLATFORM_DOTNET && NETSTANDARD1_1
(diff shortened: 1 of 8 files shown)


onovotny:
While this works for `public void` tests, I'm still getting `External` on `async Task` and `async void`.

onovotny:
I believe that the async issues are a bug in the VS Test Platform `DiaSession`.

[UnitTest1.cs.txt](https://github.com/xunit/xunit/files/664115/UnitTest1.cs.txt)

I have a repro of the issue I’m seeing with async methods. I extracted the relevant code so that I could run it in the context of a netcoreapp1.0 unit test.

To execute, create a new xUnit Unit Test .NET Core project and add the attached test.

Look at the following two tests:

1.	TestSyncMethod. This one works correctly. You can see that DiaSessionWrapperHelper determines the assembly path and leaves the method and type alone. It then passes those into DiaSession and gets data back.
2.	TestAsyncMethod. This is the issue. Our DiaSessionWrapperHelper does correctly update the type and method to point to the AsyncStateMachine StateMachine type. The problem is that when we pass those into DiaSession, we get null back.

onovotny:
This appears to also affect MSTest. Create a new .NET Core MSTest project with two test methods

```csharp
        [TestMethod]
        public void TestMethod1()
        {
        }

        [TestMethod]
        public async Task TestMethod2()
        {
            await Task.CompletedTask;
        }
```

The second one will show up as External. This must be a bug in the PortableSymbolReader within the object model library.

I think this PR is ready to go and async tests will work correctly once the object model types are fixed.

piotrpMSFT:
@dotnet/vstest

codito:
Thanks @onovotny, will look into the issue in DiaSession.

src/xunit.runner.utility/Configuration/ConfigReader_Json.cs
```diff
@@ -1,7 +1,7 @@
 using System;
 using System.IO;
 
-#if PLATFORM_DOTNET
+#if PLATFORM_DOTNET && NETSTANDARD1_1
```
bradwilson: Things like this should probably just be `#if NETSTANDARD1_1`

PR #1077, opened 2017-01-18
Fixes #1076 - public setter for IgnoreFailures

### LINKED ISSUE #1076: xunit.msbuild runner - not possible to set IgnoreFailures in msbuild file
Since IgnoreFailures is protected set; it is not possible to set the value from an msbuild file. 

To set it in an msbuild file it needs have a public setter.

lmeyer3:
Proactively created this PR since its a very simple change.

 IgnoreFailures needs to have a public setter. Fixes #1076


bradwilson:
Thanks!

PR #1097, opened 2017-02-12
Make TypeUtility.ResolveMethodArguments use MethodBase instead of MethodInfo

hughbe:
This is necessary to avoid a breaking change in the future when we release.

This is becasue I'm working on resolving arguments for ConstructorInfos


bradwilson:
Thanks!

PR #1098, opened 2017-02-12
Add support for optional, default and params parameters in test fixtures

### LINKED ISSUE #985: Even with optional params error: "The following constructor parameters did not have matching fixture data: ..."
I'd like to have a few optional params in my constructor that I would maybe override through a collection fixture. 
However this seems unsupported

hughbe:
Fixes #985

TypeUtility.GetDefaultValue(this TypeInfo typeInfo) will be used elsewhere in a subsequent PR


bradwilson:
Thanks!

PR #1099, opened 2017-02-12
Add support for optional parameters without default values in test methods

hughbe:
Relies on #1098


bradwilson:
Manually merged and pushed, thanks!

PR #1110, opened 2017-02-14
Removed extra appdomain getting created for SourceInformation

Shyam-Gupta:


onovotny:
Tagging https://github.com/Microsoft/vstest/issues/349#issuecomment-279724359 and https://github.com/Microsoft/vstest/issues/485

onovotny:
@Shyam-Gupta what do you mean by "extra" app-domain? In the desktop framework, all test runs (and discovery) are done in another appdomain to prevent the file from being locked. Otherwise rebuild will fail.

Are you saying that from that other appdomain, this is creating another one?

Shyam-Gupta:
@onovotny Yes that's right.

bradwilson:
Sorry, this fix will leave the DLL locked (we've experienced that in the past, which is why this code uses an app domain).

Shyam-Gupta:
@bradwilson The test DLL gets loaded in another [appdomain](https://github.com/xunit/xunit/blob/master/src/xunit.runner.utility/Frameworks/v2/Xunit2Discoverer.cs)which is not getting removed as part of this change and hence will not leave the DLL locked.

onovotny:
@Shyam-Gupta Have you tried VS 2015 as well? The same binaries are used there too.

bradwilson:
And VS2012 and VS2013...

KevinH-MS:
@Shyam-Gupta can this logic be triggered by some "no appdomain" argument that's passed in by vstest.console (only in the dotnet cli command line case).

Shyam-Gupta:
@bradwilson @onovotny 
I tried running this code change in VS 2015 and found that it causes test DLL to get locked.

However I think if we move the functionality of [DiaSessionWrapperHelper] class(https://github.com/xunit/xunit/blob/master/src/xunit.runner.utility/Utility/DiaSessionWrapperHelper.cs) to test.execution.desktop.dll then it will solve the problem of DLL lock as test.execution.desktop.dll gets loaded in test dll appdomain and hence the DLL lock will go away as soon as the appdomain gets unloaded. This approach also gives an additional benefit that the test dll will need to be loaded only once, rather than twice which is the current scenario.

Do you think this approach is feasible ?

bradwilson:
No, we can't.

The things which exist in runner utility are for runners. This DLL provides both forward and backward compatibility and a de-coupling of the version of runner utility vs. the version of xUnit.net that the tests were written against. Moving this would break this compatibility.

Besides, even if the class were moved, that would be irrelevant, because we would still need the app domain to prevent locking the test assembly.

Shyam-Gupta:
@bradwilson 

I implemented the change I suggested to see whether it locks test DLL. 
Tested it multiple times on VS2015. Found that it doesn't lock test DLL anymore and rebuilds work fine. This is because now for async methods we can do GetTypes() on the already loaded test DLL in test DLL appdomain rather than doing ReflectionOnlyLoad again on the test DLL in xunit.runner.utility appdomain which was the case earlier.

Please take a look at the updated PR with these changes.

src/abstractions/xunit.abstractions/Frameworks/ISourceInformationProvider.cs
```diff
@@ -15,5 +15,14 @@ public interface ISourceInformationProvider : IDisposable
...
         ISourceInformation GetSourceInformation(ITestCase testCase);
+
+        /// <summary>
+        /// Returns the source information for a test method inside a test class.
+        /// </summary>
+        /// <param name="typeName">Test class containing the method.</param>
+        /// <param name="methodName">The test case to retrieve information for.</param>
+        /// <returns>The source information, with null string and int values when the information is not available.
+        /// Note: return value should never be <c>null</c>, only the interior data values inside.</returns>
+        ISourceInformation GetSourceInformation(string typeName, string methodName);
```
bradwilson: Abstractions cannot be changed. Abstractions is our forward and backward compatibility layer for runners.

This is an iron-clad rule. There are no exceptions.

Shyam-Gupta: Removed xunit.abstractions changes in the latest commit. Please take a look.

bradwilson:
@Shyam-Gupta Upon testing with our integration project, I've found that this change breaks async test method source support in UWP, and all test methods in .NET Core. I will be reversing this change as a result.

bradwilson:
FYI, I had rebased and cleaned up your PR (it had a broken test, left over an unused file, etc.). I did push those changes into your fork. Because we have to lock down 2.2, we're going to have to move this fix into a future release, if it can be made to work correctly.

Thanks,
Brad

bradwilson:
Test Explorer with UWP:



(Tests for .NET Core are all shown as external)

Shyam-Gupta:
Apologies for the inconvenience caused. I will get back with the updated PR.

PR #1124, opened 2017-02-19
Fix incorrect JSON encoding

### LINKED ISSUE #1116: Tests not automatically reported in TeamCity after update to RC4 tooling
Using .NET SDK ```1.0.0-rc4-004788``` with xunit ```2.2.0-rc3-build3528``` xunit tests are no longer reported automatically as part of a TeamCity build when run in a PowerShell script with ```dotnet test <project>```.

Does this need to be manually configured now with the move to MSBuild, or should this still "just work"?

martincostello:
Fix incorrect escaping of quotes for encoding strings as JSON due to one too many slashes being added.

Resolves #1116.


bradwilson:
Thanks!

PR #1126, opened 2017-02-20
Add tests for Assert.Equal with IEquatable<typeof(actual>) and IComparable<typeof(actual)>

hughbe:
Tests for xunit/assert.xunit#17


bradwilson:
Thanks!

PR #1129, opened 2017-02-22
Fix Processor Architecture in Path of Build.ps1

commonsensesoftware:
I just pulled the latest and a clean build via **Build.ps1** fails because the path in the **RegexReplace** task is incorrect. Since the build configuration uses **x86**, the path should reflect that rather than **x64**.

Instead of submitting a new issue, I figured it would be easier and faster to just submit a PR. The CLA is already signed.

### DIFF
--- a/xunit.msbuild
+++ b/xunit.msbuild
@@ -78,7 +78,7 @@
     <RegexReplace
         Pattern="type=&quot;Xunit.ConsoleClient.XunitConsoleConfigurationSection, xunit.console&quot;"
         Replacement="type=&quot;Xunit.ConsoleClient.XunitConsoleConfigurationSection, xunit.console.x86&quot;"
-        Files="src\xunit.console\bin\$(Configuration)_x86\net452\win7-x64\xunit.console.x86.exe.config" />
+        Files="src\xunit.console\bin\$(Configuration)_x86\net452\win7-x86\xunit.console.x86.exe.config" />
   </Target>
 
   <Target Name="_Test32" DependsOnTargets="Build">


commonsensesoftware:
ok ... I'm confused. I don't understand why TeamCity outputs

_~\src\xunit.console\bin\Release_x86\net452\win7-**x64**\xunit.console.x86.exe.config_

in the build, but when I build locally the output is 

_~\src\xunit.console\bin\Release_x86\net452\win7-**x86**\xunit.console.x86.exe.config_

Shouldn't they be the same? Shouldn't the path include **x86** for a 32-bit build? What am I missing?

bradwilson:
I'm honestly not sure what that path means, nor why it's different on your box than on mine (or the CI machine). I do know that it's not present in the MSBuild v15 tooling any more, as you can see from my vs2017 branch: https://github.com/xunit/xunit/blob/4863f9e5a7f7bf14f9cf42978d699d834bc109a2/xunit.msbuild#L66

```xml
    <RegexReplace
        Pattern="type=&quot;Xunit.ConsoleClient.XunitConsoleConfigurationSection, xunit.console&quot;"
        Replacement="type=&quot;Xunit.ConsoleClient.XunitConsoleConfigurationSection, xunit.console.x86&quot;"
        Files="src\xunit.console\bin\$(Configuration)_x86\net452\xunit.console.x86.exe.config" />
```

commonsensesoftware:
Uggg.... You've gotta love the _bleeding edge_ at times. I think the difference is which version of the .NET Core SDK is being used. I'm fully up-to-date (RC4). I'm not sure which version the build server is using. I was able to find the known bug - [Build does not honor the runtime setting](https://github.com/dotnet/cli/issues/3387), which seems to outline the problem I'm seeing.

I'll close out this PR. There's no need to go back and forth on trying to match tooling paths; especially for a known issue. Since you're already working on VS2017, I fully expect things to resolve themselves on that path. I would suggest, however, adding a note somewhere in one of the documentation pages as to which version of the .NET Core SDK is expected and a perhaps a brief mention of this quirk.

PR #1142, opened 2017-02-27
Removed extra appdomain getting created for SourceInformation 

Shyam-Gupta:
Refer PR: https://github.com/xunit/xunit/pull/1110


bradwilson:
Thanks!

bradwilson:
Unfortunately, I'm going to have to revert this again.

This solution breaks source information for `async` tests which are using older versions of xUnit.net. A solution will need to be found that not only improves performance with current tests, but also does not break older tests.

I've updated the integration project with this broken build so you can see it in action: https://github.com/xunit/xunit.integration/tree/broken-pr1142

bradwilson:
Ideally, no changes should need to be made in the execution library. This is why this change is broken w/ prior versions of xUnit.net, because the runner *only* depends on `xunit.runner.utility`; the version of `xunit.execution.*.dll` that's being used is completely dependent on the test author, not the runner author.

PR #1180, opened 2017-03-24
Initial support for netstandard1.5 in xunit.runner.msbuild

mariusGundersen:
The msbuild runner of xUnit should support netstandard1.5.

Consider this a starting point and a RFC. It compiles, but that's about it.

### DIFF
--- a/src/common/AssemblyExtensions.cs
+++ b/src/common/AssemblyExtensions.cs
@@ -1,4 +1,4 @@
-#if NET35 || NET40 || NET452
+#if NET35 || NET40 || NET452 || NETSTANDARD1_5
 
 using System;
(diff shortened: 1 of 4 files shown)


bradwilson:
Have you tested this with `dotnet msbuild` to see if it's able to run unit tests?

mariusGundersen:
I think it did, but I gave up on trying to make it work in the short term, and decided to use `dotnet test` instead, so I haven't done any more testing on this.

bradwilson:
Thanks. I took your start and finished it up here: https://github.com/xunit/xunit/commit/fd6f472adbaf74eb35db132cc58ff31161e2c4fa

PR #1187, opened 2017-03-27
Add ability to disable TeamCity reporter with env variable

natemcmaster:
I use the TRX reporter on TeamCity and want the ability to suppress the default TC reporter. AFAIK there currently is no way do this with dotnet-test.

Unsetting TEAMCITY_PROJECT_NAME is not a good option since that's required for other build settings. Instead, this adds a check for the environment variable DISABLE_TEAMCITY_XUNIT_REPORTER when determining if TeamCityReporter should be used.

**Btw**, if there is a way to do this already with dotnet-test, please let me know.

### DIFF
--- a/src/xunit.runner.reporters/TeamCityReporter.cs
+++ b/src/xunit.runner.reporters/TeamCityReporter.cs
@@ -8,7 +8,8 @@ namespace Xunit.Runner.Reporters
             => "forces TeamCity mode (normally auto-detected)";
 
         public bool IsEnvironmentallyEnabled
-            => !string.IsNullOrWhiteSpace(EnvironmentHelper.GetEnvironmentVariable("TEAMCITY_PROJECT_NAME"));
+            => !string.IsNullOrWhiteSpace(EnvironmentHelper.GetEnvironmentVariable("TEAMCITY_PROJECT_NAME"))
+               && EnvironmentHelper.GetEnvironmentVariable("DISABLE_TEAMCITY_XUNIT_REPORTER") != "true";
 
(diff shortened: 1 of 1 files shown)


natemcmaster:
cc @bradwilson @onovotny

onovotny:
-norunnerreporter used to exist on the command line, it could probably be made into a variable too?

onovotny:
@natemcmaster Instead of a TeamCity specific setting, I think it'd be better to add back a "NO_AUTO_REPORTERS" variable like the xUnit console has. That could apply to all runner reporters.

natemcmaster:
Sounds good to me. So, close this and open an issue to re-add NO_AUTO_REPORTERS?

bradwilson:
The reading of such a variable should be in the runners, not the reporters (otherwise we force all reporter authors to implement support for the environment variable).

So, these three places:

https://github.com/xunit/visualstudio.xunit/blob/0d9bbb48cf78a376ee354760a992081d85f6b5f8/src/xunit.runner.visualstudio/VsTestRunner.cs#L510
https://github.com/xunit/xunit/blob/86b08ae87a77901c2e7e22abeb06d73db3e69c1b/src/xunit.console/CommandLine.cs#L56
https://github.com/xunit/xunit/blob/86b08ae87a77901c2e7e22abeb06d73db3e69c1b/src/xunit.runner.msbuild/xunit.cs#L366

natemcmaster:
👍 let's do that.

natemcmaster:
https://github.com/xunit/xunit/issues/1191

PR #1209, opened 2017-04-14
Add extensibility points

SaltyDH:
This pull request is a non-functional change to Xunit. It aims to provide fine-grained access to test case creation, test discovery and test runners so that when writing 3rd party libraries, sub-classing can be used to provide additional control over the test execution process.


XdX-Software:
Please merge this pull request!
Makes unit testing much better with real DI!

bradwilson:
Thanks!

Before merging, I made several changes to bring names and parameter lists in line with our other extensibility points. If you're already using this today, that means method names will be slightly different (we use `CreateXyz` rather than `GenerateXyz`, for example) and/or the parameter lists will have changed.

SaltyDH:
Fantastic, thanks for the feedback @bradwilson I'll keep that in mind for the future!

PR #1241, opened 2017-05-04
2 Modifications to GetUniqueID

blfrantz:
1) Adding MethodGenericTypes to the hash, to ensure that distinct UniqueIDs are generated for generic method tests that only differ by a type parameter.
2) Using an assembly name string that does not include version info, to make these UniqueIDs consistent across assembly version bumps.  This makes the UniqueID more useful for correlating results for the same test over time.


bradwilson:
Thanks!

PR #1280, opened 2017-05-12
Add NuGet package for xunit.console

dsplaisted:
This adds a NuGet package for xunit.console.

This would be useful to us in the [dotnet/sdk](https://github.com/dotnet/sdk) and [dotnet/cli](https://github.com/dotnet/cli) repos.  It would allow the test projects to have a main method that calls into xunit.console, so tests could be run via `dotnet run`.  This makes it easier to attach a debugger to tests, since `dotnet test` and `dotnet xunit` both launch separate processes to run the tests.

### DIFF
--- a/src/xunit.console.nuspec
+++ b/src/xunit.console.nuspec
@@ -1,0 +1,24 @@
+﻿<?xml version="1.0" encoding="utf-8"?>
+<package xmlns="http://schemas.microsoft.com/packaging/2011/10/nuspec.xsd">
+  <metadata minClientVersion="2.12">
+    <id>xunit.console</id>
+    <version>99.99.99-dev</version>
+    <title>xUnit.net [Console runner implementation]</title>
+    <authors>James Newkirk,Brad Wilson</authors>
+    <summary>.NET Core console runner, as a linkable library, for runner authors. For running tests, please use xunit.runner.console instead.</summary>
+    <description>.NET Core console runner, as a linkable library, for runner authors. For running tests, please use xunit.runner.console instead.</description>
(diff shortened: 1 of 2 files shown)


bradwilson:
I'm confused. This package already exists:

https://github.com/xunit/xunit/blob/master/src/xunit.runner.console.nuspec

bradwilson:
(Note that we accidentally broke 2.3 beta 2, but it will be fixed in 2.3 beta 3)

dsplaisted:
I didn't notice that package, because the package ID wasn't the same as the DLL I was looking for.

But that package doesn't do exactly what I want.  I want to reference xunit.console.dll and its dependencies from my test project and call into it in my main method:

```c#
public static void Main(string[] args)
{
    var newArgs = args.ToList();
    newArgs.Add(typeof(Program).Assembly.Location);
    new Xunit.ConsoleClient.Program().EntryPoint(newArgs.ToArray());
}
```

bradwilson:
I vaguely understand why you might want this in .NET Core, but never understand why you'd want this in net452.

dsplaisted:
We only need it for .NET Core.  I can imagine wanting it for net452 so that you could run tests the same way in projects that are multi-targeted.

bradwilson:
I'm going to push this for `netcoreapp1.0` and `netcoreapp2.0`, even though I still don't see your exact use case as providing anything that you can't already get by just invoking `dotnet C:\Path\to\xunit.console.dll C:\Path\to\MyTests.dll`.

dsplaisted:
@bradwilson Thanks a bunch!

PR #1287, opened 2017-05-17
Traverse multiple levels in hierarchy to find base method info

### LINKED ISSUE #1068: Test defined in base class fails to run when using IClassFixture
In a 3 tier hierarchy (child -> parent -> grandparent) if a test is defined on the grandparent and overridden in the child the test will not run.  The test needs to be also overridden in the parent.  

TestDriven.Net returns the error "The containing class must have a default constructor." and the XUnit console runner does not find the test.

The issue only presents itself when IClassFixture is being used.

I have tested this using XUnit 2.1 and the latest beta of 2.2

The following code repos the issue.  Uncomment the code in parent to get the test to run successfully.

```cs
using System;
using System.Collections.Generic;
using System.Linq;
using System.Text;
using System.Threading.Tasks;
using Xunit;

namespace ClassLibrary1
{
    public class Foo
    {
    }

    public class Child : Parent<Foo>
    {
        public Child(Foo fixture)
            : base(fixture)
        { }

        public override void Test()
        {
            base.Test();
        }
    }

    public abstract class Parent<TFixture> : GrandParent<TFixture>
        where TFixture : class
    {
        protected Parent(TFixture fixture) : base(fixture)
        { }

        /*public override void Test()
        {
            base.Test();
        }*/
    }

    public abstract class GrandParent<TFixture> : IClassFixture<TFixture>
        where TFixture : class
    {
        protected GrandParent(TFixture fixture)
        {
        }

        [Fact]
        public virtual void Test()
        {
            Assert.True(1 == 1);
        }
    }
}
```

smitpatel:
Resolves #1068 

Issue: When trying to find base method for `CustomAttributeData`, we checked base type of method's declaring type to find matching base method but if the base type did not define the method, we returned null. In multiple level of hierarchy (`GrandParent` -> `Parent` -> `Child`), if the method is overridden in `Child` class without `FactAttribute` on it (assuming it would be carried from the base method), we would try to find base method on `Parent` Class. But if the method is defined on `GrandParent` and not overridden in `Parent` class then we will not be able to locate the base method hence associated `CustomAttributeData` is not found which causes test to be ignored.

Fix is to go multiple levels towards base till we find base method.

### DIFF
--- a/src/xunit.execution/Sdk/Reflection/ReflectionMethodInfo.cs
+++ b/src/xunit.execution/Sdk/Reflection/ReflectionMethodInfo.cs
@@ -128,19 +128,24 @@ namespace Xunit.Sdk
             if (!method.IsVirtual)
                 return null;
 
-            var baseType = method.DeclaringType.GetTypeInfo().BaseType;
-            if (baseType == null)
-                return null;
-
             var methodParameters = method.GetParameters();
             var methodGenericArgCount = method.GetGenericArguments().Length;
 
-            foreach (MethodInfo m in baseType.GetMatchingMethods(method))
+            var currentType = method.DeclaringType;
+
+            while (currentType != typeof(object))
             {
-                if (m.Name == method.Name &&
-                    m.GetGenericArguments().Length == methodGenericArgCount &&
-                    ParametersHaveSameTypes(methodParameters, m.GetParameters()))
-                    return m;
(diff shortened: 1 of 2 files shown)


smitpatel:
@Eilon

bradwilson:
Thanks!

Issue #1291, opened 2017-05-19
*TestCase and  messages types should be marked as serializable


safern:
In our test suite for .NET Core we currently have two custom discoverers ([ConditionalFactDiscoverer](https://github.com/dotnet/buildtools/blob/master/src/xunit.netcore.extensions/Discoverers/ConditionalFactDiscoverer.cs#L13) and [ConditionalTheoryDiscoverer](https://github.com/dotnet/buildtools/blob/master/src/xunit.netcore.extensions/Discoverers/ConditionalTheoryDiscoverer.cs#L13)) this discoverers help discover test cases with our custom [ConditionalFactAttribute](https://github.com/dotnet/buildtools/blob/master/src/xunit.netcore.extensions/Attributes/ConditionalFactAttribute.cs#L12) and [ConditionalTheoryAttribute](https://github.com/dotnet/buildtools/blob/master/src/xunit.netcore.extensions/Attributes/ConditionalTheoryAttribute.cs#L12) which main goal is to have a condition and if that condition is false it would be marked as skipped in our discoverers.

When running our tests in netcoreapp this works fine as it doesn't support `AppDomains`, the problem comes to when running the tests against .NET Framework.

Since .NET Framework supports `AppDomains` and our discoverers live in a different `AppDomain` than xunit's   main `AppDomain`, when we return the discovered or skipped test cases this will be added to a message queue that will later be Dispatched through [MessageBus.Dispatch](https://github.com/xunit/xunit/blob/master/src/xunit.execution/Sdk/MessageBus.cs#L35). Since the objects where queued on a different `AppDomain` when trying to dispatch the message and call its delegate (to do whatever action it has to do depending on the message type) the runtime will try to marshal this objects, but since they are not marked as `Serializable` we will get a `SerializationException` which xunit currently ignores and treats this test as if it was not discovered, so we get no test execution. 

[...]

PR #1292, opened 2017-05-19
Mark messages and TestCase objects as serializable

safern:
Fixes: #1291

FYI: @danmosemsft


bradwilson:
I'm sorry, but we don't consider these things as "broken". They intentionally use MBRO instead of serialization.

safern:
Thanks @bradwilson. Will use another approach on our side.

PR #1312, opened 2017-06-05
Support value types/ienumerable in MemberDataAttributeBase

### LINKED ISSUE #1288: Support value types in `MemberDataAttributeBase`
### Problem

[`MemberDataAttributeBase`](https://github.com/xunit/xunit/blob/bccfcccf26b2c63c90573fe1a17e6572882ef39c/src/xunit.core/MemberDataAttributeBase.cs) does not support collection of value types in the result of the member.
A member returning `IEnumerable<int>` will throw an `ArgumentException` in [`GetData`](https://github.com/xunit/xunit/blob/bccfcccf26b2c63c90573fe1a17e6572882ef39c/src/xunit.core/MemberDataAttributeBase.cs#L71) because it is impossible to cast the `IEnumerable<int>` into `IEnumerable<object>`.

This issue may be a problem when creating new implementation of `MemberDataAttributeBase`.

### Steps to reproduce

```csharp
public sealed class SingleMemberDataAttribute : ImprovedMemberDataAttributeBase
{
    public SingleMemberDataAttribute(string memberName, params object[] parameters)
        : base(memberName, parameters) { }

    protected override object[] ConvertDataItem(MethodInfo testMethod, object item)
    {
        return new object[] { item };
    }
}

private static IEnumerable<int> DataProviderMethod()
{
    yield return 1;
    yield return 2;
}

[Theory]
[SingleMemberData(nameof(DataProviderMethod))]
public void ValueTypeTestWithAttribute(int value)
{
    // ...
}
```

### Suggested solution

In [`GetData`](https://github.com/xunit/xunit/blob/bccfcccf26b2c63c90573fe1a17e6572882ef39c/src/xunit.core/MemberDataAttributeBase.cs#L69-L73), replace the lines:
```csharp
var dataItems = obj as IEnumerable<object>;
if (dataItems == null)
    throw new ArgumentException($"Property {MemberName} on {type.FullName} did not return IEnumerable<object>");

return dataItems.Select(item => ConvertDataItem(testMethod, item));
```

by:
```csharp
var dataItems = obj as IEnumerable;
if (dataItems == null)
    throw new ArgumentException($"Property {MemberName} on {type.FullName} did not return IEnumerable");

return dataItems.Cast<object>().Select(item => ConvertDataItem(testMethod, item));
```

hughbe:
Fixes #1288

### DIFF
--- a/src/xunit.core/MemberDataAttributeBase.cs
+++ b/src/xunit.core/MemberDataAttributeBase.cs
@@ -1,4 +1,5 @@
 using System;
+using System.Collections;
 using System.Collections.Generic;
 using System.Linq;
 using System.Reflection;
@@ -8,7 +9,7 @@ namespace Xunit
 {
     /// <summary>
     /// Provides a base class for attributes that will provide member data. The member data must return
-    /// something compatible with <see cref="IEnumerable&lt;Object&gt;"/>.
+    /// something compatible with <see cref="IEnumerable"/>.
     /// Caution: the property is completely enumerated by .ToList() before any test is run. Hence it should return independent object sets.
     /// </summary>
     [CLSCompliant(false)]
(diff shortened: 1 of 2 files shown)


src/xunit.core/MemberDataAttributeBase.cs
```diff
@@ -66,11 +67,11 @@ public override IEnumerable<object[]> GetData(MethodInfo testMethod)
             if (obj == null)
                 return null;
 
-            var dataItems = obj as IEnumerable<object>;
+            var dataItems = obj as IEnumerable;
             if (dataItems == null)
                 throw new ArgumentException($"Property {MemberName} on {type.FullName} did not return IEnumerable<object>");
```
onovotny: Nit: Exception message stills says `IEnumerable<object>`, should now just be `IEnumerable`

hughbe:
@bradwilson does this look good to you?

hughbe:
Friendly ping @bradwilson :)

bradwilson:
Danke!

PR #1320, opened 2017-06-14
Stop on first test failure

EricJFisher:
Based heavily off of @EMarshal 's PR.

Uses the message bus to add support to console runner (issue #23) and the MSBuild runner (issue #24) for stopping on the first test failure.

In addition to this, added support to set StopOnFail via config / json files. (will require the schema files to be updated in addition to the work done here)

Note: When a test fails the assembly containing that test will stop and tests on assemblies not yet running will also not be executed. Any assemblies already running tests will continue to completion. 
(I entertained several solutions to this limitation, however; all attempted solutions had side effects I wasn't willing to accept)


bradwilson:
Reworked and pushed in c7140cbc. Thanks!

PR #1349, opened 2017-07-06
Checks for target runtime before discovery/execution

### LINKED ISSUE #1348: Run selected tests fail in NET 4.7 (visual studio)
## Steps to reproduce
1. Install [Windows 10 16226](https://blogs.windows.com/windowsexperience/2017/06/21/announcing-windows-10-insider-preview-build-16226-pc/) preview build
2. Install VS (tried with VS 2015 Update 3 Community and VS 2017 latest bits)
3. Create a hello world xunit desktop test project, use xunit 2.2 adapters
4. Try discovery, run all or run selected
5. These operations fail with an error. Run All throws warning as it sees two adapters (xunit.vs.testadapter and xunit.vs.dotnetcore.testadapter) with same URI, it moves on to execute test with the dotnetcore one and fails.

## Root cause
It happens because .NET 4.7 can now load xunit.visualstudio.dotnetcore.testadapter.dll (earlier versions of NETFx were not compatible with netstandard 1.6 API set).

## Workarounds
1. Users on an xunit version < 2.2 shouldn't face this issue since it packages a single adapter

## Possible fix
1. Use separate executor URIs for the desktop and .net core adapters
2. Allow the adapter to bail out if it's loaded in an incorrect runtime

PR #1349 implements this fix.
Customer report is here: https://github.com/Microsoft/vstest/issues/811

codito:
This change allows an adapter to bail out if it's loaded in incorrect target runtime. It can happen if .NET core adapter is loaded to run a desktop test for instance (this occurs consistently if user has .NET 4.7 installed).

Provide different executor URI identifiers for the desktop and dotnet core adapters.

Fixes #1348.

### DIFF
--- a/src/xunit.runner.visualstudio/Constants.cs
+++ b/src/xunit.runner.visualstudio/Constants.cs
@@ -2,6 +2,14 @@
 {
     public static class Constants
     {
-        public const string ExecutorUri = "executor://xunit/VsTestRunner2";
+#if NET452
+        public const string ExecutorUri = "executor://xunit/VsTestRunner2/desktop";
+#elif NETCOREAPP1_0
+        public const string ExecutorUri = "executor://xunit/VsTestRunner2/netcoreapp";
+#elif WINDOWS_UAP
+        public const string ExecutorUri = "executor://xunit/VsTestRunner2/uap";
+#else
+#error Unknown target platform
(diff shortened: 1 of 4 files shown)


bradwilson:
Thanks!

PR #1352, opened 2017-07-06
Added functionality to honor CollectSourceInformation flag from RunSettings XML

Shyam-Gupta:
Refer RFC: https://github.com/Microsoft/vstest-docs/blob/master/RFCs/0010-Source-Information-For-Discovered-Tests.md

Summary:
CollectSourceInformation = True => Test adapter needs to collect source information for the discovered tests
CollectSourceInformation = False => Test adapter doesn't need to collect source information for discovered tests. In case tests are executed from VS IDE, then Test Window will get this information using Roslyn.

By default CollectSourceInformation will be true. It will be false when:
1. VS IDE determines that it can get this information from Roslyn
2. Command line scenario: When tests are executed from vstest.console.exe, this flag will always be false unless user explicitly overrides it in the runsettings XML file.

### DIFF
--- a/src/xunit.runner.visualstudio/Utility/RunSettingsHelper.cs
+++ b/src/xunit.runner.visualstudio/Utility/RunSettingsHelper.cs
@@ -16,11 +16,13 @@ namespace Xunit.Runner.VisualStudio.TestAdapter
         /// </summary>
         public static bool DesignMode { get; private set; }
 
+        public static bool CollectSourceInformation { get; private set; }
+
         public static bool NoAutoReporters { get; private set; }
 
-/*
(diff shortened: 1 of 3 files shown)


onovotny:
Why was `ReporterSwitch` removed?

Shyam-Gupta:
@onovotny 
ReporterSwitch was already in commented out state. Its only indentation changes which came in automatically by VS.

bradwilson:
I cleaned up the RunSettingsHelper class, including enabling ReporterSwitch (understanding that nobody's listening to it yet, but that shouldn't matter).

bradwilson:
Thanks!

PR #1420, opened 2017-08-22
Add unit tests for Assert.Contains changes at xunit/assert.xunit#20

jamesqo:
/cc @bradwilson


jamesqo:
@bradwilson If it's what you were waiting for, I squashed the commits. Could you PTAL at merging?

bradwilson:
Sorry, this fell through the cracks while I was on vacation.

PR #1432, opened 2017-08-31
Performance improvements around serialization

onovotny:
This PR does two things based on some profiling

1. batches cross-appdomains calls to the serializer based on test-class instead of method. 
2. Caches `Type` to `string` in the `SerializationHelper` since those lookups were expensive. I ported a `ConcurrentDictionary` from the old TPL that's .NET 3.5-friendly for the NET 3.5 runner utility.

In my "wall time" of running 10k unit test discovery (from the `perf-tests` branch), these changes make the .NET 4.5.2 version go from approx. 30 sec to about 12 sec. Still a long time, but better.

The one downside is that with the current batching technique, there's a pause before users see output since it's not sending them over till it's done with the test class, and this pathological test has 10k tests in a single test class.

/cc @benaadams

### DIFF
--- a/src/xunit.runner.visualstudio/Sinks/VsDiscoverySink.cs
+++ b/src/xunit.runner.visualstudio/Sinks/VsDiscoverySink.cs
@@ -25,6 +25,7 @@ namespace Xunit.Runner.VisualStudio
     {
         const string Ellipsis = "...";
         const int MaximumDisplayNameLength = 447;
+        const int TestCaseDescriptorBatchSize = 100;
 
         static readonly Action<TestCase, string, string> addTraitThunk = GetAddTraitThunk();
         static readonly Uri uri = new Uri(Constants.ExecutorUri);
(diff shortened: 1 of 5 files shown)


src/xunit.runner.visualstudio/Sinks/VsDiscoverySink.cs
```diff
@@ -190,11 +190,11 @@ void HandleCancellation(MessageHandlerArgs args)
         void HandleTestCaseDiscoveryMessage(MessageHandlerArgs<ITestCaseDiscoveryMessage> args)
         {
             var testCase = args.Message.TestCase;
-            var testMethod = $"{testCase.TestMethod.TestClass.Class.Name}.{testCase.TestMethod.Method.Name}";
-            if (lastTestMethod != testMethod)
+            var testClass = $"{testCase.TestMethod.TestClass.Class.Name}";
+            if (lastTestClass != testClass)
```
onovotny: Optionally, we could also include a max count, so max batches of 500? (or some other arbitrary number)

benaadams:
This is much faster 😄

bradwilson:
There's a problem with your change in that you break the "need unique name" logic which depends on the fact that the thing we're batching is tests for a given method.

bradwilson:
I decided to push forward with batching independent of class or method. This first version has a fixed batch size of 100. We'll see how that feels. :)

PR #1446, opened 2017-09-05
Make sure output file full path exsits

luizbon:
Added a method to make sure the file path exists after the following twitter thread: https://twitter.com/xunit/status/904718357697630213


bradwilson:
Thanks!

PR #1447, opened 2017-09-09
Use TotalMilliseconds instead of Milliseconds

jnyrup:
[`Milliseconds`](https://msdn.microsoft.com/en-us/library/system.timespan.milliseconds(v=vs.110).aspx) ranges from -999 to 999, so checking for `Milliseconds < 1000` is always true,
I believe these should be `TotalMilliseconds` instead?


bradwilson:
Good catch! Thanks!

PR #1449, opened 2017-09-09
Same variable checked twice for null

jnyrup:
`originalUICulture` is checked for null in both if statements making them identical.


bradwilson:
Thanks!

Issue #1515, opened 2017-10-15
MSBuild runner does not automatically close `Xml` output file


dougbu:
1. Clone a repository that reads test result files after the `<xunit/>` target completes e.g. https://github.com/aspnet/AspNetWebStack
  This repo [includes](https://github.com/aspnet/AspNetWebStack/blob/da59ce7f2aa180515aee1e2650828a148899465d/tools/WebStack.xunit.targets#L11-L20) the following target:
  ``` xml
    <Target Name="Xunit">
        <!-- Override AppDomains default (set in xunit.runner.json). Need them with this runner. -->
        <xunit AppDomains="true" Assemblies="$(TestAssembly)" XmlV1="$(XmlPath)"/>
        <!-- Replace potentially illegal escaped characters (if they get through validations done during Save) -->
        <RegexReplace
            Files="$(XmlPath)"
            Find="&amp;#x(?&lt;char&gt;[0-9A-Fa-f]+);"
            Replace="\0x${char}"
            WarnOnNoMatch="false"/>
    </Target>
  ```
  `<RegexReplace/>` is an inline code target that [includes](https://github.com/aspnet/AspNetWebStack/blob/da59ce7f2aa180515aee1e2650828a148899465d/tools/WebStack.tasks.targets#L108) the following code:
  ``` c#
  string originalText = File.ReadAllText(fullPath);
  ```
2. Update the repo to use the 2.3.0 version of the xunit.runner.msbuild package.
2. If necessary, change the `<xunit/>` task to use `Xml` and not `XmlV1`.
3. Build a target that runs tests e.g. `UnitTest` in the [aspnet/AspNetWebStack](https://github.com/aspnet/AspNetWebStack) case

#### Expected
A clean build
#### Actual
A number of failures about file access:
```
...\AspNetWebStack\tools\WebStack.xunit.targets(15,9): error : The process cannot access the file
'...\AspNetWebStack\bin\CodeAnalysis\test\TestResults\WebMatrix.WebData.Test-XunitResults.xml'
because it is being used by another process.
```
#### Regression since 2.3 Beta 1
Haven't tested exhaustively but no similar problem is visible when using the `XmlV1` task parameter. The `TransformFactory` has the necessary `using` to close the result file immediately. Similarly, I found no similar problem when using the 2.0.0 MSBuild runner.

[...]

PR #1516, opened 2017-10-15
Close `Xml` result file as soon as possible

dougbu:
- #1515

### DIFF
--- a/src/xunit.runner.msbuild/xunit.cs
+++ b/src/xunit.runner.msbuild/xunit.cs
@@ -200,7 +200,8 @@ namespace Xunit.Runner.MSBuild
             if (NeedsXml)
             {
                 if (Xml != null)
-                    assembliesElement.Save(new FileStream(Xml.GetMetadata("FullPath"), FileMode.OpenOrCreate));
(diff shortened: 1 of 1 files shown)


dougbu:
This PR is incomplete because I'm not sure how to test the change. GC can happen at any time, making the original problem difficult to reproduce in a unit test -- especially when other tests are running in the same process. Suggestions?

src/xunit.runner.msbuild/xunit.cs
```diff
@@ -200,7 +200,8 @@ public override bool Execute()
             if (NeedsXml)
             {
                 if (Xml != null)
-                    assembliesElement.Save(new FileStream(Xml.GetMetadata("FullPath"), FileMode.OpenOrCreate));
+                    using (var xmlStream = new FileStream(Xml.GetMetadata("FullPath"), FileMode.OpenOrCreate, FileAccess.Write))
```
dougbu: I'm addressing a nit: The `FileStream` constructor defaults to `FileAccess.ReadWrite` and that's not necessary here. But, passing the additional argument isn't mandatory.

bradwilson:
I'm fine accepting this without a unit test.

bradwilson:
Thanks!

PR #1535, opened 2017-10-26
Stacktraces to delight

benaadams:
For `NET452` and `NETSTANDARD2_0` focus on the call stack and methods rather than compiler and task noise

Resolves generics; converts il entries to C#

Drops
```
--- End of stack trace from previous location where exception was thrown ---
--- End of stack trace from previous location where exception was thrown ---
--- End of stack trace from previous location where exception was thrown ---
```
Drops
```
System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw() in ExceptionServicesCommon.cs
System.Runtime.CompilerServices.TaskAwaiter.ThrowForNonSuccess(Task task) in TaskAwaiter.cs
System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task) in TaskAwaiter.cs
System.Runtime.CompilerServices.TaskAwaiter.ValidateEnd(Task task) in TaskAwaiter.cs
System.Runtime.CompilerServices.TaskAwaiter.GetResult() in TaskAwaiter.cs
```
De-garbles iterators and async

From
```
Program.<Iterator>d__4.MoveNext()
Microsoft.AspNetCore.Mvc.Razor.RazorView+<RenderPageAsync>d__15.MoveNext()
```
To
```
Program.Iterator(long count, int value)+MoveNext()
Microsoft.AspNetCore.Mvc.Razor.RazorView.RenderPageAsync(IRazorPage page, ViewContext context, bool invokeViewStarts)
```

Don't need to merge was an academic exploration :)

### DIFF
--- a/src/common/ExceptionUtility.cs
+++ b/src/common/ExceptionUtility.cs
@@ -1,5 +1,15 @@
 ﻿using System;
+using System.Collections;
 using System.Collections.Generic;
+#if NET452 || NET35 || NETSTANDARD2_0
+using System.Globalization;
+using System.Reflection;
+using System.Runtime.CompilerServices;
+#if NET452 || NETSTANDARD2_0
+using System.Runtime.ExceptionServices;
+#endif
+using System.Text;
+#endif
 using Xunit.Abstractions;
 
 #if XUNIT_FRAMEWORK
@@ -169,10 +179,10 @@ namespace Xunit
         static void ConvertExceptionToFailureInformation(Exception ex, int parentIndex, List<string> exceptionTypes, List<string> messages, List<string> stackTraces, List<int> indices)
         {
             var myIndex = exceptionTypes.Count;
-
+            
             exceptionTypes.Add(ex.GetType().FullName);
             messages.Add(ex.Message);
-            stackTraces.Add(ex.StackTrace);
+            stackTraces.Add(ex.GetCleanStackTrace());
             indices.Add(parentIndex);
 
 #if XUNIT_FRAMEWORK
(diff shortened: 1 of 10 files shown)


bradwilson:
This is a cool idea.

I need to contemplate whether the actual implementation could introduce a new execution library, as this could end up being a pretty significant change which might break older runners. An alternative might be to look for the APIs you want at runtime and manufacture efficient ways to call them when they are (and skip them when they're not).

One example where we do this today: https://github.com/xunit/xunit/blob/master/src/xunit.execution/Sdk/Utility/ExecutionContextHelper.cs

benaadams:
Wasn't quite sure on framework versions and how they work with the runner

e.g. 

Lib under testing targets .NET Standard 1.1; which means the StackTrace apis aren't available

However you'd want the runner to be running under 4.5.x+; .NetStandard 2.0; .NET CoreApp 2.0 so it could use the apis on output (and use e.g. the .NET Standard 2.0 xUnit).

Make sense? Still have a knowledge gap with it all works with what's actually the exe that runs the tests...

bradwilson:
The runner chooses the target platform; some support multiple (like the VS runner), and some only support one, and maybe even a very specific one (for example, we have three console runners: net452+, netcoreapp1.x, and netcoreapp2.x). When runners support multiple platforms (like `dotnet xunit`), the look to the unit test project to tell them which platform to use; as such, we don't explicitly support unit tests targeting `netstandard`, since it's not a platform (just an API set).

The `netstandard1.1` execution library today serves pretty much every platform other than desktop CLR, including UWP and both flavors of `netcoreapp`. I assume that these APIs might actually exist (and runtime) in pre-`netstandard2` platforms, even if they weren't exposed in .NET Standard.

As a result, lighting the functionality up dynamically really is the right way to do this.

I'm not necessarily asking you to do it. I appreciate the spike, and could take it from here if you want. Just leave the PR open so we can steal the important bits. 😁

benaadams:
>  and could take it from here if you want

Would appreciate that; 90% of this PR was trying to get the cross targeting working rather than the code - need some time to recover

benaadams:
I have come to appreciate MSBuild Log Viewer 😁

benaadams:
Working on a lib that does this better (stack traces, not testing 😄 )

benaadams:
Working on a lib that does this better (stack traces, not testing 😄 )

All sorts of other things to clean up; lambdas, local functions, tuples etc

benaadams:
Lib version might be deviating from standard stack traces; including return types and whatnot

bradwilson:
That looks awesome.

I assume I'm going to have to copy the code and do the dynamic discovery myself... 😉

benaadams:
> copy the code

Would a source only package solve that? https://github.com/benaadams/Ben.Demystifier/issues/2

bradwilson:
> Would a source only package solve that?

Not really. That's useful for when someone wants to take your code as-is without modification. I would need modification.

bradwilson:
Closing this for now. At a future point, I may consider pulling in the Ben code. :)

benaadams:
Resolved for coreclr https://github.com/dotnet/corefx/issues/24627

PR #1537, opened 2017-10-29
Update usage information for -method

jcouv:
Following up on https://github.com/xunit/xunit/pull/766 (allowing wildcards)


bradwilson:
Thanks!

PR #1546, opened 2017-11-03
Conversion of strings to Guid and DateTime argument values

adamralph:
I was going to [add this to xbehave](https://github.com/xbehave/xbehave.net/issues/155), but then I thought I'd speculate on whether you'd want it in xunit (xbehave would inherit the feature with no work).

### DIFF
--- a/src/xunit.execution/Sdk/Reflection/Reflector.cs
+++ b/src/xunit.execution/Sdk/Reflection/Reflector.cs
@@ -34,13 +34,10 @@ namespace Xunit.Sdk
 
             if (args.Length == types.Length)
             {
-                var converted = new List<object>(args.Length);
                 for (var idx = 0; idx < args.Length; idx++)
                 {
-                    converted.Add(ConvertArgument(args[idx], types[idx]));
+                    args[idx] = ConvertArgument(args[idx], types[idx]);
                 }
-
-                return converted.ToArray();
             }
 
             return args;
--- a/src/xunit.execution/Sdk/Reflection/Reflector.cs
+++ b/src/xunit.execution/Sdk/Reflection/Reflector.cs
@@ -1,5 +1,6 @@
 using System;
 using System.Collections.Generic;
+using System.Globalization;
 using System.Linq;
 using System.Reflection;
 using Xunit.Abstractions;
@@ -64,6 +65,16 @@ namespace Xunit.Sdk
                         if (type == typeof(Guid))
                         {
                             return Guid.Parse(arg.ToString());
+                        }
+
+                        if (type == typeof(DateTime))
+                        {
+                            return DateTime.Parse(arg.ToString(), CultureInfo.InvariantCulture);
+                        }
+
+                        if (type == typeof(DateTimeOffset))
+                        {
+                            return DateTimeOffset.Parse(arg.ToString(), CultureInfo.InvariantCulture);
                         }
 
                         return Convert.ChangeType(arg, type);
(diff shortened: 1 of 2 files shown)


src/xunit.execution/Sdk/Reflection/Reflector.cs
```diff
@@ -33,10 +34,13 @@ public static object[] ConvertArguments(object[] args, Type[] types)
 
             if (args.Length == types.Length)
             {
+                var converted = new List<object>(args.Length);
                 for (var idx = 0; idx < args.Length; idx++)
                 {
-                    args[idx] = ConvertArgument(args[idx], types[idx]);
+                    converted.Add(ConvertArgument(args[idx], types[idx]));
```
adamralph: I had to change this to avoid:

> System.ArrayTypeMismatchException : Attempted to access an element as a type incompatible with the array.

src/xunit.execution/Sdk/Reflection/Reflector.cs
```diff
@@ -33,10 +34,13 @@ public static object[] ConvertArguments(object[] args, Type[] types)
 
             if (args.Length == types.Length)
             {
+                var converted = new List<object>(args.Length);
```
bradwilson: What is the purpose of introducing the List<T> here?

src/xunit.execution/Sdk/Reflection/Reflector.cs
```diff
@@ -58,6 +62,21 @@ internal static object ConvertArgument(object arg, Type type)
...
                     else
                     {
+                        if (type == typeof(Guid))
+                        {
+                            return Guid.Parse(arg.ToString());
+                        }
+
+                        if (type == typeof(DateTime))
+                        {
+                            return DateTime.Parse(arg.ToString(), CultureInfo.InvariantCulture);
```
adamralph: It would have to be stated in the docs the date time literal must be in an invariant culture compatible format.

src/xunit.execution/Sdk/Reflection/Reflector.cs
adamralph: @bradwilson our comments crossed: https://github.com/xunit/xunit/pull/1546#discussion_r148840099

src/xunit.execution/Sdk/Reflection/Reflector.cs
bradwilson: args is defined as `object[]`, so this confuses me.

adamralph: That makes two of us 😕

bradwilson: Was it failing in production, or only in the tests? It might be because your tests aren't passing an object[], but rather a string[] (implicitly typed).

adamralph: You hit the jackpot. I'll push fixups.

adamralph:
@bradwilson I pushed fixups. Say the word and I'll `--autosquash` them.

bradwilson:
No worries about that, I use the squash merge on GitHub.

bradwilson:
Thanks!

adamralph:
🎉 will this be in 2.4.0?

bradwilson:
Yes, it is available now via our CI MyGet feed, and will be in 2.4 beta 1.

src/xunit.execution/Sdk/Reflection/Reflector.cs
marcind: Hate to dredge up an old review @bradwilson, but does it make sense for the `arg`'s `toString` result to be used here? Shouldn't the parsing only happen if `arg` is actually a `string`?

Issue #1185, opened 2017-03-27
'Test Failure' reported as 'Success' in TeamCity when message contains Unicode ’ char (RIGHT SINGLE QUOTATION MARK)


slawomir-brzezinski-at-interxion:
Using xUnit 2.2 and TeamCity 10.0.4 on Windows.

## Reproduction
Please see [the reproduction repo](https://github.com/slawomir-brzezinski-at-interxion/xUnitInTeamCity.ReproForUnicodeEscapingIssue). It contains [one test](https://github.com/slawomir-brzezinski-at-interxion/xUnitInTeamCity.ReproForUnicodeEscapingIssue/blob/master/xUnitInTeamCity.ReproForUnicodeEscapingIssue.Tests/IssueReproTests.cs#L9).

1. Point TeamCity at the repo
1. Add a 'Nuget Restore step' in the beginning
1. Add an off-the-shelf xUnit runner step with '**/*.Tests.dll'
1. Observe:
Despite the runner exits with non-zero code (thus correctly failing the build), the TeamCity report **shows the test itself as successful**.

## Analysis
Inspecting the log shows a message:
```
##teamcity[testFailed name='xUnitInTeamCity.ReproForUnicodeEscapingIssue.Tests.IssueReproTests.FailWithMessageContainingNonAsciiChars' details='Assert.Equal() Failure|r|n            (pos 0)|r|nExpected: ' +Ø%D|r|nActual:   anything different so that test fails|r|n            (pos 0)|r|n   at xUnitInTeamCity.ReproForUnicodeEscapingIssue.Tests.IssueReproTests.FailWithMessageContainingNonAsciiChars() in d:\Teamcity\BuildAgent\work\c80283b4cb57ca79\xUnitInTeamCity.ReproForUnicodeEscapingIssue.Tests\IssueReproTests.cs:line 25' flowId='4862240aeba54317941ec7648b984516']
Incorrect property name.
Valid property list format is (name( )*=( )*'escaped_value'( )*)* where escape symbol is "|"
```

[...]

PR #1578, opened 2017-11-20
Escape utf8 characters in teamcity output

hboisselle:
A fix for https://github.com/xunit/xunit/issues/1185 with some performance tweak as well


bradwilson:
Thanks!

Issue #1607, opened 2018-01-04
Use implicit/explicit type conversions declared on theory arguments' types


ascott18:
Related to https://github.com/xunit/xunit/issues/1013 - The PR for that issue, https://github.com/xunit/xunit/pull/1020, implemented calls to user-defined conversions, but only did so for conversions that are declared on the theory's parameter type. It did not check for conversions that are defined on the actual argument's type. Since C# allows you to define a conversion operator on either the class on the left-hand side or the right-hand side, the solution in https://github.com/xunit/xunit/pull/1020 is only half complete.

The use case is essentially the same as https://github.com/xunit/xunit/issues/1013 - wanting to make the names of tests more concise and relevant to the pertinent data. However, my particular use case necessitated that my user-defined conversions be placed on the arguments supplied to my tests rather than the parameters themselves.

Pull request incoming momentarily.

GrantErickson:
Really helpful, thanks for adding this @ascott18 and @bradwilson.

PR #1608, opened 2018-01-04
Support implicit & explicit conversions declared on the argument's type

ascott18:
This resolves https://github.com/xunit/xunit/issues/1607

### DIFF
--- a/src/xunit.execution/Sdk/TypeUtility.cs
+++ b/src/xunit.execution/Sdk/TypeUtility.cs
@@ -137,24 +137,14 @@ namespace Xunit.Sdk
 
             // Implicit & explicit conversions to/from a type can be declared on either side of the relationship.
             // We need to check both possibilities.
-            object convertedArgumentValue;
-
-            // Check for conversions declared on the parameter's type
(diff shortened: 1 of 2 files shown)


src/xunit.execution/Sdk/TypeUtility.cs
```diff
@@ -135,25 +135,51 @@ private static object TryConvertObject(object argumentValue, Type parameterType)
...
+                return convertedArgumentValue;
+
+            // Can't convert object. We don't need to throw anything here, since MethodInfo.Invoke does
+            return argumentValue;
+        }
+
+        private static bool TryPerformDefinedConversions(
+            object argumentValue,
+            Type conversionDeclaringType,
+            out object convertedArgumentValue)
```
bradwilson: Out is evil. There's no reason for out and bool return here, because a null value isn't a valid conversion, so it's a perfectly fine sentinel value for failed conversion.

Please re-write this to return the object, and above use null coalescing, like:

```
return PerformDefinedConversion(argumentValue, parameterType)
    ?? PerformDefinedConversion(argumentValue, argumentValueType)
    ?? argumentValue;
```

bradwilson:
Thanks!

Issue #1624, opened 2018-01-31
xUnit.Core v2.3.1 sets IsTestProject build property blocking consuming projects to be packed


mishra14:
xUnit.Core v2.3.1 package contains `build/xunit.core.props` and `buildMultitargeting/xunit.core.props`.

As part of https://github.com/xunit/xunit/commit/6ef433fe9a45469a3763efd6e8188bb627c4ef10, the props file was changed to set `IsTestProject`. This breaks package consumers from being able to pack their projects as part of NuGet pack [logic](https://github.com/NuGet/NuGet.Client/blob/dev/src/NuGet.Core/NuGet.Build.Tasks.Pack/NuGet.Build.Tasks.Pack.targets#L33).

To fix this, the props file should contain another property to allow consuming projects to be packed - 

`<IsPackable>true</IsPackable>`

mishra14:
If this is an unintentional change and needs to be fixed, I have raised a PR for this - https://github.com/xunit/xunit/pull/1625

cc: @bradwilson

onovotny:
Why would a project consuming `xunit.core` need to be packed? It's a unit test project?

mishra14:
Users can have their own scenarios for packing a project which have been working till now. An [example](https://developercommunity.visualstudio.com/content/problem/186365/build-of-nuget-package-on-compile-stops-working-if.html).

The root issue is that xUnit should not prevent consuming projects from being packed. Further, this is a breaking change in behavior from v 2.2.0 to v2.3.1.

@onovotny what is the reason that xUnit should explicitly prevent consuming projects from being packed?

bradwilson:
Users can override this, since .props files are effectively included before their own properties. This change should be unnecessary.

> The root issue is that xUnit should not prevent consuming projects from being packed.

This value is set by VSTest, so even if we removed it, it would still end up being set as soon as you include Microsoft.Net.SDK.Test. This value is set at the guidance of the VSTest team that unit test projects should not be packable by default (guidance we agree with).

mishra14:
Fair enough. I will close the issue and PR.

dasMulli:
@mishra14 about this.. why does NuGet "know" about `IsTestProject`? That's hardly NuGet's concern..

mishra14:
[...]

PR #1625, opened 2018-01-31
Adding IsPackable to xunit.core.props to allow consuming projects to be packable

mishra14:
Fixes: https://github.com/xunit/xunit/issues/1624

Setting `<IsTestProject>true</IsTestProject>` prevents consuming projects to be packed based on NuGet pack [logic](https://github.com/NuGet/NuGet.Client/blob/dev/src/NuGet.Core/NuGet.Build.Tasks.Pack/NuGet.Build.Tasks.Pack.targets#L33).

This PR adds a new property - `<IsPackable>true</IsPackable>`. This property will aloow the package consumers to be able to pack their projects.

### DIFF
--- a/src/xunit.core/build/xunit.core.props
+++ b/src/xunit.core/build/xunit.core.props
@@ -7,7 +7,8 @@
 
   <PropertyGroup>
     <DebugSymbols>true</DebugSymbols>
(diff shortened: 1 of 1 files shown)


bradwilson:
This behavior is by design. We will not be accepting this PR.

Issue #1627, opened 2018-02-02
Create Runner reporter for VSTS


clairernovotny:
Detect and publish test results to VSTS like we do for AppVeyor and TeamCity.

Test Run API(s) : https://www.visualstudio.com/en-us/docs/integrate/api/test/runs 
-	API(s) to query, create or update test runs

Test Results API(s) : https://www.visualstudio.com/en-us/docs/integrate/api/test/results 
-	API(s) to query, add or update test results against a run

Sample publishing csv:
https://github.com/ManojBableshwar/VstsTestRestApiSamples/blob/master/PublishResultsFromCsvWithAttachments/PublishResultsFromCsvWithAttachments.cs

PR #1630, opened 2018-02-02
Vsts reporter

onovotny:
This is the implementation for #1627.

Note it doesn't work yet since getting the access token in YAML builds doesn't appear to be complete yet:
https://github.com/Microsoft/vsts-agent/blob/master/docs/preview/yamlgettingstarted-token.md

Pending verification once the access token is available.

To test: create a YAML based build of xunit in a VSTS instance and point to the `.vsts.ci.yml` file as the yaml file. It should build and you should see test results in the tests area of the build results.


src/xunit.runner.reporters/VstsClient.cs
```diff
@@ -0,0 +1,310 @@
...
+        {
+            int? runId = null;
+            try
+            {
+                runId = await CreateTestRun();
+
+                while (!shouldExit || !addQueue.IsEmpty || !updateQueue.IsEmpty)
+                {
+                    workEvent.Wait();   // Wait for work
+                    workEvent.Reset();  // Reset first to ensure any subsequent modification sets
```
bradwilson: This is a race condition. If you want this behavior, switch from `ManualResetEventSlim` to `AutoResetEventSlim`.

src/xunit.runner.reporters/VstsClient.cs
```diff
@@ -0,0 +1,310 @@
...
+
+                    if (previousErrors)
+                        break;
+
+                    // We have to do add's before update because we need the test id from the add to inject into the update
+                    await SendTestResults(true, runId.Value, aq.ToArray()).ConfigureAwait(false);
+                    await SendTestResults(false, runId.Value, uq.ToArray()).ConfigureAwait(false);
+                }
+            }
+            catch { }
```
bradwilson: This should log a diagnostic message, at a minimum.

src/xunit.runner.reporters/VstsClient.cs
```diff
@@ -0,0 +1,310 @@
...
+            finally
+            {
+                try
+                {
+                    if (runId.HasValue)
+                        await FinishTestRun(runId.Value);
+                }
+                catch
```
bradwilson: Diagnostic message

src/xunit.runner.reporters/VstsReporter.cs
```diff
@@ -0,0 +1,34 @@
...
+            => "vsts";
+
+        public IMessageSink CreateMessageHandler(IRunnerLogger logger)
+        {
+            var accessToken = EnvironmentHelper.GetEnvironmentVariable("VSTS_ACCESS_TOKEN");
+            var collectionUri = EnvironmentHelper.GetEnvironmentVariable("SYSTEM_TEAMFOUNDATIONCOLLECTIONURI");
+            var teamProject = EnvironmentHelper.GetEnvironmentVariable("SYSTEM_TEAMPROJECT");
+
+            //build id is the id associated with the build number, which we will use to associate the test run with
+            var buildId = Convert.ToInt32(EnvironmentHelper.GetEnvironmentVariable("BUILD_BUILDID"));
```
bradwilson: Seems like all these environment variables are needed, so shouldn't they all be detected for above?

src/xunit.runner.reporters/VstsClient.cs
onovotny: We use the same pattern in `AppVeyorClient`, I assume it needs updating there too?

src/xunit.runner.reporters/VstsReporter.cs
onovotny: Those are all defined as being present on the build agent by: https://docs.microsoft.com/en-us/vsts/build-release/concepts/definitions/build/variables?tabs=batch

Question is if it's worth checking all of them? The important one is to detect we are running on the agent and that the access token is properly set.

I'm not sure about the name, should we specify `VSTS_ACCESS_TOKEN` or something more specific like `XUNIT_ACCESS_TOKEN`?

onovotny:
Confirmed working now!

image

Issue #1629, opened 2018-02-02
Setting maxParallelThreads to -1 in Json configuration doesn't work


BrennanConroy:
Didn't make a repro project since its so simple, but I can if you really want.

Our `xunit.runner.json` is as follows:
```json
{
  "longRunningTestSeconds": 30,
  "diagnosticMessages": true,
  "maxParallelThreads": -1
}
```
However, when running tests the output displays "(parallel test collections = on, max threads = 2)"

Looking through the code, it looks like your json parsing is ignoring `-1` and not setting the value in `TestAssemblyConfiguration` and later when reading it will be null and default to `Environment.ProcessorCount`
https://github.com/xunit/xunit/blob/0337edb69bc9830132c96249ae52004eb9dc70b2/src/xunit.runner.utility/Configuration/ConfigReader_Json.cs#L58

anurse:
Repro project: https://github.com/anurse/ReproXunitMaxConcurrency

1. `cd ReproXunitMaxConcurrency`
2. `dotnet xunit -fxversion 2.0.0` (needed the `-fxversion` because for some reason xunit was trying to find the version "2.0" instead of "2.0.0", might be a different bug ;))

Expected (checked by adding `-maxthreads unlimited` to the command line) 

```
... bunch of internal diagnostics ...
  Discovering: ReproXunitMaxConcurrency (method display = ClassAndMethod)
  Discovered:  ReproXunitMaxConcurrency (running 1 test case)
  Starting:    ReproXunitMaxConcurrency (parallel test collections = on, max threads = unlimited)
```

Actual (when using the JSON setting and NOT using `-maxthreads unlimited` on the command line)

```
... same internal diagnostics ...
  Discovering: ReproXunitMaxConcurrency (method display = ClassAndMethod)
  Discovered:  ReproXunitMaxConcurrency (running 1 test case)
  Starting:    ReproXunitMaxConcurrency (parallel test collections = on, max threads = 12)
```

Note the max threads is `12`

BrennanConroy:
Looks like the same is true for `[assembly: CollectionBehavior(MaxParallelThreads = n)]`
It isn't being considered when running the above scenario

@bradwilson

bradwilson:
`-maxthreads` overrides any other value.

BrennanConroy:
>Actual (when using the JSON setting and NOT using -maxthreads unlimited on the command line)

onovotny:
Closing as fixed by #1632

PR #1632, opened 2018-02-05
Fix Json config to honor -1 maxParallelThreads

BrennanConroy:
https://github.com/xunit/xunit/issues/1629

### DIFF
--- a/src/xunit.runner.utility/Configuration/ConfigReader_Json.cs
+++ b/src/xunit.runner.utility/Configuration/ConfigReader_Json.cs
@@ -55,7 +55,7 @@ namespace Xunit
                             if (numberValue != null)
                             {
                                 int maxParallelThreads;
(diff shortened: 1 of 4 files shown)


anurse:
@bradwilson you able to take a quick look? We'd like to try this out to fix an issue we're seeing in some tests, and having a build on your MyGet would be the easiest way to do that :)

anurse:
🎉

BrennanConroy:
Would it be possible to push a preview build with this to your MyGet feed?

bradwilson:
Yeah, sorry about that. I changed the build # format slightly, and that caused MyGet to delete every package that got uploaded after the change. A month ago. Thanks MyGet. 😝

bradwilson:
Build is on MyGet: `2.4.0-beta.1.build3945`. Took a little while to clean up MyGet, and then the code signing service disappeared for a bit. 🙄

Issue #1620, opened 2018-01-22
Add Overload to Assert.Contains for Dictionary


chrisoverzero:
In projects I've worked on or contributed to, I'm frequently adding an overload to `Assert.Contains` taking a dictionary key and a dictionary in which to look for that key. In the style of `Assert.Single`, it's useful to have it return the value associated with that key for further assertion.

Like this:

```csharp
var dictionary = new Dictionary<string, int>
{
    ["forty-two"] = 42
};

var value = Assert.Contains("forty-two", dictionary);
Assert.Equal(42, value);
```

I think this is meaningful enough and has enough application that it could stand in the standard assertions library.

(I've done the work for this already, but CONTRIBUTING.md asks for a "help-wanted" ticket first.)

bradwilson:
Make it so! 😄

ahsteele:
I believe this this issue should be closed now.

bradwilson:
You are correct! :smile:

jbrantly:
As @ben-reilly pointed out in a separate issue, the current implementation doesn't actually work with the example code given in the original ticket. Looking at the tests, it looks like the example code was augmented with explicit casts. Is that really the desired behavior here? Would there be support for removing one of the overloads?

PR #1641, opened 2018-02-12
Adds Tests for Assert.Contains Overloads

chrisoverzero:
Positive and negative case for each.

Implementation exists in https://github.com/xunit/assert.xunit/pull/24.

[#1620]


bradwilson:
Thanks!

PR #1657, opened 2018-03-02
Use lazy enumeration of files in directory in: XunitFrontController.

palver123:
Faster for directories with many files.

Signed-off-by: Levente Koncz <Levente.Koncz@sapagroup.com>


bradwilson:
Thanks!

Issue #1164, opened 2017-03-15
 Times shown in Output window, is too precise


codito:
*Filing on behalf of @davkean. Copied from https://github.com/Microsoft/vstest/issues/614.*

The time that is shown in the Output window is too precise, to the degree that marks it harder to read:

![image](https://cloud.githubusercontent.com/assets/1103906/23927270/c50670ba-096c-11e7-848b-0816b304dc98.png)

I'm not sure why it needs to show this much precision, it's makes the output so much harder to parse in my head.

onovotny:
FWIW, this noise is there if `diagnosticMessages` is enabled. The 2.3-beta runner now suppresses the extra verbose stuff behind an `internalDiagnosticMessages` flag, which is off by default. This lets you get normal diagnostic messages without the noise. Can you try that?

bradwilson:
I don't believe that the complaint is related to those messages (which aren't diagnostic messages, BTW), but rather the seven digits of fractional seconds. :)

onovotny:
Agreed...but thought they looked like tons of discovery diagnostic messages, but I see those are diff test assemblies. My bad :)

bradwilson:
I'm okay with this being reduced to 2 digits after the decimal instead of 7.

Silic0nS0ldier:
Think this issue is ready to be closed. Referenced commits are all in master now.

PR #1660, opened 2018-03-08
adds format string for TimeSpan in the logger, addressing #1164

xirsoi:


xirsoi:
Fixed the broken unit test.

xirsoi:
The build failures are a result of the updated C# specification.

bradwilson:
Thanks!

PR #1685, opened 2018-04-10
xunit.vs.runner to support managed sources only

mayankbansal018:
Following RFC: https://github.com/Microsoft/vstest-docs/blob/master/RFCs/0020-Improving-Logic-To-Pass-Sources-To-Adapters.md

### DIFF
--- a/src/xunit.runner.visualstudio/VsTestRunner.cs
+++ b/src/xunit.runner.visualstudio/VsTestRunner.cs
@@ -1,5 +1,6 @@
 ﻿using System;
 using System.Collections.Generic;
+using System.ComponentModel;
 using System.Diagnostics;
 using System.IO;
 using System.Linq;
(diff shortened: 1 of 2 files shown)


src/xunit.runner.visualstudio/VsTestRunner.cs
```diff
@@ -23,6 +24,9 @@ namespace Xunit.Runner.VisualStudio
     [FileExtension(".exe")]
     [DefaultExecutorUri(Constants.ExecutorUri)]
     [ExtensionUri(Constants.ExecutorUri)]
+#if !WINDOWS_UAP
+    [Category("managed")]
```
onovotny: What does the UWP runner do instead?

mayankbansal018: UWP or NetCore runner don't really have an issue, since for both of them platfrom strictly picks adapter from source location only.
The main issue is with fullclr where we have various locations we can get adapter from, hence we end up giving source dll's to adapters, which we can avoid.

bradwilson:
@mayankbansal018 What qualifies this as WIP rather than ready to merge?

mayankbansal018:
@bradwilson I'm just running some end to end test with various adapters. Once I've completed those will remove the WIP tag.

mayankbansal018:
@bradwilson , @onovotny I've done local validations, can you please complete the PR.

Issue #565, opened 2015-08-31
Add diagnostic message sink option for fixtures


nbarbettini:
I have a test fixture that is shared via `ICollectionFixture<MyFixtureClass>`. Some of my unit tests inject `ITestOutputHelper` to output diagnostic information.

Is there any way to inject `ITestOutputHelper` into `MyFixtureClass`? I want to output log messages during construction and disposal of the fixture.

If I change the constructor:

``` chsarp
private readonly ITestOutputHelper output;

public MyFixtureClass(ITestOutputHelper output)
{
    this.output = output;
}
```

I get: 

> 'MyFixtureClass' must be a non-abstract type with a public parameterless constructor in order to use it as a parameter 'TFixture' in the generic type or method 'ICollectionFixture<TFixture>'

bradwilson:
`ITestOutputHelper` is for tests. It's not for infrastructure or fixtures (both of which have longer lifetime than tests).

If you want output from your fixtures, and it's during a test, the test can pass its own injected `ITestOutputHelper` into the fixture. Alternatively, we could add a feature that lets you inject `IMessageSink` to send diagnostic messages to; that's available for infrastructure classes today, but not for fixtures.

nbarbettini:
Got it, makes sense. I forgot to clarify what exactly I wanted to log; I updated my message. I specifically want to log stuff that's happening outside of the scope of a currently running test. `IMessageSink` would work perfectly if it's available for collection fixtures in some future release.

bradwilson:
I've updated the title to reflect the feature request. :)

nbarbettini:
Thanks @bradwilson!

mikesigs:
+1

salixzs:
Is there any plans for this? I was able to partially cover this by injecting ```ITestOutputHelper``` to Fixture, but it "goes away" if I want to use it in Dispose method with exception "There is no currently active test case".

gitfool:
👍

adamrodger:
+1

MarcoRossignoli:
+1

sinclairtarget:
+1!

eriove:
I have the same need as @salixzs (i.e. getting diagnostic output in Dispose methods). Did you ever find a workaround?

marcin-chwedczuk-cybervadis:
[...]

PR #1705, opened 2018-05-14
Allow to inject IMessageSink into ClassFixture/CollectionFixture (#565)

michalkowal:
Added possibility to use DiagnosticMessageSink in ClassFixture.

I also changed a little bit current implementation of CollectionFixture creation to meet more or less ClassFixtures mechanism


bradwilson:
Thanks! Lots of people will love this. :)

MarcoRossignoli:
@michalkowal great!TYVM!

gitfool:
@michalkowal @bradwilson this works great except for one thing I noticed which is that fixture logs via message sink seem to be at the warning level. Is this by design?

Issue #1715, opened 2018-05-24
dotnet xunit fails for integration tests on ASP.NET Core 2.1-rc1


cod7alex:
Repro:

1. Download .NET Core 2.1 RC1 SDK.
2. Create solution with two projects: `dotnet new webapi`, `dotnet new xunit`.
3. Add reference from test project to api project
4. Add simple integration test
```
            var server = new TestServer(new WebHostBuilder()
                .UseStartup<Startup>());
            var client = server.CreateClient();

            var response = await client.GetAsync("/api/values");

            Assert.Equal(HttpStatusCode.OK, response.StatusCode);
```
5. `dotnet xunit` from test project folder results in `System.IO.FileNotFoundException : Could not load file or assembly 'Microsoft.AspNetCore.Hosting, Version=2.1.0.0, Culture=neutral, PublicKeyToken=adb9793829ddae60'. The system cannot find the file specified.`
6. Apply modifications from https://stackoverflow.com/questions/50401152/integration-and-unit-tests-no-longer-work-on-asp-net-core-2-1-failing-to-find-as

Results:

`dotnet xunit` with `2.3.1` runner results in `System.IO.FileNotFoundException: Could not load file or assembly 'Microsoft.AspNetCore.Mvc.Core, Version=2.1.0.0, Culture=neutral, PublicKeyToken=adb9793829ddae60'. The system cannot find the file specified.`

`dotnet xunit` with `2.4.0-beta.2.build4003` runner results in 
```It was not possible to find any compatible framework version
The specified framework 'Microsoft.AspNetCore.App', version '2.1.0-rc1' was not found.
  - Check application dependencies and target a framework version installed at:
      C:\Program Files\dotnet\
  - Installing .NET Core prerequisites might help resolve this problem:
      http://go.microsoft.com/fwlink/?LinkID=798306&clcid=0x409
  - The .NET Core framework and SDK can be installed from:
      https://aka.ms/dotnet-download
  - The following versions are installed:
      2.1.0-rc1-final at [C:\Program Files\dotnet\shared\Microsoft.AspNetCore.App].
```

cc @bradwilson

bradwilson:
2.3.1 is known to be broken.

We don't make any requests of frameworks, so I'm left wondering whether the second behavior you described is actually anything related to us or not. /cc @natemcmaster

bradwilson:
[...]

PR #1716, opened 2018-05-25
Only use --fx-version if user specified it

natemcmaster:
Fixes https://github.com/xunit/xunit/issues/1715

--fx-version is only required when overriding what is in runtimeconfig.json. This change detects the runtimeconfig.json file path, and changes `dotnet exec` to only use `--fx-version` if the user has specified it on dotnet-xunit.

### DIFF
--- a/src/dotnet-xunit/Program.cs
+++ b/src/dotnet-xunit/Program.cs
@@ -349,9 +349,12 @@ class Program
                 if (runtimeFrameworkVersion == "2.0")
                     runtimeFrameworkVersion = "2.0.0";
 
-                var fxVersion = FxVersion ?? runtimeFrameworkVersion;
-                WriteLine($"Running .NET Core {fxVersion} tests for framework {targetFramework}...");
-                return RunDotNetCoreProject(outputPath, assemblyName, targetFileName, extraArgs, fxVersion, $"netcoreapp{version.Major}.0");
+                if (runtimeFrameworkVersion == "2.1")
+                    runtimeFrameworkVersion = "2.1.0";
+
+                var displayFxVersion = FxVersion ?? runtimeFrameworkVersion;
+                WriteLine($"Running .NET Core {displayFxVersion} tests for framework {targetFramework}...");
+                return RunDotNetCoreProject(outputPath, assemblyName, targetFileName, extraArgs, FxVersion, $"netcoreapp{version.Major}.0");
             }
             if (targetFrameworkIdentifier == ".NETFramework" && version >= Version452)
             {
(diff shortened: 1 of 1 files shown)


bradwilson:
Why?

I've been fairly clear that I always want `dotnet xunit` to run with _**exactly**_ the version the user is compiled against, so my initial reaction is to turn this down.

src/dotnet-xunit/Program.cs
```diff
@@ -349,9 +349,12 @@ int RunTargetFramework(string testProject, string targetFramework, bool amendOut
                 if (runtimeFrameworkVersion == "2.0")
                     runtimeFrameworkVersion = "2.0.0";
 
-                var fxVersion = FxVersion ?? runtimeFrameworkVersion;
-                WriteLine($"Running .NET Core {fxVersion} tests for framework {targetFramework}...");
-                return RunDotNetCoreProject(outputPath, assemblyName, targetFileName, extraArgs, fxVersion, $"netcoreapp{version.Major}.0");
+                if (runtimeFrameworkVersion == "2.1")
```
bradwilson: PLEASE tell me the team didn't fuck this up AGAIN. :angry:

natemcmaster:
Example scenarios:

Test app that uses Microsoft.**AspNetCore**.App 2.1.999
 
  - $(RuntimeFrameworkVersion) = 2.1.0
  - Microsoft.NETCore.App version 2.1.0
  - runtimeconfig.json
    ```json
    {
      "runtimeOptions": {
        "tfm": "netcoreapp2.1",
        "framework": {
          "name": "Microsoft.AspNetCore.App",
          "version": "2.1.999"
        }
      }
    }
    ```

By running `dotnet exec --fx-version 2.1.0`, you're force downgrading the framework version to 2.1.0.

src/dotnet-xunit/Program.cs
natemcmaster: I don't know if they did or not. Was just being defensive.

natemcmaster:
By the way, I get your point about wanting to ensure tests run on exactly the sharedfx users declare. That is still available to them by running `dotnet xunit -fxversion`. 

Or, they can hand-craft their runtimeconfig.json to disable patch rollforward.
```json
{
    "runtimeOptions": {
        "framework": {
            "name": "Microsoft.NETCore.App",
            "version": "2.1.0"
        },
        "applyPatches": false
    }
}
```
I've never seen this done in the wild, but the knob exists for users who 100% absolutely for sure want to ensure they don't get lifted to patches.

bradwilson:
I've decided to deprecate and stop shipping `dotnet xunit`. Thanks.

cod7alex:
Not sure how to interpret the last comment from PR.

mikesigs:
@bradwilson Other than your comment above about not shipping `dotnet xunit` anymore. Is this mentioned anywhere else? I just spent the last ish hours fighting various fights with it, only to finally see here that I shouldn't be using it at all. I suppose the recommendation then is to use `dotnet test`? Some form of deprecation notice would be helfpul when running the tool perhaps.

Thanks!

natemcmaster:
@mikesigs if you're on Twitter, the xunit account is pretty good about these kinds of change. https://twitter.com/xunit/status/1001321969189437440. Also, it's in the release notes: https://xunit.github.io/releases/2.4-beta2.  Unfortunately, NuGet.org doesn't provide a way to mark a package as deprecated.

bradwilson:
Yeah, the big clue should've been when you tried to upgrade it to 2.4.0 and it said "what 2.4.0?" :grin:

PR #1734, opened 2018-06-07
Add Test Case Exclusions

### LINKED ISSUE #1701: Proposal: Method / Class / Namespace exclusions as XUnitFilters
<h2>Problem</h2>
Currently XUnit doesn't support a way of dynamically skipping methods, classes or namespaces. Disabling tests requires adding annotations to source code, which is often not feasible or convenient - e.g. when multiple projects need to run the same test suite. Considering that assemblies can be explicitly excluded from running, exclusion of methods seems like a fairly straightforward addition. 

<h2>Proposal</h2>
Mirror the implementation for included methods, namespaces and classes in `XunitFilters.cs` to allow for exclusions of the above. 

https://github.com/xunit/xunit/blob/6bd342df4ea50477ada6e128f366cb8756b3016d/src/xunit.runner.utility/Project/XunitFilters.cs#L35-L55

This would allow for Xunit to be run with a (for example) `-nomethod`  or `-noclass` option from the command line. 

<h2>Comments</h2>
The behavior of Included/excluded traits seems to specify that inclusion supersedes exclusion, so this will be mimicked here as well.
The only mention of this I could find was in #791 , which seems to have been closed for age. 

Has this feature been considered before? 
Seems like the implementation would be straight-forward and if the change is considered appropriate to the project, I would submit a PR with the changes.

A-And:
Fixes #1701. 

Mirror the implementation for including and excluding methods, classes and namespaces.

cc @ViktorHofer

### DIFF
--- a/src/dotnet-xunit/Program.cs
+++ b/src/dotnet-xunit/Program.cs
@@ -229,10 +229,18 @@ class Program
         Console.WriteLine("  -method \"name\"         : run a given test method (can be fully specified or use a wildcard;");
         Console.WriteLine("                         : i.e., 'MyNamespace.MyClass.MyTestMethod' or '*.MyTestMethod')");
         Console.WriteLine("                         : if specified more than once, acts as an OR operation");
+        Console.WriteLine("  -nomethod \"name\"       : do not run a given test method (can be fully specified or use a wildcard;");
+        Console.WriteLine("                         : i.e., 'MyNamespace.MyClass.MyTestMethod' or '*.MyTestMethod')");
+        Console.WriteLine("                         : if specified more than once, acts as an AND operation");
         Console.WriteLine("  -class \"name\"          : run all methods in a given test class (should be fully");
         Console.WriteLine("                         : specified; i.e., 'MyNamespace.MyClass')");
         Console.WriteLine("                         : if specified more than once, acts as an OR operation");
+        Console.WriteLine("  -noclass \"name\"        : do not run any methods in a given test class (should be fully");
+        Console.WriteLine("                         : specified; i.e., 'MyNamespace.MyClass')");
+        Console.WriteLine("                         : if specified more than once, acts as an AND operation");
         Console.WriteLine("  -namespace \"name\"      : run all methods in a given namespace (i.e.,");
+        Console.WriteLine("                         : 'MyNamespace.MySubNamespace')");
+        Console.WriteLine("  -nonamespace \"name\"    : do not run any methods in a given namespace (i.e.,");
         Console.WriteLine("                         : 'MyNamespace.MySubNamespace')");
         Console.WriteLine("                         : if specified more than once, acts as an OR operation");
         Console.WriteLine("  -noautoreporters       : do not allow reporters to be auto-enabled by environment");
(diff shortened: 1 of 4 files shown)


src/dotnet-xunit/Program.cs
```diff
@@ -229,11 +229,19 @@ static void PrintUsage()
...
+        Console.WriteLine("                         : if specified more than once, acts as an AND operation");
         Console.WriteLine("  -class \"name\"          : run all methods in a given test class (should be fully");
         Console.WriteLine("                         : specified; i.e., 'MyNamespace.MyClass')");
         Console.WriteLine("                         : if specified more than once, acts as an OR operation");
+        Console.WriteLine("  -noclass \"name\"        : do not run any methods in a given test class (should be fully");
+        Console.WriteLine("                         : specified; i.e., 'MyNamespace.MyClass')");
+        Console.WriteLine("                         : if specified more than once, acts as an AND operation");
         Console.WriteLine("  -namespace \"name\"      : run all methods in a given namespace (i.e.,");
         Console.WriteLine("                         : 'MyNamespace.MySubNamespace')");
+        Console.WriteLine("  -nonamespace \"name\"    : do not run any methods in a given namespace (i.e.,");
```
ViktorHofer: L245 `Console.WriteLine("                         : if specified more than once, acts as an OR operation");` belongs to the command above "-namespace". If the OR also applies to "-nonamespace" you should probably copy and paste that line above.

src/xunit.console/ConsoleRunner.cs
```diff
@@ -266,11 +266,19 @@ void PrintUsage(IReadOnlyList<IRunnerReporter> reporters)
...
             Console.WriteLine("                         : specified; i.e., 'MyNamespace.MyClass')");
             Console.WriteLine("                         : if specified more than once, acts as an OR operation");
+            Console.WriteLine("  -noclass \"name\"        : do not run any methods in a given test class (should be fully");
+            Console.WriteLine("                         : specified; i.e., 'MyNamespace.MyClass')");
+            Console.WriteLine("                         : if specified more than once, acts as an AND operation");
             Console.WriteLine("  -namespace \"name\"      : run all methods in a given namespace (i.e.,");
             Console.WriteLine("                         : 'MyNamespace.MySubNamespace')");
+            Console.WriteLine("  -nonamespace \"name\"    : do not run any methods in a given namespace (i.e.,");
+            Console.WriteLine("                         : 'MyNamespace.MySubNamespace')");
             Console.WriteLine("                         : if specified more than once, acts as an OR operation");
```
ViktorHofer: same here.

src/xunit.runner.utility/Project/XunitFilters.cs
```diff
@@ -40,15 +47,30 @@ public XunitFilters()
...
+
         /// <summary>
         /// Gets the set of method filters for tests to include.
         /// </summary>
         public ICollection<string> IncludedMethods => includedMethods;
 
+        /// <summary>
+        /// Gets the set of assembly filters for tests to exclude.
+        /// </summary>
+        public HashSet<string> ExcludedNameSpaces { get; }
```
ViktorHofer: nit: "ExcludedNamespaces", according to https://docs.microsoft.com/en-us/dotnet/csharp/programming-guide/namespaces/using-namespaces

A-And: Here I've stuck to the convention above, i.e. `IncludedNameSpaces` .

src/dotnet-xunit/Program.cs
A-And: Good catch, thanks!

src/xunit.runner.utility/Project/XunitFilters.cs
bradwilson: I'll change it after merge.

bradwilson:
Thanks!

PR #1741, opened 2018-06-13
Use HTTPS for package sources

AArnott:
This fixes the error the msbuild /restore gives (at least with VS 15.8).

### DIFF
--- a/NuGet.Config
+++ b/NuGet.Config
@@ -1,7 +1,7 @@
 ﻿<?xml version="1.0" encoding="utf-8"?>
 <configuration>
   <packageSources>
-    <add key="myget.org/F/xunit" value="http://www.myget.org/F/xunit/api/v3/index.json" protocolVersion="3" />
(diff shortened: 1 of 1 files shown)


NuGet.Config
```diff
@@ -1,6 +1,8 @@
 ﻿<?xml version="1.0" encoding="utf-8"?>
 <configuration>
   <packageSources>
+    <clear />
```
onovotny: Why do we need to avoid machine configured sources? That could break caching improvements on build servers or other local overrides couldn't it?

AArnott:
> Why do we need to avoid machine configured sources?

No, this doesn't disable local fallback package sources. But it does ensure the package restore works every time regardless of machine configuration, and it never fails due to lack of security tokens when the machine configuration includes sources that require authentication.

Don't you remember [this conversation](https://twitter.com/aarnott/status/992608379972419584?s=19)?

AArnott:
Also, I think nuget restore is as slow as your slowest package source. So adding more package sources that are faster doesn't actually speed up package restore time. Something about the nuget algorithm requires asking every package source for every package I believe.

onovotny:
Right, forgot about that.

PR #1750, opened 2018-06-25
First dispose async and then usual dispose #1749

### LINKED ISSUE #1749: XunitTestCollectionRunner - IAsyncLifetime's DisposeAsync and IDisposable's Dispose run in parallel
**Summary:**

When a test collection implements both IAsyncLifetime and IDisposable, corresponding Dispose methods are executed in parallel despite the documentation:
```C#
     /// <summary>
    /// Called when an object is no longer needed. Called just before <see cref="M:System.IDisposable.Dispose" />
    /// if the class also implements that.
    /// </summary>
    Task DisposeAsync();
```

**Source:**
https://github.com/xunit/xunit/blob/89d11eca58bed6754304a2b3897bf2a0f6dcd838/src/xunit.execution/Sdk/Frameworks/Runners/XunitTestCollectionRunner.cs#L56

**Workaround**
Just use one or another interface

maxcherednik:
Fixes #1749

### DIFF
--- a/test/test.xunit.execution/Sdk/Frameworks/Runners/XunitTestCollectionRunnerTests.cs
+++ b/test/test.xunit.execution/Sdk/Frameworks/Runners/XunitTestCollectionRunnerTests.cs
@@ -67,9 +67,9 @@ public class XunitTestCollectionRunnerTests
     {
         public bool Disposed;
 
-        public bool DisposeAsyncCalled { get; private set; }
+        public bool DisposeAsyncCalled;
 
-        public TaskCompletionSource<bool> DisposeAsyncSignaler { get; } = new TaskCompletionSource<bool>();
+        public TaskCompletionSource<bool> DisposeAsyncSignaler = new TaskCompletionSource<bool>();
 
         public void Dispose()
         {
--- a/test/test.xunit.execution/Sdk/Frameworks/Runners/XunitTestCollectionRunnerTests.cs
+++ b/test/test.xunit.execution/Sdk/Frameworks/Runners/XunitTestCollectionRunnerTests.cs
@@ -39,6 +39,58 @@ public class XunitTestCollectionRunnerTests
     }
 
     [Fact]
+    public static async void DisposeAndAsyncLifetimeShouldBeCalledInTheRightOrder()
+    {
+        var collection = new TestCollection(Mocks.TestAssembly(), Reflector.Wrap(typeof(CollectionForFixtureAsyncLifetimeAndDisposableUnderTest)), null);
+        var testCase = Mocks.XunitTestCase<XunitTestCollectionRunnerTests>("DisposeAndAsyncLifetimeShouldBeCalledInTheRightOrder", collection);
+        var runner = TestableXunitTestCollectionRunner.Create(testCase);
+
+        var runnerSessionTask = runner.RunAsync();
+
+        await Task.Delay(500);
+
+        var fixtureUnderTest = runner.CollectionFixtureMappings.Values.OfType<FixtureAsyncLifetimeAndDisposableUnderTest>().Single();
+
+        Assert.True(fixtureUnderTest.DisposeAsyncCalled);
+        Assert.False(fixtureUnderTest.Disposed);
+
+        fixtureUnderTest.DisposeAsyncSignaler.SetResult(true);
+
+        await runnerSessionTask;
+
+        Assert.True(fixtureUnderTest.Disposed);
+    }
+
+    class CollectionForFixtureAsyncLifetimeAndDisposableUnderTest : ICollectionFixture<FixtureAsyncLifetimeAndDisposableUnderTest> { }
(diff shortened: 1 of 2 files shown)


maxcherednik:
@bradwilson Do we want a unit test showing the order of the calls?

bradwilson:
I'm not sure that that would be sufficient for a test, because they are currently called in the right order.

I think the right test would be something like: implement an object with both, and in the async cleanup method you (a) delay ~ 500ms, and then (b) set a completion flag; in the non-async cleanup method, you validate that the flag has been set. It's not perfect, but it should be sufficient.

maxcherednik:
@bradwilson That's what meant.

test/test.xunit.execution/Sdk/Frameworks/Runners/XunitTestCollectionRunnerTests.cs
```diff
@@ -38,6 +38,58 @@ public class XunitTestCollectionRunnerTests
         Assert.True(fixtureUnderTest.Disposed);
     }
 
+    [Fact]
+    public static async void DisposeAndAsyncLifetimeShouldBeCalledInTheRightOrder()
```
maxcherednik: `async void` - not super Idiomatic, but just following the rest of the code

test/test.xunit.execution/Sdk/Frameworks/Runners/XunitTestCollectionRunnerTests.cs
```diff
@@ -38,6 +38,58 @@ public class XunitTestCollectionRunnerTests
...
+    {
+        var collection = new TestCollection(Mocks.TestAssembly(), Reflector.Wrap(typeof(CollectionForFixtureAsyncLifetimeAndDisposableUnderTest)), null);
+        var testCase = Mocks.XunitTestCase<XunitTestCollectionRunnerTests>("DisposeAndAsyncLifetimeShouldBeCalledInTheRightOrder", collection);
+        var runner = TestableXunitTestCollectionRunner.Create(testCase);
+
+        var runnerSessionTask = runner.RunAsync();
+
+        await Task.Delay(500);
```
maxcherednik: Can't wait it any other way here, since I don't control the lifecycle of the fixture.

test/test.xunit.execution/Sdk/Frameworks/Runners/XunitTestCollectionRunnerTests.cs
```diff
@@ -38,6 +38,58 @@ public class XunitTestCollectionRunnerTests
...
+
+        var fixtureUnderTest = runner.CollectionFixtureMappings.Values.OfType<FixtureAsyncLifetimeAndDisposableUnderTest>().Single();
+
+        Assert.True(fixtureUnderTest.DisposeAsyncCalled);
+        Assert.False(fixtureUnderTest.Disposed);
+
+        fixtureUnderTest.DisposeAsyncSignaler.SetResult(true);
```
maxcherednik: Releasing the disposeAsync call.

test/test.xunit.execution/Sdk/Frameworks/Runners/XunitTestCollectionRunnerTests.cs
bradwilson: It's idiomatic for xUnit. :wink:

test/test.xunit.execution/Sdk/Frameworks/Runners/XunitTestCollectionRunnerTests.cs
```diff
@@ -38,6 +38,58 @@ public class XunitTestCollectionRunnerTests
...
+        Assert.True(fixtureUnderTest.Disposed);
+    }
+
+    class CollectionForFixtureAsyncLifetimeAndDisposableUnderTest : ICollectionFixture<FixtureAsyncLifetimeAndDisposableUnderTest> { }
+
+    class FixtureAsyncLifetimeAndDisposableUnderTest : IAsyncLifetime, IDisposable
+    {
+        public bool Disposed;
+
+        public bool DisposeAsyncCalled { get; private set; }
```
bradwilson: Why one a property but other a field? (fields are fine here IMO)

maxcherednik: Just a habit. Sorry, did not follow the code. Fixing it now.

test/test.xunit.execution/Sdk/Frameworks/Runners/XunitTestCollectionRunnerTests.cs
bradwilson: :+1:

bradwilson:
Just the one nitpick. :)

bradwilson:
Thanks!

PR #1773, opened 2018-07-25
First dispose async and then usual dispose #1772

### LINKED ISSUE #1772: XunitTestClassRunner - IAsyncLifetime's DisposeAsync and IDisposable's Dispose run in parallel
**Summary:**

When a test class fixture implements both IAsyncLifetime and IDisposable, corresponding Dispose methods are executed in parallel despite the documentation:

```csharp
     /// <summary>
    /// Called when an object is no longer needed. Called just before <see cref="M:System.IDisposable.Dispose" />
    /// if the class also implements that.
    /// </summary>
    Task DisposeAsync();
```

**Source:**
https://github.com/xunit/xunit/blob/1fb6b672768bed98f5e7b6656c1bbb3792ed0530/src/xunit.execution/Sdk/Frameworks/Runners/XunitTestClassRunner.cs#L156-L164

**Workaround**
Just use one or another interface

---

This is the same bug as https://github.com/xunit/xunit/issues/1749 but for `XunitTestClassRunner`

campersau:
test case copied from https://github.com/xunit/xunit/pull/1750

Fixes #1772


bradwilson:
This should be targeted against the `v2` branch.

Issue #1683, opened 2018-04-09
Could not find file 'C:\...\bin\Debug\netcoreapp2.0\content.store.shared.tests.xunit.runner.json'.


ronenfe:
When debugging an xunit test using .net core 2.0.0 and visual studio, I get the above exception when exception settings is set to show all exceptions. 
I think you should not throw an exception when this file does not exist as it's not obligatory to use a json setting file.
This excpetion is thrown multiple times for each test class in the project and for each project:
Stack:

> System.Private.CoreLib.dll!System.IO.FileStream.OpenHandle(System.IO.FileMode mode, System.IO.FileShare share, System.IO.FileOptions options) Line 55	C#
 	System.Private.CoreLib.dll!System.IO.FileStream.FileStream(string path, System.IO.FileMode mode, System.IO.FileAccess access, System.IO.FileShare share, int bufferSize, System.IO.FileOptions options) Line 192	C#
 	System.Private.CoreLib.dll!System.IO.FileStream.FileStream(string path, System.IO.FileMode mode, System.IO.FileAccess access, System.IO.FileShare share) Line 132	C#
 	System.IO.FileSystem.dll!System.IO.File.OpenRead(string path) Line 340	C#
 	xunit.runner.utility.netcoreapp10.dll!Xunit.ConfigReader_Json.File_OpenRead(string path) Line 159	C#
 	xunit.runner.utility.netcoreapp10.dll!Xunit.ConfigReader_Json.LoadFile(string configFileName) Line 128	C#
 	xunit.runner.utility.netcoreapp10.dll!Xunit.ConfigReader_Json.Load(string assemblyFileName, string configFileName) Line 120	C#
 	xunit.runner.utility.netcoreapp10.dll!Xunit.ConfigReader.Load(string assemblyFileName, string configFileName) Line 18	C#
 	xunit.runner.visualstudio.dotnetcore.testadapter.dll!Xunit.Runner.VisualStudio.VsTestRunner.LoadConfiguration(string assemblyName) Line 378	C#
 	xunit.runner.visualstudio.dotnetcore.testadapter.dll!Xunit.Runner.VisualStudio.VsTestRunner.Microsoft.VisualStudio.TestPlatform.ObjectModel.Adapter.ITestExecutor.RunTests.AnonymousMethod__7_3(System.Linq.IGrouping<string, Microsoft.VisualStudio.TestPlatform.ObjectModel.TestCase> group) Line 180	C#
[...]

PR #1787, opened 2018-08-17
check if config exists before load

YevhenLukomskyi:
Resolves https://github.com/xunit/xunit/issues/1683

### DIFF
--- a/src/xunit.runner.utility/Configuration/ConfigReader_Json.cs
+++ b/src/xunit.runner.utility/Configuration/ConfigReader_Json.cs
@@ -138,6 +138,12 @@ namespace Xunit
         {
             try
             {
+#if !NETSTANDARD1_1
+                if(!File.Exists(configFileName))
+                {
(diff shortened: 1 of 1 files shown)


bradwilson:
@onovotny Is the File.Exists check appropriate for devices?

onovotny:
Need to verify, but I *think* it's okay for content files. It's assemblies that may get munged (AOT, .NET Native, security) and have loader restrictions.

bradwilson:
This should be targeted against the `v2` branch.

KirillOsenkov:
Will v2 be merged to master at some point? Is v3 going to start from master? How do we get the last few fixes from v2 back into master?

bradwilson:
It's undetermined at this time how much v2 code will make it into v3, as I haven't even really started the design. Config files are surely going to be part of v3, though, so that code will likely come in. I'll probably just merge v2 into master before doing the v3 work.

Issue #1788, opened 2018-08-18
Some generic theories pass on Windows while failing on Linux 


safakgur:
Windows 10 Pro 17134.112

The project file:

```xml
<Project Sdk="Microsoft.NET.Sdk">

  <PropertyGroup>
    <TargetFrameworks>netcoreapp2.0</TargetFrameworks>
    <IsPackable>false</IsPackable>
  </PropertyGroup>

  <ItemGroup>
    <PackageReference Include="Microsoft.NET.Test.Sdk" Version="15.8.0" />
    <PackageReference Include="xunit" Version="2.4.0" />
    <PackageReference Include="xunit.runner.visualstudio" Version="2.4.0" />
    <PackageReference Include="coverlet.msbuild" Version="2.2.1" />
  </ItemGroup>

</Project>
```

The test:

```c#
using Xunit;

namespace Tests
{
    public sealed class Tests
    {
        [Theory]
        [InlineData(null)]
        [InlineData(1)]
        [InlineData("S")]
        public void GenericTest<T>(T value)
        {
            var temp = value;
        }
    }
}
```

`dotnet test` (SDK 2.1.400-win-x64) succeeds:

```
Test run for ...\bin\Debug\netcoreapp2.0\Tests.dll(.NETCoreApp,Version=v2.0)
Microsoft (R) Test Execution Command Line Tool Version 15.7.0
Copyright (c) Microsoft Corporation.  All rights reserved.

Starting test execution, please wait...

Total tests: 3. Passed: 3. Failed: 0. Skipped: 0.
Test Run Successful.
Test execution time: 1,2411 Seconds
```

Then I start `wsl` on the same device.

* `lsb_release -a` gives me "Ubuntu 16.04.4"
* `dotnet test` (SDK 2.1.400-linux-x64) fails:

```
Test run for /mnt/c/.../bin/Debug/netcoreapp2.0/Tests.dll(.NETCoreApp,Version=v2.0)
Microsoft (R) Test Execution Command Line Tool Version 15.8.0
Copyright (c) Microsoft Corporation.  All rights reserved.

Starting test execution, please wait...
[xUnit.net 00:00:01.41]     Tests.Tests.GenericTest [FAIL]
Failed   Tests.Tests.GenericTest
Error Message:
 System.NotSupportedException : Specified method is not supported.
Stack Trace:
   at System.Reflection.RuntimeMethodInfo.ThrowNoInvokeException()
[...]

Issue #1781, opened 2018-08-09
Generic method broken with xunit 2.4


ViktorHofer:
This sample which worked in xunit 2.3.1 doesn't work anymore in xunit 2.4 and results in a "Specified method is not supported" exception:

```
[Theory, MemberData(nameof(ReverseData))]
public void RunOnce<T>(IEnumerable<T> source, T dummy)
{
    T[] expected = source.ToArray();
    Array.Reverse(expected);

    IEnumerable<T> actual = source.Reverse();

    Assert.Equal(expected, actual);
}

public static IEnumerable<object[]> ReverseData()
{
    var integers = new[]
    {
        Array.Empty<int>(), // No elements.
        new[] { 1 }, // One element.
        new[] { 9999, 0, 888, -1, 66, -777, 1, 2, -12345 }, // Distinct elements.
        new[] { -10, 0, 5, 0, 9, 100, 9 }, // Some repeating elements.
    };
    
    // TODO: Remove workarounds when xUnit is updated to include xunit/xunit#965.
    return integers
        .Select(collection => new object[] { collection, 0 })
        .Concat(
            integers.Select(c => new object[] { c.Select(i => i.ToString()), string.Empty })
        );
}
```

Exception:
```
Error Message:
 System.NotSupportedException : Specified method is not supported.
Stack Trace:
   at System.Reflection.RuntimeMethodInfo.ThrowNoInvokeException()
   at System.Reflection.RuntimeMethodInfo.InvokeArgumentsCheck(Object obj, BindingFlags invokeAttr, Binder binder, Object[] parameters, CultureInfo culture)
   at System.Reflection.RuntimeMethodInfo.Invoke(Object obj, BindingFlags invokeAttr, Binder binder, Object[] parameters, CultureInfo culture)
```

Is that a known bug? This currently blocks us in corefx adopting the xunit 2.4 runner. It would ideal if we can fix that issue so that we can either take a dependency on a nightly build or a service release.

ViktorHofer:
cc @bradwilson @onovotny

ViktorHofer:
Any hints how to debug that issue? I looked through the commits if something related shouts out but didn't find anything.

bradwilson:
Looks like a new bug. I don't know of any workaround, other than staying on 2.3.1.

bradwilson:
Sorry, no. I've done no investigation into the problem, only verified that it was broken.

ViktorHofer:
Ok then I'll start digging.

[...]

PR #1790, opened 2018-08-20
Exclude byreflike types from type conversion

ViktorHofer:
Fixes https://github.com/xunit/xunit/issues/1781
Fixes https://github.com/xunit/xunit/issues/1788

With netcoreapp2.1 we added implicit conversions between string and `ReadOnlySpan<char>` which are picked here and tried to be invoked. The execution fails as ByRefLike Types aren't currently supported to be invoked by `MethodInfo.Invoke`. The error currently doesn't happen on .NET Framework as we haven't yet added these implicit conversions there.

This approach uses reflection as the `Type.ByRefLike` property isn't exposed in .NET Standard. It will probably be included with .NET Standard vNext.

cc @jkotas @bradwilson @onovotny

### DIFF
--- a/src/xunit.execution/Sdk/TypeUtility.cs
+++ b/src/xunit.execution/Sdk/TypeUtility.cs
@@ -158,16 +158,41 @@ namespace Xunit.Sdk
 
             // Check if we can implicitly convert the argument type to the parameter type
             var implicitMethod = conversionDeclaringType.GetRuntimeMethod("op_Implicit", methodTypes);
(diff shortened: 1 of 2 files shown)


bradwilson:
Looks good to me. This should be targeted against the `v2` branch, though.

bradwilson:
Closed in favor of #1791.

ViktorHofer:
Opened https://github.com/xunit/xunit/pull/1791. Do we also merge this into master?

PR #1791, opened 2018-08-20
Exclude byreflike types from type conversion

ViktorHofer:
Fixes https://github.com/xunit/xunit/issues/1781
Fixes https://github.com/xunit/xunit/issues/1788

With netcoreapp2.1 we added implicit conversions between string and `ReadOnlySpan<char>` which are picked here and tried to be invoked. The execution fails as ByRefLike Types aren't currently supported to be invoked by `MethodInfo.Invoke`. The error currently doesn't happen on .NET Framework as we haven't yet added these implicit conversions there.

This approach uses reflection as the `Type.ByRefLike` property isn't exposed in .NET Standard. It will probably be included with .NET Standard vNext.

cc @jkotas @bradwilson @onovotny

### DIFF
--- a/src/xunit.execution/Sdk/TypeUtility.cs
+++ b/src/xunit.execution/Sdk/TypeUtility.cs
@@ -158,16 +158,41 @@ namespace Xunit.Sdk
 
             // Check if we can implicitly convert the argument type to the parameter type
             var implicitMethod = conversionDeclaringType.GetRuntimeMethod("op_Implicit", methodTypes);
+#if FEATURE_REFLECTION
+            if (implicitMethod != null && implicitMethod.IsStatic && !IsByRefLikeType(implicitMethod.ReturnType))
+#else
             if (implicitMethod != null && implicitMethod.IsStatic)
+#endif
                 return implicitMethod.Invoke(null, methodArguments);
 
             // Check if we can explicitly convert the argument type to the parameter type
             var explicitMethod = conversionDeclaringType.GetRuntimeMethod("op_Explicit", methodTypes);
+#if FEATURE_REFLECTION
+            if (explicitMethod != null && explicitMethod.IsStatic && !IsByRefLikeType(explicitMethod.ReturnType))
+#else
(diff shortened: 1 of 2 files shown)


bradwilson:
Thanks!

ViktorHofer:
You are welcome, please let me know when a package is available on the official feed to be consumed.

bradwilson:
I don't have any CI set up for the `v2` branch yet, so there is no MyGet version yet. It's on my TODO list (I was hoping I wouldn't ship another 2.x build, but the broken abstractions 2.0.2 package is reason enough to ship a 2.4.1).

ViktorHofer:
Fair enough. Of course you could also just manually update the execution and the abstractions packages if you believe there won't be any further bugfixes.

bradwilson:
There are 3 open PRs, so I may as well include them if I'm going to the trouble to build new packages.

ViktorHofer:
I retargeted those open PRs against v2 to speed things up… We are kind of dependent on that fix in corefx.

ViktorHofer:
I'm not 100% sure but there's a chance that I introduced a regression with this change:

```
System.InvalidCastException: Unable to cast object of type 'Microsoft.DotNet.XUnitExtensions.SkipOnTargetFrameworkDiscoverer' to type 'Xunit.Sdk.ITraitDiscoverer'.
```

Let me verify that and push another PR if necessary.

bradwilson:
FYI, `2.4.1-alpha.1.build4054` is on MyGet now, with all 3 of the PRs merged.

ViktorHofer:
Okay that failure was an intermediate one which went away after clean rebuilding. Change is safe!

ViktorHofer:
@bradwilson do you have a timeline when 2.4.1 will make it to the nuget feed? should we stay on the myget feed for now or is it worth waiting a day or two?

bradwilson:
Unlikely there will be anything before the weekend.

ViktorHofer:
Thanks for letting me know.

Issue #1681, opened 2018-03-28
Make it possible to deliver custom reporters via nuget packages


nvborisenko:
Currently reporter should be placed near `xunit.console.exe`. It would be really nice if reporter can be placed near test assembly. For end users it will mean that user can easily install nuget package into his project and execute tests via `xunit.console.exe`. So using 1 instance of runner user will be able to set reporters per test assembly.

bradwilson:
Tagging as "Discussion" because I'm not sure what the best mechanism is here. We don't have access to any package information when tests are run.

nvborisenko:
That's true. Currently xunit.console.exe loads reporters from working directory. Is it possible to include the directory where test assembly is located and try to load custom reporter from there?

nvborisenko:
It would be nice that `visualstudio.adapter` finds reporters in the folder where test assembly located.

nvborisenko:
@bradwilson it seems working solution. User installs custom reporter via nuget, and explicitly says to xunit.exe (console or adapter) to load this reporter via assembly attribute. There is no performance degradation due xunit.exe will not scan all assemblies in test directory.

bradwilson:
Closing this due to the redesign of reporters in v3. They can now be shipped as a NuGet package, and then registered by the test project in question. More information: https://xunit.net/docs/getting-started/v3/custom-runner-reporter

PR #1800, opened 2018-08-26
Add reporter flag

ViktorHofer:
Adds a reporter flag to xunit.console to only load a specified reporting assembly. This avoids loading all assemblies from the current working directory into memory which can cause a) huge memory consumption depending on the number of assemblies in the same directory and b) FileLoadExceptions if the assembly wasn't meant to be loaded by the runner.

cc @bradwilson

### DIFF
--- a/src/xunit.console/CommandLine.cs
+++ b/src/xunit.console/CommandLine.cs
@@ -55,7 +55,9 @@ namespace Xunit.ConsoleClient
 
         public bool Wait { get; protected set; }
 
-        public IRunnerReporter ChooseReporter(IReadOnlyList<IRunnerReporter> reporters)
+        public string Reporter { get; protected set; }
+
+        public IRunnerReporter ChooseReporter(IEnumerable<IRunnerReporter> reporters)
         {
             var result = default(IRunnerReporter);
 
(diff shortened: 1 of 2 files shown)


bradwilson:
I have two issues:

1. I tend to want there to be parity between the console and MSBuild runners.
2. The MSBuild runner already uses `Runner` as its attribute name to select the runner, so this switch conflicts philosophically with the existing MSBuild attribute.

I'm trying to think of a better switch name. I'm open to suggestions.

ViktorHofer:
@bradwilson do you think it makes sense to follow-up on this? I can easily make the necessary changes also in the msbuild runner but what about the option name?

bradwilson:
I think at this point it's unlikely to make it into 2.x as I'm really trying to limit the changes for 2.4.1, so we can reevaluate for 3.0.

nvborisenko:
Subscribing. It seems these changes conflict with https://github.com/xunit/xunit/issues/1681
Then I will try to create PR to resolve above issue.

ViktorHofer:
It would be great if we could at least have an option to avoid loading all the assemblies in the cwd but still add features that you mentioned in the your issue. Maybe something like `-reportersFilter .;C:\fancyxunitreporters` which would use both the current directory and a given path to search for reporter assemblies.

nvborisenko:
Yeah, it might work to me. xUnit team, your input is highly appreciated.

LonghronShen:
Any update to this?
Now, `dotnet test -xml` option still not work, how can we export a junit style xml to the ci system?

ViktorHofer:
You could give https://github.com/spekt/xunit.testlogger a try. It worked for me.

ViktorHofer:
Closing.

Issue #1794, opened 2018-08-21
Generic test method with type argument in Func throws


ViktorHofer:
Repro:

```csharp
[Theory]
[MemberData(nameof(CreateRangeSliceWithSelectorData))]
public void CreateRangeSliceWithSelector<TResult>(IEnumerable<int> source, int start, int length, Func<int, TResult> selector)
{
    var expected = source.Skip(start).Take(length).Select(selector);
    Assert.Equal(expected, ImmutableArray.CreateRange(source.ToImmutableArray(), start, length, selector));
}

public static IEnumerable<object[]> CreateRangeSliceWithSelectorData()
{
    yield return new object[] { new[] { 4, 5, 6, 7 }, 0, 0, new Func<int, float>(i => i + 0.5f) };
    yield return new object[] { new[] { 4, 5, 6, 7 }, 0, 0, new Func<int, double>(i => i + 0.5d) };
    yield return new object[] { new[] { 4, 5, 6, 7 }, 0, 0, new Func<int, int>(i => i) };
    yield return new object[] { new[] { 4, 5, 6, 7 }, 0, 1, new Func<int, int>(i => i * 2) };
    yield return new object[] { new[] { 4, 5, 6, 7 }, 0, 2, new Func<int, int>(i => i + 1) };
    yield return new object[] { new[] { 4, 5, 6, 7 }, 0, 4, new Func<int, int>(i => i) };
    yield return new object[] { new[] { 4, 5, 6, 7 }, 3, 1, new Func<int, int>(i => i) };
    yield return new object[] { new[] { 4, 5, 6, 7 }, 3, 0, new Func<int, int>(i => i) };
    yield return new object[] { new[] { 4, 5, 6, 7 }, 4, 0, new Func<int, int>(i => i) };
}
```

Exception:

```
System.Collections.Immutable.Tests.ImmutableArrayTest.CreateRangeSliceWithSelector [FAIL]
        System.IndexOutOfRangeException : Index was outside the bounds of the array.
```

This isn't really blocking us in corefx as we just disable these tests for now (only 4 tests). If I have time I will look into it.

ViktorHofer:
cc @hughbe

bradwilson:
Is this related to #1781? Or something different?

ViktorHofer:
It's a different one. I believe it's related to https://github.com/xunit/xunit/pull/965. Probably an oversight.

bradwilson:
Fixed by #1825.

PR #1812, opened 2018-09-10
Fix resolving a generic func parameter in index 1

hughbe:
Fixes https://github.com/xunit/xunit/issues/1794

/cc @ViktorHofer


bradwilson:
If you want this in 2.4.1, can you please re-target this against the `v2` branch?

bradwilson:
This was updated and merged as part of #1825. Thanks!

PR #1825, opened 2018-10-05
Fix generic oversight v2

ViktorHofer:
Port from @hughbe's change: https://github.com/xunit/xunit/pull/1812 to v2 + feedback.


bradwilson:
Thanks both of you!

PR #1831, opened 2018-10-14
Tests updated for the changes made for (#1806)

### LINKED ISSUE #1806: Include full value in Assert.Empty output
Example:
```csharp
[Fact]
public void Empty() {
    Assert.Empty(new[] { 1, 2, 3 });
}
```

produces

```
Message: Assert.Empty() Failure
```

which is not very useful.

Would be good to get something like this instead:
```
Message: Assert.Empty() Failure
Expected: <empty>
Actual:   { 1, 2, 3 }
```

arms11:
Fixes #1806.
Updated unit tests to meet the change made in sub-module.


bradwilson:
Retargeted to the `v2` branch and merged. Thanks!

PR #1837, opened 2018-10-20
Changed assert messages for Single

### LINKED ISSUE #1351: Add more to the output of Assert.Single
When I call `Assert.Single(list, "expectedValue");` I get the following on a failure:
```
Message: The collection contained 0 matching element(s) instead of 1.
```
It would be helpful to know which `expectedValue` wasn't found, especially if a test contains multiple `Assert.Single` assertions. 

Perhaps:

```
Message: Expected 1 of "expectedValue" but collection contains X
```

Or, based on the number of matches, customize the message:

```
Collection is missing "expectedValue"
Collection should only have 1 of "expectedValue" but found 2
```

Perhaps even print out the collection like `Assert.Contains` does, if the collection isn't too big?

Also, would it be possible to use an expression for the predicate and walk the expression to print it out too? That might be a bit much.

davidwengier:
Test changes for https://github.com/xunit/xunit/issues/1351 which are in this PR: https://github.com/xunit/assert.xunit/pull/28

Fixes #1351

### DIFF
--- a/test/test.xunit.assert/Asserts/CollectionAssertsTests.cs
+++ b/test/test.xunit.assert/Asserts/CollectionAssertsTests.cs
@@ -883,7 +883,7 @@ public class CollectionAssertsTests
             Exception ex = Record.Exception(() => Assert.Single(collection));
 
             Assert.IsType<SingleException>(ex);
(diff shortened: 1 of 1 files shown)


bradwilson:
Retargeted to the `v2` branch and merged. Thanks!

Issue #1842, opened 2018-10-26
Linux arm/arm64 intermittent failures


echesakovMSFT:
We recently switched to using **xunit.console.dll** for running tests in https://github.com/dotnet/coreclr on Linux arm and Linux arm64 and we immediately started seeing the following failures during the test runs https://github.com/dotnet/coreclr/issues/20392
```
$CORE_ROOT/corerun $CORE_ROOT/xunit.console.dll [List of test assemblies]

System.IO.FileLoadException: Could not load file or assembly 'Exceptions.Finalization.XUnitWrapper, Version=0.0.0.0, Culture=neutral, PublicKeyToken=null'. An operation is not legal in the current state. (Exception from HRESULT: 0x80131509 (COR_E_INVALIDOPERATION))
System.IO.FileLoadException: Could not load file or assembly 'Exceptions.ForeignThread.XUnitWrapper, Version=0.0.0.0, Culture=neutral, PublicKeyToken=null'. An operation is not legal in the current state. (Exception from HRESULT: 0x80131509 (COR_E_INVALIDOPERATION))
```

These are hard to reproduce on a local machine issues and sometimes it would take 20-30 runs to actually see the failure. But they are failing pretty consistently in our CI jobs.

I tried to investigate the failures to see if this is our runtime issue or an Xunit issue (you can see https://github.com/dotnet/coreclr/issues/20594 for more information). What I suspect causes this exception (System.IO.FileLoadException) is another exception (System.InvalidOperationException) thrown when multiple threads call LoadManagedDll and modify managedAssemblyCache https://github.com/xunit/xunit/blob/9d10262a3694bb099ddd783d735aba2a7aac759d/src/common/AssemblyResolution/DependencyContextAssemblyCache.cs#L156-L165

My question if this is a known issue? Would you be interested to investigate? 

I also was trying to add a simple instrumentation code to Xunit to see it this is a case. But as soon as I compile my own version of Xunit and use this to run our tests I can't reproduce the issue anymore.

I also see differences in IL code for my compiled version of xunit.console.dll (ad20ed120a3653d6af533b76cf8a04f096095349) and for a version published on NuGet (https://www.nuget.org/packages/xunit/2.4.1-pre.build.4059 ). This is an example of such code difference 
```
17753c18019
<       IL_002a:  ldstr      "File not found: {0}"
---
>       IL_002a:  ldstr      "File not found: "
17755,17756c18021,18022
[...]

PR #1846, opened 2018-10-29
Use ConcurrentDictionary in DependencyContextAssemblyCache

echesakovMSFT:
`AssemblyHelper.OnResolving` https://github.com/xunit/xunit/blob/c54cc52ffb275c81afed022521870193bbca6c39/src/common/AssemblyResolution/AssemblyHelper_NetCoreApp.cs#L84

can be called from multiple threads. This can cause hard to reproduce race conditions in `DependencyContextAssemblyCache.LoadManagedDll` modifying underlying `managedAssemblyCache`
https://github.com/xunit/xunit/blob/c54cc52ffb275c81afed022521870193bbca6c39/src/common/AssemblyResolution/DependencyContextAssemblyCache.cs#L156
 For some reason it happens more often on Linux/arm Linux/arm64 (see https://github.com/dotnet/coreclr/issues/20594 and https://github.com/xunit/xunit/issues/1842) but also we also saw this happening on OSX/x64. This problem manifests themselves as `InvalidOperationException` *"Operations that change non-concurrent collections must have exclusive access. A concurrent update was performed on this collection and corrupted its state. The collection's state is no longer correct."*

This PR switches to using `ConcurrentDictionary` instead of `Dictionary` for `managedAssemblyCache` and `unmanagedAssemblyMap` in `DependencyContextAssemblyCache` to mitigate issue https://github.com/xunit/xunit/issues/1842.

### DIFF
--- a/src/common/AssemblyResolution/DependencyContextAssemblyCache.cs
+++ b/src/common/AssemblyResolution/DependencyContextAssemblyCache.cs
@@ -1,6 +1,7 @@
 ﻿#if NETCOREAPP
 
 using System;
+using System.Collections.Concurrent;
 using System.Collections.Generic;
 using System.IO;
 using System.Linq;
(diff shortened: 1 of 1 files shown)


bradwilson:
Considering that this is used by code which is no longer supported, and will not be used in 3.0, it's unlikely that we will merge this PR.

echesakovMSFT:
@bradwilson Can we merge this as a servicing fix for v2?

bradwilson:
We have no plans for any more 2.x releases. I'm happy to leave it open in the event that we do, but the likelihood is almost zero.

bradwilson:
Also, this is opened against the wrong branch. It should be opened against `v2`.

echesakovMSFT:
@bradwilson Updated this to be against v2

BruceForstall:
@bradwilson We continue to depend on this component, and expect to for some time. We can't use "dotnet test" because we don't have a version of the dotnet CLI on all the platforms we need to run on. If the given code is no longer supported, is there any downside to taking a fix, to support teams that depend on it?

bradwilson:
There's no planned vehicle to ship it. You're better off building it yourself rather than hoping some other showstopper bug shows up that might cause me to accept this PR.

mattzink:
I think having no delivery mechanism for fixes until 3.0 comes out one day is a bad way to go

PR #1863, opened 2018-12-05
Refactor VsTestRunner to support multiple reporters

nvborisenko:
Currently VsTestAdapter uses only single reporter to report results. It is not clear which exactly. So this PR tries to use all available reporters.

If we use `dotnet vstest folderA\Tests.dll`, VsTestAdapter tries to use all available reporters in `folderA`. 

It provides a way to put custom reporters near test assembly, so custom reporters can be installed via NuGet.


bradwilson:

bradwilson:
This code has been moved to [xunit/visualstudio.xunit](https://github.com/xunit/visualstudio.xunit), but it's also not clear to me that this performs a function that is required or desirable. I think you should consider opening a [discussion](https://github.com/xunit/xunit/discussions) related to the feature you're after here (with specific examples) so that @clairernovotny can determine the best way to support your request.

PR #1944, opened 2019-05-12
Fix #1924: Enumerable equal assert

### LINKED ISSUE #1924: Assert.Equal for long arrays cuts of the actual difference
E.g.

```cs
Assert.Equal(
    new string[] { "1", "2", "3", "4", "5", "6", "7", "8", "9", "10", "11", "12", "13", "14", "15", "16" },
    new string[] { "1", "2", "3", "4", "5", "6", "7", "8", "9", "10", "11", "12", "13", "14", "15", "RUBBISH" }
)
``` 

We get an assertion failure, but it only starts from the start of the array and not where the failure occurs. I propose something like for strings:
```cs
Assert.Equal(
    "This is a very long string and I'm interested in it's full value, before and after.",
    "This is a very long string and |I'm interested in it's full value, before and after."
);
```
```
Message: Assert.Equal() Failure
                                 ↓ (pos 31)
Expected: ···ery long string and I'm interested in it's full value, before···
Actual:   ···ery long string and |I'm interested in it's full value, befor···
                                 ↑ (pos 31)
```

daveMueller:
Closes #1924

### DIFF
--- a/ISSUE_TEMPLATE.md
+++ b/ISSUE_TEMPLATE.md
@@ -2,8 +2,11 @@ Before you file a bug...
 
 * Is this a question? Ask it on our Slack server: https://xunit.slack.com/
   Not a member? It's easy to sign up: https://xunit-slackin.herokuapp.com/
-* Is this a bug related to Mono on Linux/OS X? Sorry, we don't officially support Mono.
-* Did you read the documentation @ https://xunit.github.io/ ?
+* Is this a bug with VSTest (aka Visual Studio Test Explorer, `dotnet test`,
+  `dotnet vstest`, or `vstest.console.exe`)? Open those issues here:
+  https://github.com/xunit/visualstudio.xunit/issues/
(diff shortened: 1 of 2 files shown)


daveMueller:
Don't know why the build fails. I would be grateful if someone could help me out.

bradwilson:
Because this change depends on the other change being merged first, it will naturally fail until that other merge takes place. It's nothing to be concerned about, as I usually hand-merge them both on my box when doing verification.

bradwilson:

bradwilson:
Thanks! This was recently merged with https://github.com/xunit/assert.xunit/commit/bb5d19df7322627f649e55e8006157cd0af31ac6 and https://github.com/xunit/xunit/commit/f4be18e8bf926b3ec0ead036fca5df7fb2114116 and as such, this PR is now redundant, so I'm closing it.

PR #1956, opened 2019-06-04
remove some redundant type checking

SimonCropp:
minor perf improvement, and less code


bradwilson:
Thanks! I'm surprised there weren't more places to fix than this. :wink:

PR #1958, opened 2019-06-10
Fixes issue where traits cannot be inherited from multiple levels of inheritance.

### LINKED ISSUE #1397: Inherited traits overriden by class
When assigning a trait to a parent class, it is overriden by the childs class trait instead of added to.

```
[Trait("Category", "Integration")]
public class TraitExample { }

[Trait("Category", "Test")]
public class InheritedTraitExample : TraitExample
{
    [Fact]
    public void TestExample()
    {
        Assert.True(true);
    }
}
```
However this makes TestExample marked as integration:

```
[Trait("Category", "Integration")]
public class TraitExample { }

public class InheritedTraitExample : TraitExample
{
    [Fact]
    [Trait("Category", "Test")]
    public void TestExample()
    {
        Assert.True(true);
    }
}
```

 Adding traits to the test itself doesn't override, but adding any traits to the class does.
 not sure if this is a runner (visual studio) problem or xunit itself.

christophano:
Fixes #1397 

The issue was caused due to `ITraitAttribute` using the default `AttributeUsageAttribute` (as it can't have one of its own, not being an actual `Attribute`), which means the `AllowMultiples` property is false.

This change adds a default `AttributeUsageAttribute` instance straight into the cache when it is initialised. `ConcurrentDictionary` does not work with collection initialisers, so I've added an `IEnumerable<KeyValuePair<Type, AttributeUsageAttribute>>` into the constructor. This is far from neat, so perhaps it would be better to add a static constructor to the `ReflectionAttributeInfo` class and seed the cache at that point?

Also, I have added unit tests to prove the fix, but these could do with some improvement too - I'm open to suggestions. The actual method that ought to be tested is `ReflectionAttributeInfo.GetCustomAttributes(Type, string)`, but it is internal.

### DIFF
--- a/src/xunit.v3.core.tests/Sdk/Frameworks/XunitTestCaseTests.cs
+++ b/src/xunit.v3.core.tests/Sdk/Frameworks/XunitTestCaseTests.cs
@@ -136,6 +136,134 @@ public class XunitTestCaseTests
 		{
 			public BugAttribute(int id) { }
 		}
+
+		public static TheoryData<Type, IEnumerable<string>> CustomAttributeTestCases() =>
+			new TheoryData<Type, IEnumerable<string>>
+			{
+				{ typeof(ClassWithSingleTrait), new[] { "One" } },
+				{ typeof(ClassWithMultipleTraits), new[] { "One", "Two" } },
+				{ typeof(InheritedClassWithOnlyOwnTrait), new[] { "One" } },
+				{ typeof(InheritedClassWithOnlyOwnMultipleTraits), new[] { "One", "Two" } },
+				{ typeof(InheritedClassWithSingleBaseClassTrait), new[] { "BaseOne" } },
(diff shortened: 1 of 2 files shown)


dougbu:
Does the logic at https://github.com/xunit/xunit/blob/cfd23d71c54d440e645e21451162a225fe55f14b/src/xunit.v3.execution/Sdk/Frameworks/XunitTestCase.cs#L151-L153 also need to change❔ That seems to put assembly-level `ITraitAttribute` implementations ahead of class-level ones.

bradwilson:
@dougbu The order doesn't matter. You can have multiple values for the same key, and it doesn't matter where they come from; the list is merged. Another way to say it is this: you cannot "remove" or "override" a trait that comes from another source; you're only ever adding additional key/value pairs.

bradwilson:

bradwilson:
Thanks!

Issue #1980, opened 2019-07-08
xunit json parser fails to parse unicode escaped strings


tmds:
I'm looking into an issue where xunit test won't run with a source-build .NET Core: https://github.com/microsoft/vstest/issues/2079.

The root cause is in xunit json parsing for unicode escaped strings:

https://github.com/xunit/xunit/blob/2c6c63b903b6d15b6eab056f04fd713329fae138/src/common/Json.cs#L306-L325

This code throws `IndexOutOfRangeException` at `_codePointBuffer[i] = (char)next;` because the `StringBuilder` has a `Length` of `0`.

tmds:
> a source-build .NET Core: microsoft/vstest#2079

Some more info about the relation with the .NET Core 3.0 sdk:

The new `System.Text.Json` classes are escaping `+` (https://github.com/dotnet/corefx/issues/38354) which cause the escape sequence to show up in the `deps.json` `sha512` properties.

tmds:
@bradwilson do you have an idea when this will be part of an xunit release? We don't want xunit to stop working with .NET Core 3.0.

ericstj:
Just looked at this with @nguerrera, would be nice to get a fix for this.  We're working on more [fine-grained escaping](https://github.com/dotnet/corefx/issues/37192) but that isn't in yet.

nguerrera:
I think we need to fix https://github.com/dotnet/core-setup/issues/7137 regardless. I don't think we want to force every project to take a new xunit reference to use .net core 3.0.

nguerrera:
By the way, is it vstest or this repo that ends up catching this bug and simply not discovering any tests?

PR #1981, opened 2019-07-08
Fix IndexOutOfRange when parsing json string with unicode escaped characters

tmds:
Fixes https://github.com/xunit/xunit/issues/1980


bradwilson:
Thanks!

PR #2015, opened 2019-09-25
Use cast instead of as for IReflectionTypeInfo

SimonCropp:
since reflectionTestMethodType is always used. I would assume "non IReflectionTypeInfo" is not allowed since it would result in null ref. A cast is better since if it is incorrect an invalid cast will be easier to debug than a null ref


bradwilson:
Yeah, this points out an interesting hole in the definition, as the related accessors should really be redefined here. Something for me to ponder, if this survives into v3.

PR #2018, opened 2019-09-26
Add -inclusionfile command line option

KeithLRobertson:
Allows specifying command line options in a file to ex/include tests, one per line.
Accepted options are -[no]trait, -[no]method, -[no]class, and -[no]namespace.
Accepts blank lines and comments starting with # and //.

### DIFF
--- a/src/xunit.console/CommandLine.cs
+++ b/src/xunit.console/CommandLine.cs
@@ -308,26 +308,20 @@ namespace Xunit.ConsoleClient
                     if (option.Value == null)
                         throw new ArgumentException("missing argument for -trait");
 
-                    var pieces = option.Value.Split('=');
-                    if (pieces.Length != 2 || string.IsNullOrEmpty(pieces[0]) || string.IsNullOrEmpty(pieces[1]))
+                    if (!Trait.TryParse(option.Value, out Trait trait))
                         throw new ArgumentException("incorrect argument format for -trait (should be \"name=value\")");
 
-                    var name = pieces[0];
-                    var value = pieces[1];
-                    project.Filters.IncludedTraits.Add(name, value);
+                    project.Filters.IncludedTraits.Add(trait.Name, trait.Value);
                 }
                 else if (optionName == "notrait")
                 {
                     if (option.Value == null)
                         throw new ArgumentException("missing argument for -notrait");
 
-                    var pieces = option.Value.Split('=');
-                    if (pieces.Length != 2 || string.IsNullOrEmpty(pieces[0]) || string.IsNullOrEmpty(pieces[1]))
+                    if (!Trait.TryParse(option.Value, out Trait notrait))
                         throw new ArgumentException("incorrect argument format for -notrait (should be \"name=value\")");
 
-                    var name = pieces[0];
-                    var value = pieces[1];
-                    project.Filters.ExcludedTraits.Add(name, value);
+                    project.Filters.ExcludedTraits.Add(notrait.Name, notrait.Value);
                 }
                 else if (optionName == "class")
                 {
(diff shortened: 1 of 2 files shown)


KeithLRobertson:
The error above looks like a technical failure. I don't know how to resolve it. Can someone on the project please help?

bradwilson:

src/xunit.console/CommandLine.cs
```diff
@@ -395,6 +396,71 @@ protected XunitProject Parse(Predicate<string> fileExists)
             return project;
         }
 
+        static void IncludeFromFile(XunitProject project, String path)
+        {
+            using (var input = File.OpenText(path))
+            {
+                string line;
+                int linenum = 0;
+                var spaceOrTab = " \t".ToCharArray();
```
meirkr, line 405: I'm less familiar with the files modeling in this project,
But isn't there any place for holding such static vars which will never change?
If not, at least move such thing to a static member var in this class, to be rused, rhather than generating such conversions on each iteration.
What do you think?
(I know it is minor, but maybe such accumulated improvements might bring better performance.
and asssume developper's environment is a Raspberry-Pi...)

src/xunit.console/CommandLine.cs
```diff
@@ -395,6 +396,71 @@ protected XunitProject Parse(Predicate<string> fileExists)
...
+                while ((line = input.ReadLine()?.Trim()) != null)
```
meirkr, line 406: If starting to parsing files,
What do you say about using ReadLineAssync
Meaning changing the upstream flow to Async model as well, starting from the root (Program.Main)
?

src/xunit.console/CommandLine.cs
```diff
@@ -395,6 +396,71 @@ protected XunitProject Parse(Predicate<string> fileExists)
             return project;
         }
 
+        static void IncludeFromFile(XunitProject project, String path)
+        {
+            using (var input = File.OpenText(path))
```
meirkr, line 401: Consider splitting the mothos to have one method which accepts an argument of some input stream.
Then you can unit test that method using a mocked stream.
Then you chan check your varioius string parsers logic.
What do you think about that?

src/xunit.console/CommandLine.cs
KeithLRobertson: I considered that, but this runs during command line processing at startup; we would expect it to run once at most. Thus, it would actually be more expensive to compute this whether needed or not and save it in a static forever when no longer needed.

src/xunit.console/CommandLine.cs
KeithLRobertson: Changing the application from sync to async may be valuable, but I would consider it way out of scope for the intention of this change.

src/xunit.console/CommandLine.cs
KeithLRobertson: It's not a bad idea at all. But considering this has languished for a year with no activity and was closed with no explanation, I don't believe it is worth my time. Apparently either this project is dead, or the people who could merge this don't consider this addition valuable, so I'm not going to waste my efforts further.

meirkr (commented):
I've understood that @bradwilson is pushing towards the next architecture of xunit V3.
V2 will not have additional builds anymore.
So if this PR is so old I would guess it is not relevant to V3.
If this is correct, I would propose to cancel this PR.
(I just thought it is part of V3 work, but was I guess I was wrong).

bradwilson:
I've left this PR (and #2032, which is similar) open because I do want to support a response file. I'm not sure which strategy I would want to use yet as I'm not ready to integrate it. I may want to pull from one or both of the PRs for code & ideas.

bradwilson:
As part of the effort to clean up older PRs, this is being closed due to age. If this is still applicable with v3, please open a new issue to discuss the feature so it can be re-evaluated and categorized.

Thanks!

PR #2019, opened 2019-09-26
Emit MethodName as a test-case property in -nunit output

KeithLRobertson:
CI post-processing scripts may need the test method name; many tools use nunit-style output.
When the test is a Theory, the method name has to be parsed from test-case name. Much worse,
when xunit is used with TechTalk.SpecFlow, the test-case name is based on the Scenario name;
the method name is completely lost! Emit method name as property MethodName in nunit output.


KeithLRobertson:
The error above looks like a technical failure. I don't know how to resolve it. Can someone on the project please help?

bradwilson:

bradwilson:
Thanks!

PR #2026, opened 2019-10-07
Reintroduce license.txt

### LINKED ISSUE #2022: Nuget package license file link does not work
With change 3a40411e1fa3066f84065b35d2f7865b199666a2 (2019/09/28) the license file link in the nuget package does not work anymore.
Can be reproduced by opening https://www.nuget.org/packages/xunit/ and clicking on "License info".
This leads to a failure in any automated license check process.

josteink:
This fixes the license-link on Nuget.org for older versions of xUnit. 

This closes #2022.

### DIFF
--- a/license.txt
+++ b/license.txt
@@ -1,1 +1,16 @@
-LICENSE
+Unless otherwise noted, the source code here is covered by the following license:
+
+    Copyright (c) .NET Foundation and Contributors
+    All Rights Reserved
+
+    Licensed under the Apache License, Version 2.0 (the "License");
+    you may not use this file except in compliance with the License.
+    You may obtain a copy of the License at
+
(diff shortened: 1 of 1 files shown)


vDiBo:
Any news on this issue?

bradwilson:
This file has moved so it can be in the standard place GitHub expects it to be. We won't be accepting this PR.

josteink:
Ok. But then you should publish a new version soon, so that the license link on Nuget.org actually works 🙂

vDiBo:
> Ok. But then you should publish a new version soon, so that the license link on Nuget.org actually works 🙂

And the new Nuget package should not reference the master branch but a specific commit. So that already released packages are not broken by subsequent changes. :-)

Issue #2027, opened 2019-10-08
Xunit doesn't have an Assert.Fail() operation


josteink:
In many unit-test frameworks there is an explicit Assert-operation for triggering a test-failure in a context where something is *wrong* (for whatever reason) but there's nothing concrete to assert on.

This typically is implemented as `Assert.Fail(message)`. The best you can do in xUnit right now to emulate that is using `Assert.True(false, "Message")`, but this can cause confusion and noise in the test-output:

````
message

Expected: True
Actual:   False
````

In the case of an explicit failure the only wanted output should be:

````
message
````

It would be nice if xUnit added an `Assert.Fail(string message)` operation which didn't include unnecessary assert-information,

**Comment by bradwilson** [comment]
Historically, we have not supported `Assert.Fail` for the same reason our asserts don't (usually) have error messages: because we believe that a custom assertion is a better option than general purpose failure. Our workaround for this is `Assert.True(false, message)`.

**Comment by josteink** [comment]
But it’s not a good workaround. And it makes migrating from other frameworks much harder than it needs to be. 

And all the reasons provided against ‘Assert.Fail()’  *clearly* apply even more so to ‘Assert.True(false)’. 

It’s not a good workaround, and it’s a even worse migration story.

I’d love to have you reconsider this issue. The wort is going to be there either way, so let’s choose the least ugly one.

**Comment by bradwilson** [comment]
We provide the assertion library as both a Git submodule as a source-level NuGet package, in the event that you'd like to extend it for your own personal use (f.e., to add Fail).

**Comment by josteink** [comment]
> We provide the assertion library as both a Git submodule as a source-level NuGet package, in the event that you'd like to extend it for your own personal use (f.e., to add Fail).

That's not really ergonomical though, and again, another pain users will encounter/figure out/solve when trying to migrate from other test-frameworks, something which no doubt will reduce Xunit-adoption.

As a general open-source proponent, I value trying to make it easier for new contributors to join in on "my" projects, and I value projects which feels welcoming for me to join. I see that as a necessity for a project to stay healthy, and not go stale once the "old" contributors decide to retire.

I like to think a community-project like Xunit should be such a community *and* reflect the wishes of the community. But if you unilaterally *insist* this is something you don't want and you wont add, I guess I'll just close all my issues and PRs and limit my involvement in the future.

Just be aware by making such decisions, you've also decided to turn away contributors which could help grow Xunit in the future.

Your call.

**Comment by SGmuwa** [comment]
```
[...]

PR #2028, opened 2019-10-08
Add tests for Assert.Fail().

josteink:
This PR depends on https://github.com/xunit/assert.xunit/pull/34 being merged before compiling and running cleanly.

When both are merged, this closes https://github.com/xunit/xunit/issues/2027.

### DIFF
--- a/test/test.xunit.assert/Asserts/FailAssertsTests.cs
+++ b/test/test.xunit.assert/Asserts/FailAssertsTests.cs
@@ -1,0 +1,34 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+using Xunit;
+using Xunit.Sdk;
+
+namespace test.xunit.assert.Asserts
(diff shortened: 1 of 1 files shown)


test/test.xunit.assert/Asserts/FailAssertsTests.cs
```diff
@@ -0,0 +1,34 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+using Xunit;
+using Xunit.Sdk;
+
+namespace test.xunit.assert.Asserts
```
Kerl1310: Looking at the other tests, this isn't required

josteink: Fair enough. Namespace removed.

josteink:
@bradwilson: same as other thread

josteink:
If there’s no interest for these PRs, I’ll just close them.

PR #2029, opened 2019-10-08
Add output for skip message in HTML transform

### LINKED ISSUE #2004: Skip messages are not output to the html result file
In a current project (targeting .net framework 4.7.2), we rely on the HTML test result file to show the test result after running "xunit.console.exe myTests.dll -html myTestResults.html".

We have a few xUnit tests containing Skip message dynamically created at runtime.  We found that even though the skip messages were output to the console, the HTML file simply showed the names of skipped tests without the messages.

Because the skip messages typically contain useful information for users, would it be a quick fix to output the skip messages to the HTML file?

Thanks,

mchughmk:
Fixes #2004 

Adds a new row containing the reason a test was skipped in the HTML output file.

### DIFF
--- a/src/xunit.console/HTML.xslt
+++ b/src/xunit.console/HTML.xslt
@@ -30,6 +30,7 @@
           .indent { margin: 0.25em 0 0.5em 2em; }
           .clickable { cursor: pointer; }
           .testcount { font-size: 85%; }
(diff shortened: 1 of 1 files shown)


bradwilson:

PR #2044, opened 2019-10-30
Enable C# 8 Nullable Reference Types

### LINKED ISSUE #2011: Add C# 8.0 Nullable Attributes to Assert methods
C# 8.0 introduces Nullable annotations.

Sometimes an API is too complex to be annotated and requires custom attributes.

In particular, the following methods in Xunit.Assert could benefit from attributes:

```csharp
public class Assert
{
                public static void False( [DoesNotReturnIf(true)] bool condition);

		public static void False( [DoesNotReturnIf(true)] bool condition, string? userMessage);

                public static void True( [DoesNotReturnIf(false)] bool condition);

		public static void True( [DoesNotReturnIf(false)] bool condition, string? userMessage);

                public static void NotNull( [NotNull] object? @object); // see https://github.com/dotnet/roslyn/issues/37544
}
```

This would prevent me getting warnings in the following cases:
```csharp
Assert.True(dic.TryGetValue(key, out var value));
value.DoSomething();
```

And 

```csharp
var x = GetX();
Assert.NotNull(x);
x.DoSomething();
```

### LINKED ISSUE #2033: Feature Request: Support C# 8.0 Nullable Reference Types
I'm currently experimenting with C# 8.0 and Nullable Reference Types. One of the things that currently annoy me when writing unit tests with Nullable Reference Types is how the complier doesn't complain about code like this.

```
Exception exception = Record.Exception(() => SomeCode());
```

The complier doesn't complain about this because static method signature for Record.Exception() doesn't indicate that it can return a null. There may be other areas in the Xunit code base that do this as well and makes for a confusing experience when writing unit tests in C# 8.0 with Nullable Reference Types.

To be fair I can only come up with this example right now. I believe C# 8.0 and Nullable Reference Types is out of preview now.

drewnoakes:
Fixes #2011 and #2033.

There is a parallel PR that annotates `Assert.NotNull` in https://github.com/xunit/assert.xunit/pull/36.

The commit here enables NRT on two projects (`xunit.assert` and `xunit.core`), setting the language version to C# 8 on both. Ultimately it would probably make sense to enable it for all code in the repository, but it's prudent to roll it out in small steps such as this.

This requires Visual Studio 16.3 or later, or a new enough compiler. I expect this to PR to fail in various ways in CI, as this no doubt builds in environments where a new enough compiler is not available. In such cases it is possible to opt-in to a specific version of the compiler via its NuGet package. I'll see what happens in CI first.

I've been dogfooding NRT for a while now so am happy to field any questions about this PR or the feature in general. My goal is to improve the experience of using xUnit in tandem with NRT.

(I read the contribution guidelines and see that the associated issue is not flagged as help-wanted. This PR is meant to demonstrate what's required to get this to work, at least on a single Windows machine with VS2019 16.3 installed.)

### DIFF
--- a/src/common/Annotations.cs
+++ b/src/common/Annotations.cs
@@ -1,0 +1,11 @@
+// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+namespace System.Diagnostics.CodeAnalysis
+{
+    /// <summary>Applied to a method that will never return under any circumstance.</summary>
+    [AttributeUsage(AttributeTargets.Method, Inherited = false)]
+    internal sealed class DoesNotReturnAttribute : Attribute
+    { }
+}
--- a/LICENSE
+++ b/LICENSE
@@ -14,3 +14,30 @@ Unless otherwise noted, the source code here is covered by the following license
     WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
     See the License for the specific language governing permissions and
     limitations under the License.
+
+The file src/common/Annotations.cs is covered by the following license
+(copied from https://github.com/dotnet/runtime/blob/391dbfa87d8aba2a5bf338b433830de5f5c1cb3b/LICENSE.TXT):
+
+    The MIT License (MIT)
+
+    Copyright (c) .NET Foundation and Contributors
+
+    All rights reserved.
+
+    Permission is hereby granted, free of charge, to any person obtaining a copy
+    of this software and associated documentation files (the "Software"), to deal
+    in the Software without restriction, including without limitation the rights
+    to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+    copies of the Software, and to permit persons to whom the Software is
+    furnished to do so, subject to the following conditions:
+
+    The above copyright notice and this permission notice shall be included in all
+    copies or substantial portions of the Software.
+
+    THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+    IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+    FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+    AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+    LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+    OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+    SOFTWARE.
--- a/src/xunit.v3.assert/xunit.v3.assert.csproj
+++ b/src/xunit.v3.assert/xunit.v3.assert.csproj
@@ -1,6 +1,7 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
+    <DefineConstants>$(DefineConstants);XUNIT_NULLABLE</DefineConstants>
     <Description>Includes the assertion library from xUnit.net (xunit.v3.assert.dll). Supports .NET Standard 2.0.</Description>
     <GenerateDocumentationFile>true</GenerateDocumentationFile>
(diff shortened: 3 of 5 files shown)


src/xunit.assert/xunit.assert.csproj
```diff
@@ -3,6 +3,11 @@
   <PropertyGroup>
     <GenerateDocumentationFile>true</GenerateDocumentationFile>
     <TargetFramework>netstandard2.0</TargetFramework>
+    <LangVersion>8</LangVersion>
```
canton7: To be pedantic, `8` is not in the [list of supported langVersions](https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/compiler-options/langversion-compiler-option).

drewnoakes: Thanks. It certainly works. I've asked internally whether this is a deficiency in the docs or something that should be changed in the project.

drewnoakes: The conclusion is that either is fine. https://github.com/dotnet/roslyn/blob/master/src/Compilers/CSharp/Portable/LanguageVersion.cs#L340

src/common/Annotations.cs
```diff
@@ -0,0 +1,86 @@
+﻿// Copyright (c) Microsoft.  All Rights Reserved.  Licensed under the Apache License, Version 2.0.  See License.txt in the project root for license information.
+
+namespace System.Diagnostics.CodeAnalysis
+{
+    /// <summary>Specifies that null is allowed as an input even if the corresponding type disallows it.</summary>
+    [AttributeUsage(AttributeTargets.Field | AttributeTargets.Parameter | AttributeTargets.Property, Inherited = false)]
+    internal sealed class AllowNullAttribute : Attribute
```
pengweiqhca: Can use [Nullable](https://www.nuget.org/packages/Nullable/) package

drewnoakes: Could do, but I don't think we should force a new package dependency for a single file.

Orygeunik:
I'm sorry, but anyone have plan or roadmap to merge this RP and others RPs? Maybe in next release (which planned date?) this RP will be merged?

And thanks for project xunit :)

Nihlus:
I'd suggest using https://github.com/manuelroemer/Nullable instead of redefining the attributes directly in xunit - it'll reduce the footprint of enabling NRTs while achieving the same effect.

drewnoakes:
@Nihlus see https://github.com/xunit/xunit/pull/2044#discussion_r349866939

src/common/Annotations.cs
```diff
@@ -0,0 +1,86 @@
+﻿// Copyright (c) Microsoft.  All Rights Reserved.  Licensed under the Apache License, Version 2.0.  See License.txt in the project root for license information.
```
NextTurn:
```suggestion
// Licensed to the .NET Foundation under one or more agreements.
// The .NET Foundation licenses this file to you under the MIT license.
// See the LICENSE file in the project root for more information.
```

drewnoakes: Technically this file was copied from another project (corefx, now dotnet/runtime) and this was the original copyright header. They can be thought of as part of the framework rather than the library, and so it makes sense that they share the same license as the framework IMO.

NextTurn: > Technically this file was copied from another project (corefx, now dotnet/runtime) and this was the original copyright header.

Impossible. corefx always uses the MIT banner. Even Roslyn which is licensed under Apache uses the MIT banner for this file.

drewnoakes: My mistake. It was copied from dotnet/project-system which is Apache 2.0. I'll update the license there too.

src/common/Annotations.cs
idg10: You know it's a development dependency right? If you take a dependency on Nullable, this will not cause anyone using your component to also depend on Nullable. So if that's the cause for concern, it's a non-issue. (If on the other hand, you want to minimize even  your build-time dependencies, then ignore me.)

drewnoakes:
I looked into rebasing this but wasn't sure whether the `xunit.v3.*` folders were an intermediary step or the anticipated end state.

bradwilson:
Everything is moving to `xunit.v3.*` folder names (to match their new assembly names).

drewnoakes:
@bradwilson where should I rebase this? `dev`, `master` or `WIP`?

bradwilson:
WIP is going away and I hope to have dev merged into master tomorrow. I don't think anything you're doing should be impacted by dev, so go ahead and use master.

bradwilson:

bradwilson:
Thanks!

PR #2059, opened 2019-12-12
Make available xunit.console pause option also in Release builds.

tmds:
This option makes the console runner wait for the user to press a key at start,
which makes it easy to attach a debugger to the test process.

### DIFF
--- a/src/xunit.console/CommandLine.cs
+++ b/src/xunit.console/CommandLine.cs
@@ -39,9 +39,7 @@ namespace Xunit.ConsoleClient
 
         public bool NoLogo { get; protected set; }
 
-#if DEBUG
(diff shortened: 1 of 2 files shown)


tmds:
@bradwilson @onovotny ptal

I'm not sure what the -debug option does on Windows, but it doesn't seem to do anything for me on Linux.
Exposing this option provides a way to attach the vscode debugger to the test process on Linux.

cc @ViktorHofer

ViktorHofer:
@tmds if you are interested in this change for dotnet/runtime, feel free to submit a PR to https://github.com/dotnet/arcade/tree/master/src/Microsoft.DotNet.XUnitConsoleRunner (the fork of the not maintained runner here).

PR #2089, opened 2020-04-14
Remove duplicate reference from build.csproj

vchirikov:
Subj.
PackageReference `McMaster.Extensions.CommandLineUtils` is duplicated.


vchirikov:
@bradwilson ?

bradwilson:
Sorry to just get around to this now. There was indeed this dupe, as well as two others, so I've just done a commit for that: 943351ef2335d110806acd23e471f8c8b7bf88ea

Thanks!

Issue #621, opened 2015-10-09
Add test context object


takinosaji:
Hello. 
I've already investigated https://github.com/xunit/xunit/issues/398 and https://github.com/xunit/xunit/issues/416
Sorry for catching up this topic again, but as for me it's very important thing. 
I have to know is my test failed or not in my Dispose method for example to make screenshot on failure or so. How can I do it without having something like ex TestContext in NUnit?

agray:
+1, for being able to get the test name in a manner similar to NUnit.

TakinosaJi:
For example I build my own framework based on awesome xUnit and I extremely need such a features as current test status and so on.

anderssonjohan:
:+1: ****bump****

We use these features of Nunit when generating screenshots with Selenium WebDriver.
What we do today with Nunit that needs to be ported to xUnit:

```csharp
public static IWebDriver SaveScreenShot( this IWebDriver driver, string filename = null )
{
	var takesScreenShot = driver as ITakesScreenshot;
	if ( null == takesScreenShot ) return driver;

	var directory = TestContext.CurrentContext.WorkDirectory;
	var testName = TestContext.CurrentContext.Test.FullName;
	var name = filename ?? string.Format( CultureInfo.InvariantCulture, "Screenshot-{0}-{1:yyyyMMddHHmmss}", testName.Substring(0, testName.Length > 98 ? 98 : testName.Length), DateTime.Now );
	Screenshot sc = takesScreenShot.GetScreenshot();

	string file = Path.Combine( directory,  string.Concat( name, ".png" ) );
	Console.WriteLine( "Saving screenshot to {0}", file );
	sc.SaveAsFile( file, ImageFormat.Png );

	return driver;
}

public static IWebDriver SaveScreenShotOnTestFailure( this IWebDriver driver, string filename = null )
{
	if( TestContext.CurrentContext.Result.State == TestState.Cancelled && TestContext.CurrentContext.Result.Status == TestStatus.Failed )
		SaveScreenShot( driver, filename );
	return driver;
}
```

bradwilson:
We have no test context in xUnit.net.

@anderssonjohan Where do you imagine this code lives in an xUnit.net test?

anderssonjohan:
[...]

PR #2102, opened 2020-06-10
Expose ITest as a parameter in BeforeAfterTestAttribute

SimonCropp:
@bradwilson would this be possible to include in v3?

### DIFF
--- a/src/common.tests/TestDoubles/SpyBeforeAfterTest.cs
+++ b/src/common.tests/TestDoubles/SpyBeforeAfterTest.cs
@@ -1,5 +1,6 @@
 ﻿using System;
 using System.Reflection;
+using Xunit.Abstractions;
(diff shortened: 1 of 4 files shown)


bradwilson:
In general, I would say "yes", though I'd want to give some thought as to whether that exact set of parameters is the right thing. For example, if we go down the path of [creating a text context object](https://github.com/xunit/xunit/issues/621), maybe it makes more sense to pass the text context there (and thus keep the signature simple at a single parameter).

SimonCropp:
do u want test context to be an interface?

bradwilson:
I'm gonna merge this as-is and deal with the test context stuff at a later time.

bradwilson:
Because of the spaces to tabs conversion I had to hand-merge this, and committed it as 94f514ea.

Thanks!

PR #2104, opened 2020-06-13
allow assembly level BeforeAfterTestAttribute

SimonCropp:
thoughts on this one? and if u like it, is it something you want an explicit test for?


bradwilson:
Yeah, I think this would be fine. I'm definitely planning to try to enable more assembly-level stuff in v3. I would consider tests to be required. ;) But I don't think they'd be too complex to add.

SimonCropp:
ok i will push a test to this branch

bradwilson:
Ping on adding a test, am ready to merge otherwise. Thanks! 👍🏼

bradwilson:
(also, apologies in advance for the merge conflicts from spaces vs. tabs ☹️)

Issue #994, opened 2016-10-09
Make collection equality failures easier to spot.


Clockwork-Muse:
Currently, when collections are checked for equality, it can be very difficult to determine which element is different, because of the error message.  For instance, something like this:

```
[Fact]
public static void Eq_Short()
{
    List<int> expected = Enumerable.Repeat(0, 15).ToList();
    List<int> actual = Enumerable.Repeat(0, 14).Concat(Enumerable.Repeat(1, 1)).ToList();
    Assert.Equal(expected, actual);
}
```

produces this message:

```
Result Message:
Assert.Equal() Failure
Expected: List<Int32> [0, 0, 0, 0, 0, ...]
Actual:   List<Int32> [0, 0, 0, 0, 0, ...]
```

Instead, it would be better if, when there are too many elements to display, the message looked more like this:

```
Result Message:
Assert.Equal() Failure
Expected: List<Int32> [...14 identical elements..., 0]
Actual:   List<Int32> [...14 identical elements..., 1]
```

This is already how `string` works, so makes sense to do something similar:

```
[Fact]
public static void Eq_Short()
{
    string expected = string.Join(" ", Enumerable.Repeat(0, 15));
    string actual = string.Join(" ", Enumerable.Repeat(0, 14)) + " 1";
    Assert.Equal(expected, actual);
}
```

```
Result Message:
Assert.Equal() Failure
                                 ↓ (pos 28)
Expected: ···0 0 0 0 0 0 0 0 0 0 0
Actual:   ···0 0 0 0 0 0 0 0 0 0 1
                                 ↑ (pos 28)
```

bradwilson:
Any reason for the "14 identical elements" message rather than replicating the string-style message we have today?

Clockwork-Muse:
No, sorry.  I just picked something temporarily (before confirming the string message).  Closer to the string message is fine.

For sets/dictionaries, would the message take the order of the "actual" collection?  Since those collections usually don't have to be in order, it would be confusing if the message printed the expected in it's order, ie:
```
 Result Message:
Assert.Equal() Failure
                                   ↓ (pos 28)
Expected: ···11 10 0 2 4 6 8 3 7 9 5
Actual:   ···9 3 11 4 5 6 8 7 2 10 1
                                   ↑ (pos 28)
```

tmat:
[...]

PR #2142, opened 2020-09-01
Optimize using Span<T> Assert.Equal for T[] where T is unmanaged

### LINKED ISSUE #1907: Assert.Equals on two large arrays takes a long time
I noticed this when comparing 6GB worth of array data.  Assert.Equals takes a very long time: over 10 minutes.  Using Span.SequenceEqual instead is much faster: under 10s.  Here's a micro comparison on a 1MB array that shows a 220x difference.  Sample code attached. [xunitPerf.zip](https://github.com/xunit/xunit/files/2917368/xunitPerf.zip)

Method | Elapsed
------------ | -------------
Assert.Equal | 00:00:00.3534817
Span.SequenceEqual | 00:00:00.0015998

Not sure how many folks are doing assertions with large arrays, but you have an opportunity for an overload with better performance.

louis-z:
Fixes #1907 

- Needs https://github.com/xunit/assert.xunit/pull/39 to be merged
- Enable optimization by
  - Updating the `xunit/assert.xunit` submodule reference accordingly
  - Adding `XUNIT_SPAN` definition to [Directory.Build.props' `<DefineConstants>`](https://github.com/xunit/xunit/blob/f4be18e8bf926b3ec0ead036fca5df7fb2114116/src/Directory.Build.props#L11)

### DIFF
--- a/src/xunit.v3.assert.tests/Asserts/EqualityAssertsTests.cs
+++ b/src/xunit.v3.assert.tests/Asserts/EqualityAssertsTests.cs
@@ -1100,46 +1100,79 @@ public class EqualityAssertsTests
 	}
 
 #if XUNIT_ASSERT_ARRAY_AS_SPAN
-	public class Equal_ArrayOfUnmanagedT_AsSpan
+	public abstract class Equal_ArrayOfUnmanagedT_AsSpan<T>
+		where T : unmanaged, IEquatable<T>
 	{
 		[Fact]
-		public void IsMuchFaster()
+		public void BenchmarkPerformanceGain()
 		{
-			var buffer1 = new byte[1024 * 1024];
-			for (var i = 0; i < buffer1.Length; i++)
-				buffer1[i] = (byte)(i % 256);
-
-			var buffer2 = buffer1.ToArray();
+			var buffer1 = GetBigRandomBuffer();
+			var buffer2 = CopyBuffer(buffer1);
 
 			var sw = Stopwatch.StartNew();
 			Assert.Equal(buffer1, buffer2);
-			var eqFastElapsedTime = sw.Elapsed.TotalMilliseconds;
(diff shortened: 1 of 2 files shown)


bradwilson:
Thanks for this!

Unfortunately, this breaks a recent feature that was introduced to improve the output of enumerable Equal failures: https://github.com/xunit/xunit/issues/994

Specifically, all the new tests using `int[]` here have reverted to the old (undesired) behavior: https://github.com/xunit/xunit/commit/f4be18e8bf926b3ec0ead036fca5df7fb2114116?w=1

I suspect that the feature we'll end up with is `Span<T>` and `Memory<T>` specific assertions which trade off the quality of the message for performance. I don't think we want to change the existing behavior for `T[]` (especially not for just some types of `T`).

louis-z:
Thanks for the quick reply.

Maybe we could envisage calling the Span-based test first and, if it fails, use the slower one to get the improved output? (Assuming assertions fail less frequently than they succeed, we might still get an overall performance gain there.) A check could even be made on the array length to determine if it's worth risking the double calls, or just go ahead with the single call that offers the better output.

If that's not an option, no big deal, you can just close the 2 PRs. 😉

bradwilson:
That's not a bad idea. Bad performance for failing tests vs. better performance for passing tests is a good trade off. 😄

louis-z:
Here's an updated version: https://github.com/xunit/assert.xunit/pull/39/commits/f87879d828535f0e5f2a6ee5ca57002ecb834d7f

bradwilson:
I merged the assertion PR, thanks!

Looking at this, I'm not comfortable with having a performance test that masquerades as a unit test. We already have functional coverage for this (as evidenced by the tests that were originally failing after this assert was brought in). So I went ahead and enabled the feature in 7c71000a352fca686cf67f80b244865382784da1 and I think you can close this PR now.

PR #2147, opened 2020-09-05
Unit Tests for adding Span and Memory support: Issue 1966

jonlegendre:
Fix 1966
See associated pull request for the Asserts submodule for the changes.
Note: I just realized that this Issue was not identified as help wanted. No worries if it is rejected for that reason. This was a great learning experience for me in dealing with the innards of ref struct types.


bradwilson:
Thanks!

I made formatting changes, but since you did yours on the main branch of your fork, I didn't want to push my changes up into your fork, so I merged them by hand over here: f72992db8b175ae6d35d96c41753ffa6760d1757

(They are currently in the `dev` branch rather than the `main` branch because I want to do some further performance optimizations; namely, I want to ensure these tests run against Span instead of arrays or strings, or else we lose the performance benefits associated with using these data structures. I don't want to ask you to do this, so I'll do it myself on that branch, and then push it all into `main` when it's ready.)

PR #2161, opened 2020-10-01
Dispose ClassData class if disposable

jgowdy:
I am attempting to fix the fact that ClassData attributed classes that are IDisposable are not disposed.

### DIFF
--- a/src/xunit.v3.core/Sdk/Frameworks/Runners/XunitTheoryTestCaseRunner.cs
+++ b/src/xunit.v3.core/Sdk/Frameworks/Runners/XunitTheoryTestCaseRunner.cs
@@ -106,6 +106,8 @@ namespace Xunit.Sdk
 					foreach (var dataRow in data)
 					{
 						toDispose.AddRange(dataRow.OfType<IDisposable>());
+						if (data is IDisposable)
+							toDispose.Add((IDisposable)data);
 
 						ITypeInfo[]? resolvedTypes = null;
 						var methodToRun = TestMethod;


jgowdy:
Use case for this is security/secret related classes that throw exceptions if their finalizer is hit, since dispose is mandatory for erasure.  These exceptions are thrown when test data generated through the use of ClassData attributes doesn't get a Dispose call.  This is for: https://github.com/godaddy/asherah

gokhurana (approved):

src/xunit.v3.core/Sdk/Frameworks/Runners/XunitTheoryTestCaseRunner.cs
```diff
@@ -103,11 +103,12 @@ protected override async Task AfterTestCaseStartingAsync()
 						continue;
 					}
 
+					if (data is IDisposable)
```
bartelink, line 106: use `as` to avoid paying for the type check twice

jgowdy:
Would you mind adding a hacktoberfest-accepted label to this PR?  I didn't make this change specifically for hacktoberfest, I really need it, but I am participating in hacktoberfest.  Thanks!

bradwilson:
I've been doing some work with `DisposalTracker` on the [`dev` branch](https://github.com/xunit/xunit/tree/dev) to ensure that places which support `IDisposable` also support `IAsyncDisposable`.

1. Your change for `XunitTheoryTestCaseRunner.cs` should use the `DisposableTracker` instance that I've already set up for the individual data items. That should be a simple change.

2. Your change for `TheoryDiscoverer.cs` is going to be more problematic, because calling `IAsyncDisposable.DisposeAsync` requires that you be in an async method, which means changing the signature of `IXunitTestCaseDiscoverer.Discover` to be async as well.

I think I'd prefer to contemplate the impact of change #2 before I take it. In truth this should probably end up being an async method, but I don't want to rush into that change quite yet.

jgowdy:
Okay, I'll make the change to use DisposableTracker and see what making Discover be async looks like.

PR #2190, opened 2020-11-10
Add missing word "A good example of this testing..."

maxalmonte14:
Hello xUnit maintainers 👋

I think there is a missing word in the modified paragraph, please let me know if it was supposed to be like that.

Thanks in advance, have a great rest of the day.


bradwilson:
Thanks!

PR #2192, opened 2020-11-11
Add environment variable to override AppDomain.Unload

sharwell:
This pull request provides a mitigation path for dotnet/roslyn#49024.

I considered three approaches:

* A new option `appDomainUnload`, with default value `true`. This is the cleanest option for a user, but is difficult to implement throughout the internals without breaking API changes.
* Changing `AppDomainSupport` to have an enum flag for this. This seemed like a reasonable possibility, but likely unnecessarily complex considering the current maintenance strategy for the v2 branch.
* An environment variable. This is the simplest option, and likely only works for CI scenarios, but at least it mitigates the acute problems we're facing today.

### DIFF
--- a/src/xunit.runner.utility/AppDomain/AppDomainManager_AppDomain.cs
+++ b/src/xunit.runner.utility/AppDomain/AppDomainManager_AppDomain.cs
@@ -36,6 +36,8 @@ namespace Xunit
 
         public bool HasAppDomain => true;
 
+        private bool SkipAppDomainUnload => string.Equals(Environment.GetEnvironmentVariable("XUNIT_SKIP_APPDOMAIN_UNLOAD"), "true", StringComparison.OrdinalIgnoreCase);
+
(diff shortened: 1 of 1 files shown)


sharwell:
@bradwilson I'm having trouble building this for test. If CI could publish a build to the package feed we can update our CI and verify over the next week that the timeouts are gone.

bradwilson:
There's no CI for v2 any more, because there are no plans to release any more v2 releases. The AppVeyor account was closed down some time ago.

sharwell:
@bradwilson I verified locally that this change fixes at least one of the common delays. This is certainly a multiple-times-per-day pain point for the Roslyn team.

bradwilson:
I could try to help figure out your build issues so that you could publish your own internal package.

sharwell:
📝 I am able to build this from https://github.com/xunit/xunit/commit/dde9f82.

sharwell:
@bradwilson this is no longer needed. We implemented a change to patch the x86/x64 assembly in memory at runtime and it met our needs.

Issue #2227, opened 2021-01-24
Parallelizing test assemblies in TeamCity may cause some tests to not be reported


alefranz:
I've been noticing for a long time that when running on Teamcity tests on a solution with many assemblies, using parallelize assemblies, sometimes a few tests are not reported (e.g. one or two every thousand)
On close inspection of the logs, I've noticed that sometimes two logs lines e.g.
```
##teamcity[testFinished ...] ##teamcity[testFinished ...]
```
and when this happens TeamCity only parse the second message, so the first test outputted is not reported.

This is really frequent when running on Linux.

bradwilson:
I just validated in [ConsoleRunnerLogger](https://github.com/xunit/xunit/blob/v2/src/xunit.runner.utility/Utility/ConsoleRunnerLogger.cs) that all calls to `Console.WriteLine()` have locks around them, so it's not clear to me how we can end up with two messages on the same line. ☹️

alefranz:
Hi Brad,
thanks for looking at this.
Please have a look at my proposed PR #2228 , as on linux/mac setting the color is causing a Write and I guess if the escape sequences get mixed up it could somehow be lost

alefranz:
Raised to JetBrains on [TW-70266](https://youtrack.jetbrains.com/issue/TW-70266)

bradwilson:
Fixed in #2228.

PR #2228, opened 2021-01-24
Fix Console colors concurrency handling

alefranz:
Hello,
While I was looking for a cause of https://github.com/xunit/xunit/issues/2227, I've noticed that the console colors changes are not part of the lock, which will cause the colors to be mixed together.

I believe this would also potentially cause invalid escape codes to be emitted, but I am not sure if it could generate codes that can cause a previous character (the newline) to be lost).
I thought it was worth it to raise this small PR even if it has not been discussed yet in the issue.

There are currently not test and would not be easy to write one for this problem, so I haven't added one.

Please let me know what do you think.

Thank you,
 Alessio


bradwilson:
Thanks!

PR #2242, opened 2021-02-03
Backport: Fix Console colors concurrency handling

alefranz:
Hi Brad, 
as the bug fixed in https://github.com/xunit/xunit/pull/2228 is also present on v2, it would be nice to backport it.
However not sure on what is the current policy and process for v2 bug fixes, so let me know if this is not the desirable action.

Thank you,
 Alessio

### DIFF
--- a/src/xunit.runner.utility/Utility/ConsoleRunnerLogger.cs
+++ b/src/xunit.runner.utility/Utility/ConsoleRunnerLogger.cs
@@ -38,32 +38,32 @@ namespace Xunit
         /// <inheritdoc/>
         public void LogError(StackFrameInfo stackFrame, string message)
         {
-            using (SetColor(ConsoleColor.Red))
-                lock (LockObject)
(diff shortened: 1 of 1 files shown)


bradwilson:
We have no plans to release any more v2 builds. Thanks, though!

alefranz:
HI Brad,
thanks for looking into it.

The inability to fix the current release is unfortunate as I've seen https://github.com/xunit/xunit/issues/2227 causing a fluctuation of about 0.5% in number of tests reported, which means to be able to trust CI you are forced to run all tests sequentially.

My main worry is that many projects are affected but it is unnoticed as the number of reported tests is a metric not closely monitored, and this could mean a risk of failed tests slipping through CI.

I will try to investigate if other CI systems are affected.

bradwilson:
Now that we've decided to release at least one more v2 release, I'm merging this. Thanks again!

PR #2299, opened 2021-05-05
Fix typo in comparisons.markdown

ericmutta:
The sentence:

> xUnit.net ships with support for data-driven tests call Theories

Should read:

> xUnit.net ships with support for data-driven tests **called** Theories


bradwilson:
Thanks!

PR #2301, opened 2021-05-10
Fix typo in running-tests-in-parallel.html

ericmutta:
The following sentence:

> There are several default pieces of behavior that be configured by the...

Should read:

> There are several default pieces of behavior that **can** be configured by the...


bradwilson:
Thanks!

PR #2302, opened 2021-05-10
Fix another typo in running-tests-in-parallel.html

ericmutta:
The following sentence:

> ...and command line options can be used to configuration the parallelism options...

Should read:

> ...and command line options can be used to **configure** the parallelism options...


bradwilson:
Thanks!

PR #2303, opened 2021-05-10
Fix typos in shared-context.html

ericmutta:
The following sentence:

> Just before the first tests in MyDatabaseTests is run...

Should read:

> Just before the first test in MyDatabaseTests is run...

The following sentence:

> ...share a single object instance among tests in several test class.

Should read:

> ...share a single object instance among tests in several test classes.

### DIFF
--- a/docs/shared-context.html
+++ b/docs/shared-context.html
@@ -207,7 +207,7 @@ public class MyDatabaseTests : IClassFixture<DatabaseFixture>
 {% endhighlight %}
 
 <p>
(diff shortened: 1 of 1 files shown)


bradwilson:
Thanks!

ericmutta:
Many thanks @bradwilson for accepting my PRs!

I just started writing my first xUnit tests yesterday and it has dramatically improved the way I write tests (the Theory and MemberData attributes are especially cool!). Thanks for maintaining this framework 👍

PR #2341, opened 2021-07-02
1822 test display name for doubles

### LINKED ISSUE #1822: TestDisplayName does not print enough digits to differentiate certain inputs
**Repro Steps**
* Create a new 'xUnit Test Project (.NET Core)
  * I used the built-in template from VS 2017 15.8
* Set the contents of `UnitTest1.cs` to the following:
```csharp
using Xunit;

public class UnitTest1
{
    [Theory]
    [InlineData(1.17549421E-38f)]
    [InlineData(1.17549435E-38f)]
    public void Test1(float f)
    {

    }
}
```
* Run the tests and examine the "TestDisplayName" that is printed to the result logs

**Expected Behavior**
The inline data is printed with enough precision to differentiate the two inputs.

**Actual Behavior**
The `TestDisplayName` for both scenarios is `Test1(f: 1.175494E-38)`.

Xcodo:
Before you submit a PR...

  https://github.com/xunit/xunit/issuesq=is%3Aopen+is%3Aissue+label%3A%22help+wanted%22)? Not actually 'help wanted', sorry I got carried away. It is on the roadmap. Feel free to reject if it's not the way you want to go!
* Did you read the project governance @ https://xunit.net/governance ? Yes
* Does the code follow existing coding styles? (tabs, comments, no regions, etc.) Yes
* Did you write unit tests? Yes

Requires [assert.xunit#47](https://github.com/xunit/assert.xunit/pull/47).

Fixes #1822.


bradwilson:
Thanks!

src/xunit.v3.assert.tests/Asserts/Sdk/ArgumentFormatterTests.cs
```diff
@@ -120,6 +120,18 @@ public static void CharacterValue(char value, string expected)
 			Assert.Equal(expected, ArgumentFormatter.Format(value));
 		}
 
+		[CulturedFact]
+		public static void FloatValue()
+		{
+			Assert.Equal(Math.PI.ToString("G7"), ArgumentFormatter.Format((float)Math.PI));
```
tannergooding, line 126: `G7` isn't enough digits for `float`, you need at least `9` digits to guarantee uniqueness
Likewise, `G19` is more precise than necessary for `double`, you only need `17` to guarantee uniqueness

`R` will generally be unique on .NET Framework (outside bugs) and will always be unique on .NET Core 3.0+

Issue #1548, opened 2017-11-03
AssemblyRunner's "OnTestFailed" does not give enough information for AggregateExceptions.


danrozenberg:
Hello, we have some tests running through the ``AssemblyRunner`` class, and it works fine, in general.

However, when a test run fails with an ``AggregateException``, it becomes hard to gather more information about the failure... Consider the example:

```cs
            runner.OnTestFailed = info =>
            {
               _logger.LogInformation($"The following test has failed: {info.TestDisplayName} after {info.ExecutionTime} seconds, with error: {info.ExceptionMessage} and stack trace {info.ExceptionStackTrace}. The exception type is: {info.ExceptionType}.");
            };
```

When a test fails, here is what would be output:

```The following test has failed: MyTest after 30 seconds, with error: One or more errors occurred. and stack trace . The exception type is: System.AggregateException.```

Sadly, I couldn't find anything in ``TestFailedInfo`` that could display the inner exceptions. Is there any way to see all inner exceptions? If not, could something like this be added in the future?

Thank you!

danrozenberg:
Can we have this for target 3.0? :)

bradwilson:
I have not yet decided what to do with all the existing open issues w/rt 3.0. I don't even know if AssemblyRunner is something that's going to exist in 3.0. I'm leaving them all as-is for now, to be re-evaluated once 3.0 starts to take shape.

So merely being open is good enough to be a candidate for 3.0. :)

ohadschn:
It should be noted that `AggregateException` is only one such case our of many. For example, `HttpRequestException` can say something like `The SSL connection could not be established, see inner exception`. And it's not necessarily the inner exception(s) that's missing, for example in some type load exceptions it's some other property if I recall correctly. And of course inner exception could also have inner exceptions, and other properties, and so on. The point it, only `Exception.ToString` will work correctly.

Fortunately, the solution is very simple, add something like `TestFailedInfo.FullExceptionInfo` and populate it with `exception.TosSring()`.

> I don't even know if AssemblyRunner is something that's going to exist in 3.0

I hope there will be an equivalent though? We heavily rely on the ability to execute tests from a loaded assembly **without** launching an additional process.

bradwilson:
[...]

PR #2368, opened 2021-08-25
Print full exception details in error message

ymarcov:
When a test fails, print full exception details.

This provides a crucial stop-gap for many developers needing good log messages from CI test runs (in our case an important service within Microsoft).

* Fixes active issue https://github.com/xunit/xunit/issues/1548
* Relevant tests have been adjusted and all tests pass following this change.

### DIFF
--- a/src/xunit.v3.common/v3/Abstractions/_IErrorMetadata.cs
+++ b/src/xunit.v3.common/v3/Abstractions/_IErrorMetadata.cs
@@ -22,6 +22,11 @@ namespace Xunit.v3
 		string[] Messages { get; }
 
 		/// <summary>
+		/// Gets the details of the exception(s).
+		/// </summary>
+		string? Details { get; }
+
+		/// <summary>
 		/// Gets the stack trace(s) of the exception(s).
 		/// </summary>
 		string?[] StackTraces { get; }
--- a/src/xunit.v3.common/v3/Messages/_ErrorMessage.cs
+++ b/src/xunit.v3.common/v3/Messages/_ErrorMessage.cs
@@ -12,6 +12,7 @@ namespace Xunit.v3
 		int[]? exceptionParentIndices;
 		string?[]? exceptionTypes;
 		string[]? messages;
+		string? details;
 		string?[]? stackTraces;
 
 		/// <inheritdoc/>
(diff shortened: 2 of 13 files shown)


dnfadmin:
[CLA assistant check](https://cla.dotnetfoundation.org/xunit/xunit?pullRequest=2368) <br/>Thank you for your submission, we really appreciate it. Like many open source projects, we ask that you sign our [Contributor License Agreement](https://cla.dotnetfoundation.org/xunit/xunit?pullRequest=2368) before we can accept your contribution.<br/><br/>:x: ymarcov [sign now](https://cla.dotnetfoundation.org/xunit/xunit?pullRequest=2368)<br/><sub>You have signed the CLA already but the status is still pending? Let us [recheck](https://cla.dotnetfoundation.org/check/xunit/xunit?pullRequest=2368) it.</sub>

ymarcov:
The main point is to get the output of `exception.ToString()`, as it already contains the entire exception tree. From what I saw, at the point where the error details reach the `AssemblyRunner` those details are already lost, as they've been extracted into `Message`, `Stacktraces`, etc.

To pass them along I had to go back to the point of capture, which is deeper inside.

Calling `exception.ToString()` also means that formatting is delegated, and because at that point we're dealing with free-text, filtering it manually is a bit precarious.

Perhaps a more elegant solution would be to implement `Stacktraces` as a tree rather than a list. That would allow us to represent the exception tree propertly, as well as format it accordingly. Another equivalent option is to add some textual metadata as prefixes to the stacktrace strings that represents the location of the exception in the hierarchy (e.g. 0:2:1 would mean Root->3rd sibling->2nd sibling), that we could format an indented hierarchy out of in the output. This is of course a very contrived exception tree, but I think it does need to be solved for the general case, just as `exception.ToString()` does.

bradwilson:
@ymarcov The hierarchy is preserved via `ExceptionParentIndices` (see https://github.com/xunit/abstractions.xunit/blob/main/src/xunit.abstractions/Messages/BaseInterfaces/IFailureInformation.cs).

ymarcov:
@bradwilson Oh nice. I see that `xunit.v3.runner.utility` already depends on `xunit.abstractions`. So if I understand correctly, I could simply utilize `IFailureInformation` to print out the full hierarchy that's already been captured at that point?

bradwilson:
As part of the effort to clean up older PRs, this is being closed due to age. If this is still applicable with v3, please open a new issue to discuss the feature so it can be re-evaluated and categorized.

Thanks!

Issue #1855, opened 2018-11-19
System.InvalidOperationException: Collection was modified after the enumerator was instantiated.


benaadams:
Seems in Disposal
```
Unhandled Exception: System.InvalidOperationException: Collection was modified after the enumerator was instantiated.
   at System.Collections.Generic.Stack`1.Enumerator.MoveNext()
   at Xunit.Sdk.DisposalTracker.Dispose() in C:\Dev\xunit\xunit\src\xunit.execution\Sdk\DisposalTracker.cs:line 26
   at Xunit.Sdk.ExtensibilityPointFactory.Dispose() in C:\Dev\xunit\xunit\src\xunit.execution\Sdk\ExtensibilityPointFactory.cs:line 54
   at Xunit.Sdk.TestFramework.Dispose() in C:\Dev\xunit\xunit\src\xunit.execution\Sdk\Frameworks\TestFramework.cs:line 49
   at System.Threading.ThreadPoolWorkQueue.Dispatch()
```

Seen in CI https://ci.dot.net/job/dotnet_coreclr/job/master/job/x64_checked_osx10.12_innerloop_tst_prtest/7197/console

https://github.com/dotnet/coreclr/pull/21076

bradwilson:
I suppose it's too much to hope for a simple repro?

benaadams:
Might be a secondary error, their OSX environment seems a bit flaky

am11:
It is one of the hard to repro issue. We seldom see this exception under high load in different legs of dotnet/runtime CI: https://github.com/dotnet/runtime/issues/11063. If we switch from `foreach` to `for` loop, I think it will prevent it from happening.

@bradwilson, if this is an acceptable change: https://github.com/am11/xunit/commit/7ecc5dcbef5a16cedc734c81b899815019ad0aa7, and issue can falls under `help wanted` category, I can issue a PR. :)

bradwilson:
I did recently [add a bunch of double disposal protection](https://github.com/xunit/xunit/commit/9e8d5d13fbda4224de7f996d6812855650980c1c) including in `DisposalTracker`, which should stop, and in the process I stopped reusing the `DisposalTracker` from `ExtensibilityPointFactory`, which I hope will put this bug to rest.

@am11 Your fix looks like it would probably not dispose everything, because as you pop items off the stack the item count would go down, so you'll end up only disposing about the first half of the items (total count goes down as your item count goes up).

[...]

PR #2375, opened 2021-09-08
Switch from foreach to for loop in DisposalTracker

maryamariyan:
Based fix on this commit: https://github.com/am11/xunit/commit/7ecc5dcbef5a16cedc734c81b899815019ad0aa7 

Fixes (v2 branch): https://github.com/xunit/xunit/issues/1855
Fixes: https://github.com/dotnet/runtime/issues/11063 (this fix would help improve the test pass rate in dontet/runtime infra.)

@am11 @bradwilson do we know if this issue is caused by a threading issue? if so maybe it makes preparing a test here a bit more trivial.

/cc: @ViktorHofer

---

TODO:
- [x] Add test

### DIFF
--- a/src/xunit.execution/Sdk/ExtensibilityPointFactory.cs
+++ b/src/xunit.execution/Sdk/ExtensibilityPointFactory.cs
@@ -39,12 +39,8 @@ namespace Xunit.Sdk
 
             var disposable = result as IDisposable;
             if (disposable != null)
-            {
                 lock (disposalTracker)
-                {
                     disposalTracker.Add(disposable);
-                }
-            }
 
             return result;
         }
@@ -56,9 +52,7 @@ namespace Xunit.Sdk
         {
             instances.Clear();
             lock (disposalTracker)
-            {
                 disposalTracker.Dispose();
-            }
         }
 
         /// <summary>
--- a/src/xunit.execution/Sdk/ExtensibilityPointFactory.cs
+++ b/src/xunit.execution/Sdk/ExtensibilityPointFactory.cs
@@ -39,7 +39,12 @@ namespace Xunit.Sdk
 
             var disposable = result as IDisposable;
             if (disposable != null)
-                disposalTracker.Add(disposable);
+            {
+                lock (disposalTracker)
+                {
+                    disposalTracker.Add(disposable);
+                }
+            }
 
             return result;
         }
@@ -50,7 +55,10 @@ namespace Xunit.Sdk
         public static void Dispose()
         {
             instances.Clear();
-            disposalTracker.Dispose();
+            lock (disposalTracker)
+            {
+                disposalTracker.Dispose();
+            }
         }
 
         /// <summary>
--- a/src/xunit.execution/Sdk/ExtensibilityPointFactory.cs
+++ b/src/xunit.execution/Sdk/ExtensibilityPointFactory.cs
@@ -11,9 +11,6 @@ namespace Xunit.Sdk
     /// </summary>
     public static class ExtensibilityPointFactory
     {
-        static readonly DisposalTracker disposalTracker = new DisposalTracker();
-        static readonly ConcurrentDictionary<Tuple<Type, IMessageSink>, object> instances = new ConcurrentDictionary<Tuple<Type, IMessageSink>, object>();
-
         static object CreateInstance(IMessageSink diagnosticMessageSink, Type type, object[] ctorArgs)
         {
             ctorArgs = ctorArgs ?? new object[0];
@@ -37,22 +34,7 @@ namespace Xunit.Sdk
                 }
             }
 
-            var disposable = result as IDisposable;
-            if (disposable != null)
-                lock (disposalTracker)
-                    disposalTracker.Add(disposable);
-
             return result;
-        }
-
-        /// <summary>
-        /// Disposes the instances that are contained in the cache.
-        /// </summary>
-        public static void Dispose()
-        {
-            instances.Clear();
-            lock (disposalTracker)
-                disposalTracker.Dispose();
         }
 
         /// <summary>
@@ -69,7 +51,7 @@ namespace Xunit.Sdk
         /// <returns>The instance of the type.</returns>
         public static TInterface Get<TInterface>(IMessageSink diagnosticMessageSink, Type type, object[] ctorArgs = null)
         {
-            return (TInterface)instances.GetOrAdd(Tuple.Create(type, diagnosticMessageSink), () => CreateInstance(diagnosticMessageSink, type, ctorArgs));
+            return (TInterface)CreateInstance(diagnosticMessageSink, type, ctorArgs);
         }
 
         /// <summary>
--- a/src/xunit.execution/Sdk/Frameworks/TestFramework.cs
+++ b/src/xunit.execution/Sdk/Frameworks/TestFramework.cs
@@ -45,7 +45,6 @@ namespace Xunit.Sdk
             // of things that will be disconnected.
             await Task.Delay(1);
 
-            ExtensibilityPointFactory.Dispose();
             DisposalTracker.Dispose();
 
             LongLivedMarshalByRefObject.DisconnectAll();
--- a/test/test.xunit.execution/Sdk/ExtensibilityPointFactoryTests.cs
+++ b/test/test.xunit.execution/Sdk/ExtensibilityPointFactoryTests.cs
@@ -1,6 +1,7 @@
 ﻿using System;
 using System.Collections.Generic;
 using System.Linq;
+using System.Threading.Tasks;
 using Xunit;
 using Xunit.Abstractions;
 using Xunit.Sdk;
@@ -46,6 +47,12 @@ public class ExtensibilityPointFactoryTests
             var result = ExtensibilityPointFactory.GetXunitTestCollectionFactory(spy, attr, assembly);
 
             Assert.IsType(expectedType, result);
+        }
+
+        [Fact]
+        public void DisposeOnMultipleThreadsThrowsNoException()
+        {
+            Parallel.For(0, 2, (_) => { for (int i = 0; i < 10000; i++) { ExtensibilityPointFactory.Dispose(); } });
         }
 
         [Fact]
--- a/test/test.xunit.execution/Sdk/ExtensibilityPointFactoryTests.cs
+++ b/test/test.xunit.execution/Sdk/ExtensibilityPointFactoryTests.cs
@@ -50,12 +50,6 @@ public class ExtensibilityPointFactoryTests
         }
 
         [Fact]
-        public void DisposeOnMultipleThreadsThrowsNoException()
-        {
-            Parallel.For(0, 2, (_) => { for (int i = 0; i < 10000; i++) { ExtensibilityPointFactory.Dispose(); } });
-        }
-
-        [Fact]
         public void UserCanChooseCustomCollectionFactory()
         {
             var factoryType = typeof(MyTestCollectionFactory);


src/xunit.execution/Sdk/DisposalTracker.cs
```diff
@@ -23,8 +23,10 @@ public void Add(IDisposable disposable)
         /// <inheritdoc/>
         public void Dispose()
         {
-            foreach (var disposable in toDispose)
-                disposable.Dispose();
+            for (int i = 0; i < toDispose.Count; ++i)
```
am11, line 26:
```suggestion
            for (int i = toDispose.Count; i > 0; --i)
```
to avoid fetching the count (which stack.Count is changing every iteration) [example run](https://sharplab.io/#v2:EYLgxg9gTgpgtADwGwBYA0AXEUCuA7AHwAEAmARgFgAoIgBgAIiyUBua6gNwEMp6F6AvPTwwA7vQDKGLmADWAHgCWeDAD4AFAEo2VaggB0ABRwBnABbqy2vUdMWS1qgePn1AZkfUAZtHXKM9IqCfPoAwhD4GCyB9Kr0tNFwcIqa1PTpjGQAnOrOEAAOWo5AA)

am11:
> do we know if this issue is caused by a threading issue? if so maybe it makes preparing a test here a bit more trivial.

Yes, I seems to happen intermittently under high load when dozens of test cases are being discovered during the initialization.

IMO, the most safest/natural type in this case is `ConcurrentStack`: https://github.com/xunit/xunit/commit/3d61952f181cf2e2e9d95670d32e8ee819e84a2d. xUnit code is using `ConcurrentDictionary` in other places, so it won't be too strange I suppose. 😁

src/xunit.execution/Sdk/DisposalTracker.cs
bradwilson: This also would've failed to dispose everything, because the count was changing (it would've only disposed roughly half the items).

src/xunit.execution/xunit.execution.csproj
```diff
@@ -2,11 +2,11 @@
 
   <PropertyGroup>
     <AssemblyName Condition=" '$(TargetFramework)' == 'net452' ">xunit.execution.desktop</AssemblyName>
-    <AssemblyName Condition=" '$(TargetFramework)' == 'netstandard1.1' ">xunit.execution.dotnet</AssemblyName>
+    <AssemblyName Condition=" '$(TargetFramework)' == 'netstandard2.0' ">xunit.execution.dotnet</AssemblyName>
     <DefineConstants>$(DefineConstants);XUNIT_FRAMEWORK</DefineConstants>
     <GenerateDocumentationFile>true</GenerateDocumentationFile>
     <RootNamespace>Xunit.Sdk</RootNamespace>
-    <TargetFrameworks>net452;netstandard1.1</TargetFrameworks>
+    <TargetFrameworks>net452;netstandard2.0</TargetFrameworks>
```
ericstj, line 9: This change should be undone

ericstj:
Here's a one-liner that repros the bug:
```C#
Parallel.For(0, 2, (_) => { for (int i = 0; i < 10000; i++) { ExtensibilityPointFactory.Dispose(); } });
```

```
Unhandled exception. System.AggregateException: One or more errors occurred. (Collection was modified after the enumerator was instantiated.)
 ---> System.InvalidOperationException: Collection was modified after the enumerator was instantiated.
   at System.Collections.Generic.Stack`1.Enumerator.MoveNext()
   at Xunit.Sdk.DisposalTracker.Dispose() in C:\Dev\xunit\xunit\src\xunit.execution\Sdk\DisposalTracker.cs:line 26
   at Xunit.Sdk.ExtensibilityPointFactory.Dispose() in C:\Dev\xunit\xunit\src\xunit.execution\Sdk\ExtensibilityPointFactory.cs:line 53
   at xunitTest.Program.<>c.<Main>b__0_0(Int32 _) in C:\scratch\xunitTest\Program.cs:line 13
   at System.Threading.Tasks.Parallel.<>c__DisplayClass19_0`1.<ForWorker>b__1(RangeWorker& currentWorker, Int32 timeout, Boolean& replicationDelegateYieldedBeforeCompletion)
--- End of stack trace from previous location ---
   at System.Threading.Tasks.Parallel.<>c__DisplayClass19_0`1.<ForWorker>b__1(RangeWorker& currentWorker, Int32 timeout, Boolean& replicationDelegateYieldedBeforeCompletion)
   at System.Threading.Tasks.TaskReplicator.Replica`1.ExecuteAction(Boolean& yieldedBeforeCompletion)
   at System.Threading.Tasks.TaskReplicator.Replica.Execute()
   --- End of inner exception stack trace ---
   at System.Threading.Tasks.TaskReplicator.Run[TState](ReplicatableUserAction`1 action, ParallelOptions options, Boolean stopOnFirstFailure)
   at System.Threading.Tasks.Parallel.ForWorker[TLocal](Int32 fromInclusive, Int32 toExclusive, ParallelOptions parallelOptions, Action`1 body, Action`2 bodyWithState, Func`4 bodyWithLocal, Func`1 localInit, Action`1 localFinally)
--- End of stack trace from previous location ---
   at System.Threading.Tasks.Parallel.ThrowSingleCancellationExceptionOrOtherException(ICollection exceptions, CancellationToken cancelToken, Exception otherException)
   at System.Threading.Tasks.Parallel.ForWorker[TLocal](Int32 fromInclusive, Int32 toExclusive, ParallelOptions parallelOptions, Action`1 body, Action`2 bodyWithState, Func`4 bodyWithLocal, Func`1 localInit, Action`1 localFinally)
   at System.Threading.Tasks.Parallel.For(Int32 fromInclusive, Int32 toExclusive, Action`1 body)
   at xunitTest.Program.Main(String[] args) in C:\scratch\xunitTest\Program.cs:line 13
```

Perhaps we can add a test case?

bradwilson:
FWIW, I removed ExtensibilityPointFactory.Dispose (well, it had been moved to DisposeAsync) precisely because it has a very significant bug related to improper caching. Might be worth also just removing the Dispose functionality in v2 since the bug has been around pretty much since ExtensibilityPointFactory was introduced.
https://github.com/xunit/xunit/commit/b4ca6874dc499b20cc42efaa36d7b982ce74f7ff#diff-25ba7e6eb02a399570c746572389784ac3b5d86b428de88357a671c508d8d9e2

ericstj:
> Might be worth also just removing the Dispose functionality in v2

If it was removed in the same way it would be binary breaking.  If instead we just removed the internal usage / method body what's the likelihood it would break folks?  Do you imagine there is anyone out there that was relying on the lifetime management they were getting from `ExtensibilityPointFactory`?

bradwilson:
> If it was removed in the same way it would be binary breaking. If instead we just removed the internal usage / method body what's the likelihood it would break folks? Do you imagine there is anyone out there that was relying on the lifetime management they were getting from `ExtensibilityPointFactory`?

The only code that should be calling that Dispose is internal code. It was never intended to be called by third parties.

src/xunit.execution/Sdk/ExtensibilityPointFactory.cs
```diff
@@ -39,7 +39,12 @@ static object CreateInstance(IMessageSink diagnosticMessageSink, Type type, obje
 
             var disposable = result as IDisposable;
             if (disposable != null)
-                disposalTracker.Add(disposable);
+            {
```
bradwilson, line 42: Coding style for this project is to not use curly braces for single statement bodies (that would include both the `if` and the `lock` here).

maryamariyan:
PR ready for review

ericstj:
> the Dispose here happens basically just before the runner itself tears down.

We did notice this happening many times throughout the execution of tests (many assemblies).  So at least in our repro it might be possible for someone to observe a Dispose not happening right away if they happened to depend on that Dispose.  We didn't depend on the dispose since none of the objects that were created from `ExtensibilityPointFactory` actually implemented `IDisposable`, but I don't know if others might have disposable `IDataDiscoverer`, `ITestCaseOrderer`, `ITestCollectionOrderer`, `ITestFrameworkTypeDiscoverer`, `ITraitDiscoverer`, or `IXunitTestCollectionFactory`.

src/xunit.execution/Sdk/Frameworks/TestFramework.cs
```diff
@@ -45,7 +45,6 @@ public async void Dispose()
             // of things that will be disconnected.
             await Task.Delay(1);
 
-            ExtensibilityPointFactory.Dispose();
```
ericstj, line 48: I think @bradwilson was suggesting moving the `Dispose` method entirely, along with the disposal tracker and instance collection, but I might be mistaken.  I think if we remove this without removing the collection it could result in the instance collection growing unbounded?

maryamariyan: this would be a bigger change. I thought we wanted to do the minimal amount of work to get unblocked

bradwilson:
Maybe I missed it, but wasn't this also supposed to fix the actual bug in DisposalTracker? That class is still in use even after the changes proposed here.

maryamariyan:
> Maybe I missed it, but wasn't this also supposed to fix the actual bug in DisposalTracker? That class is still in use even after the changes proposed here.

what do you think about doing the minimal fix now to unblock dotnet/runtime and then tackle the DisposalTracker bug separately?

/cc: @ericstj

bradwilson:
> what do you think about doing the minimal fix now to unblock dotnet/runtime and then tackle the DisposalTracker bug separately?

Are you positive that 100% of your failures were in the ExtensibilityFactory?

bradwilson:
Let me be more clear.

This PR was originally opened to fix the DisposalTracker bug. At some point it pivoted to fix a _**different**_ bug in ExtensibilityFactory. I'm okay with fixing both bugs, but it's strange to me that this PR shifted from fixing a different bug than it was opened for rather than just opened a second PR. The code provided here no longer matches the text or description of the PR, as it is **_not_** a fix for #1855.

ericstj:
> Maybe I missed it, but wasn't this also supposed to fix the actual bug in DisposalTracker? That class is still in use even after the changes proposed here.

We didn't observe any cases where a `DisposalTracker` was accessed from multiple threads other than when called by `ExtensibilityPointFactory`.  Our assumption is that `DisposalTracker` is not thread-safe and requires more than just a `ConcurrentDictionary` to make it so;  For instance, it would need locking to prevent calls to `Add` happening between enumeration and clear in `Dispose`.  Rather than try to add thread-safety to an instance type that wasn't designed for it, we should just fix the cases that try to access it from multiple threads.  `ExtensibilityPointFactory` is clearly exposing API that accesses `DisposalTracker` from multiple threads.  In every case of the issue we're attempting to fix we observed `ExtensibilityPointFactory.Dispose` on the stack.

> This PR was originally opened to fix the DisposalTracker bug. At some point it pivoted to fix a different bug in ExtensibilityFactory. I'm okay with fixing both bugs, but it's strange to me that this PR shifted from fixing a different bug than it was opened for rather than just opened a second PR. The code provided here no longer matches the text or description of the PR, as it is not a fix for #1855.

This PR was opened with a community suggestion to fix #1855 based on a symptom, not root cause.  #1855 does show the same call-stack we're discussing here with `ExtensibilityPointFactory`.  This does fix that bug.

We root caused the issue reported and found a more targeted fix, as discussed above.  We believe this is the only issue.

bradwilson:
>This PR was opened with a community suggestion to fix #1855 based on a symptom, not root cause.

So the PR fixes the symptom and not the cause. Why not fix the cause?

bradwilson:
Another important point: I've already fixed both of these bugs in `main`. It should be trivial to port that code back here into `v2`.

ericstj:
> So the PR fixes the symptom and not the cause. Why not fix the cause? 

The root cause here is concurrent access to a non-thread-safe type: `DisposalTracker`.  The smallest fix which we are proposing is to make that access thread safe, by locking around the access to the non-thread-safe static member.  We're doing this rather than changing `DisposalTracker`  to be thread-safe since we have no evidence that `DisposalTracker` was meant to be thread safe in other usage.  

We do believe that this is a fix to the cause of the bug, but I may be misunderstanding what you mean here by cause.  I could imagine are you suggesting we should remove the static API which causes the concurrent access?  We could consider doing that as we discussed earlier, but it seems to be a larger fix.  We will do what you feel is best, we are trying to gauge your bug-bar for servicing v2.

If we apply the typical bug bar for .NET libraries servicing, we would not remove public API (even if meant to be called only internally) or add thread-safety to an instance type.  We would save these sort of changes for a major release where we permitted breaking changes or were willing to take bigger risks.

> Another important point: I've already fixed both of these bugs in main. It should be trivial to port that code back here into v2.

We could do that, but every change carries with it some risk.  If we pull back larger changes they could have side effects that impact existing users and make it harder for them to adopt a new version of the library.

Again, we'll defer to your judgement here on what you feel is best for v2.

bradwilson:
> The root cause here is concurrent access to a non-thread-safe type: `DisposalTracker`.

Agreed. And this PR (as currently submitted) does not fix that issue.

> The smallest fix which we are proposing is to make that access thread safe, by locking around the access to the non-thread-safe static member.

That's not what this PR does. It removes one instance (but not all instances) of calls to `DisposalTracker`. The code that was submitted here is to fix a different caching issue with `ExtensibilityFactory`. The actual issue being fixed in this PR is that we can't actually safely cache anything, and that has nothing to do with thread safety; it has to do with the fact that it's legal to reuse message sink instances, which causes us to incorrectly cache type lookups.

It's coincidental that removing the cache also removes the DisposalTracker.

> We're doing this rather than changing `DisposalTracker` to be thread-safe since we have no evidence that `DisposalTracker` was meant to be thread safe in other usage.

*I* am telling you it was meant to be thread safe. The fact that I already fixed this bug in the `main` branch _also_ tells you that it was meant to be thread safe.

> We do believe that this is a fix to the cause of the bug, but I may be misunderstanding what you mean here by cause.

It's not the case of the bug. It's just coincidentally the instance you're currently experiencing. There's no guarantee that removing this one instance won't also then push you into hitting this bug elsewhere, because this isn't the only usage of `DisposalTracker`.

> I could imagine you are suggesting we should remove the static API which causes the concurrent access.

That's not what I'm saying at all. The fact that `ExtensibilityFactory` is static is orthogonal to the fact that `DisposalTracker` isn't thread safe.

Here's what I'm saying: this is a bug, and the _**actual bug**_ should be fixed in the next v2 release. We can _also_ fix the caching bug, although I'm the only person who's ever hit it. (Still, it's a known bug that I've already fixed in `main` and we may as well fix it at the same time.) This PR, as submitted, does not fix the _**actual bug**_, it only coincidentally fixes the one instance you know of that you're currently hitting.

So that's the crux of my problem:

- This PR doesn't fix the actual bug
- This PR does fix a different bug (which is also worth fixing)
- This PR _might_ actually fix your issue, but that would be coincidental

In blunt terms: I'm not interested in just having your issue fixed. If we're shipping a new release, we're fixing the real bug, not just the symptom. I'm willing to accept the cache bug fix as well (which this PR addresses), but I am *not* interested in *only* solving *your* symptom. So if you want this issue fixed, then I will absolutely require a PR which fixes the core bug before I'm willing to accept this secondary bug fix.

> [...] every change carries with it some risk.

I agree, every change includes risk. That's why these builds will be available in pre-release form for the entire community to use and validate, including the .NET team, before an RTW fix is issued.

If you'd been led to believe that we would be accepting this one fix and then immediately issuing an RTW version of xUnit.net, then there has been some misunderstanding somewhere.

bradwilson:
If you want the smallest change for a lock, then look here: eacf43523a86a907574d085c51dbcd3944c02ff3

ericstj:
I think there was some miscommunication here.  We don't aim to propose the best/right fix yet.  We're just trying to enumerate some options, do our best to describe the risks, and work with you to find out what's the best for Xunit v2.

I apologize about my statements around `root cause` and `symptom`.  I meant to use those to contrast the first suggestion that was changing how Dispose enumerated the collection, which you also mentioned was not sufficient to fix the issue.  In the original issue you mentioned you'd like to track down where the bad usage is coming from.  We tried to do that.

In the case we happen to see, the concurrent access was from `ExtensibilityFactory.Dispose` so we wanted to propose this as a minimal option that would likely fix the issue for the cases that are hitting our CI.   The current state of the PR was just a reaction to [your suggestion](https://github.com/xunit/xunit/pull/2375#issuecomment-918350309) that removing the disposal logic is another option.  I agree this is a coincidental fix.  Let's discard it.

I'll enumerate all the options here that would likely fix this issue for us and perhaps we can decide on what's best (not just for us, but for anyone hitting variants of the bug) and how to go about making that fix.

1. Locking around `ExtensibilityFactory.disposalTracker` https://github.com/xunit/xunit/pull/2375/commits/52137f2a1da66c432fb50bab8f7519f46f5e5dc3.  Fixes the one case we've seen that's harming our CI reliability the most, but won't fix all possible cases as you mentioned. Let's **discard** this option unless there is some super-strong desire for risk aversion.
2. Port https://github.com/xunit/xunit/commit/b4ca6874dc499b20cc42efaa36d7b982ce74f7ff#diff-25ba7e6eb02a399570c746572389784ac3b5d86b428de88357a671c508d8d9e2.  Has the same problem as 1, where it will fix the case we see the most of but not all potential cases.  Let's **discard** it.
3. Port https://github.com/xunit/xunit/commit/eacf43523a86a907574d085c51dbcd3944c02ff3 and potentially some parts of https://github.com/xunit/xunit/commit/9e8d5d13fbda4224de7f996d6812855650980c1c to make it work correctly.  This should protect all concurrent access to `DisposalTracker`.  I believe this is your **preferred** fix.  If so, let us know and perhaps describe the scope of the port.

>If you'd been led to believe that we would be accepting this one fix and then immediately issuing an RTW version of xUnit.net, then there has been some misunderstanding somewhere.

Not at all.  Really we're just trying to help in the best way we can.  It's your choice to make, please let us know how we can help best or provide you with any additional information you need.

Thanks!
Eric

maryamariyan:
Closing in lieu of https://github.com/xunit/xunit/pull/2382

PR #2378, opened 2021-09-10
Update to build with VS 2022 and .NET 6

clairernovotny:
Tested locally with VS 2022 and .NET 6.0. 

Next up is getting the CI build running.


clairernovotny:
This builds locally with VS 2022 (not using the build.ps1 script).

Next step is to backport your Actions and get versioning working, but this is the first step so it builds.

PR #2381, opened 2021-09-13
Update to build with VS 2022 and .NET 6

clairernovotny:
- Support for building with VS 2022 and MSBuild 17.
- Support for reproducible builds
- Automatic versioning with NB.GV


version.json
```diff
@@ -0,0 +1,11 @@
+{
+  "version": "2.4",
+  "publicReleaseRefSpec": [
+    "^refs/heads/v2", // we release out of v2
+    "^refs/tags/v\\d+\\.\\d+", // we also release tags starting with vN.N
+    "^refs/heads/rel/v\\d+\\.\\d+" // Also release branches starting with rel/vN.N
```
bradwilson, line 6: This should change to:

`^refs/tags/v2\\.\\d+\\.\\d+`
`^refs/heads/rel/v2\\.\\d+\\.\\d+`

It not only makes it clear that these will only be used for v2 tags/branches (clarity) but also we use 3 digit versions, not 2.

version.json
```diff
@@ -0,0 +1,11 @@
+{
+  "version": "2.4",
```
bradwilson, line 2: This should be `2.4.2` right now. Also, where are you adding height?

bradwilson (changes requested):

bradwilson (approved):

Issue #2397, opened 2021-10-06
Include flowId and parseServiceMessagesInside in testStdOut message to TeamCity


andrew-at-octopus:
When running xunit tests, we found that the output was not being grouped with the test in TeamCity, this appears to be because the testStdOut message is missing the flowId.

Also we wanted to include ##teamcity service messages in the output, and have them acted upon by teamcity for the purpose of adding test metadata such as links to external information.

We have implemented this in a fork, and shown that it does meet our integration needs with TeamCity.

bradwilson:
Actually, not a dup, now that I look at your PR.

PR #2398, opened 2021-10-06
Include flowId and parseServiceMessages for stdOut

andrew-at-octopus:
When running tests in parallel, we faced two issues, the output didn't seem to match the test in TeamCity, and we were unable to include extra service messages to add metadata to the test (like a link).

Including these extra parameters allowed us to implement this.
Fixes: https://github.com/xunit/xunit/issues/2397


bradwilson:
Unfortunately this fix also requires unit test changes, and your PR branch is not open for me to make that change, so please fix the broken tests: https://github.com/xunit/xunit/runs/3915488943#step:8:146

I have ported the `tc:tags` bit to the `main` branch (it already had the missing flow ID): 2055c63a353016d933bd161e03c1851863f8cf72

bradwilson:
Thanks!

PR #2417, opened 2021-11-07
Fixes #2401 Type inference gets confused when one parameter is null

### LINKED ISSUE #2401: Type inference gets confused when one parameter is null
Consider the following code:

```cs
    public class ExampleFailure
    {
        public struct Empty<T>
        {
        }

        [Theory]
        [MemberData(nameof(GetData))]
        public void Example<T1, T2>(T1 t1, Empty<T2> t2)
        {
            
        }

        public static IEnumerable<object?[]> GetData()
        {
            yield return new object?[] {1, default(Empty<int>)};
            yield return new object?[] {"1", default(Empty<int>)};
            yield return new object?[] {null, default(Empty<int>)};
        }
    }    
```

If you run this, two of the cases will pass and one will fail, due to it incorrectly determining the type parameters of the third case to be `<object, object>` and not `<object, int>`. It feel like when it encounters a null it gives up on working the rest of the type parameters out and just hopes `object` will be okay,

SteveDunn:
Fixes #2401 Type inference gets confused when one parameter is null

### DIFF
--- a/src/xunit.v2.tests/Acceptance/Xunit2TheoryAcceptanceTests.cs
+++ b/src/xunit.v2.tests/Acceptance/Xunit2TheoryAcceptanceTests.cs
@@ -465,6 +465,29 @@ public class Xunit2TheoryAcceptanceTests
 			[Theory, MemberData("GenericData")]
 			public void GenericTest<T>(T value) { }
 		}
+
+		public class GenericParameter_Null_DoesNotImpactFollowingTypeLookup
+		{
+#nullable enable
+			public struct Empty<T> { }
+
+			[Theory]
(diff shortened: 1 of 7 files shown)


SteveDunn:
> Thanks for this!
> 
> I've opted for a different implementation after looking at your PR, including solving it for both v2 and v3. Here are my changes:
> 
> v2: [9057a3c](https://github.com/xunit/xunit/commit/9057a3ca36ca805c4d4f38981b7c15933e195432) v3: [6032446](https://github.com/xunit/xunit/commit/60324463686db1a128382bd31c9aba823afe837f)

Very welcome @bradwilson - I'll pick up another issue soon 👍

Issue #1983, opened 2019-07-16
MemberData in derived classes fails in v2.4.1 on dotnet core (NotSupportedException)


csMACnz:
I've hit this problem twice and both times figured it out without find anyone else having this issue.

When using `MemberData` in an abstract class, and using a name of a static method from a derived class throws an exception when run: `System.NotSupportedException : Specified method is not supported.`

```
// Apologies for the obfuscated example
public abstract class MyTestBase
{
    public const string ExpectedMemberDataMethodName = "MyData";

    [Theory]
    [MemberData(ExpectedMemberDataMethodName)]
    public void Test(int input)
    {
        // ... test
    }
}

public sealed class MyRealTest : MyTestBase
{
    public static IEnumerable<object[]> MyData()
        {
            yield return new object[] { 1 };
            yield return new object[] { 5 };
        }
}
```

Much digging later, and the problem is here: (deep linked into the commit from the v2.4.1 release)
https://github.com/xunit/xunit/blob/c54cc52ffb275c81afed022521870193bbca6c39/src/xunit.execution/Sdk/Reflection/ReflectionMethodInfo.cs#L70

The call is being conditionally compiled out for NetStandard1.1 lib builds.

However, this works in v2.4.0 (which has the same code) but started failing when upgrading to v2.4.1.  The conditional compilation was already there, however the target framework for the library was changed: (again using deep link specific commits comparing v2.4.0 to v2.4.1)
[1fb6b67...c54cc52#diff-e5be6f8d4c0b7123a51f447a0fd450afL10](https://github.com/xunit/xunit/compare/1fb6b672768bed98f5e7b6656c1bbb3792ed0530...c54cc52ffb275c81afed022521870193bbca6c39#diff-e5be6f8d4c0b7123a51f447a0fd450afL10)

This change affected which conditional compilation flags where used and meant that v2.4.0 used the `return Reflector.Wrap(MethodInfo.ReflectedType);` code when running on dotnet core 2.X and above, but v2.4.1 now uses `throw new NotSupportedException();` on all non-full framework platforms.

Note that the code causing this has already been fixed and the conditional compilation removed, so the next version (v3?) shouldn't have this problem anymore.

In the meantime, my workaround solution is to keep using v2.4.0 until a newer version with the fix comes out.

csMACnz:
Since the cause is already fixed in master (though not the latest release) I'll close this, and hopefully find it again next time I try to update xunit and it fails.

bedej:
[...]

PR #2419, opened 2021-11-16
Support runtimes where ReflectedType exists such as NetStandard2.0 compatible runtimes

csMACnz:
This relates to issue #1983

The calling code that fails is this particular call in DataDiscoverer:
https://github.com/xunit/xunit/blob/4db2e39c72e70a3d1ffd308a3131a397e39ecffc/src/xunit.core/Sdk/DataDiscoverer.cs#L32

This change should restore the functionality that worked on NetStandard 2.0 compatible runtimes with v2.4.0 that breaks when upgrading to 2.4.1 (where NetStandard2.0 target was removed breaking this particular call.

I think the missing unit tests are possible but still working out where they go into this file. Potentially alongside this test case?
https://github.com/xunit/xunit/blob/4db2e39c72e70a3d1ffd308a3131a397e39ecffc/test/test.xunit.execution/Acceptance/Xunit2TheoryAcceptanceTests.cs#L869-L888
EDIT: After some playing around I've realised none of these theory tests light up for non .Net Full Framework anyway.

I've tried to do this as low impact as possible.

I investigated just using DeclaringType but there is a subtle difference, which is exactly what the DataDiscoverer is using it for.
(DeclaringType would resolve to the base class while ReflectedType will correctly use the derived test class type, [according to Stack Overflow](https://stackoverflow.com/a/21884304))

Thanks in Advance. All feedback appreciated

### DIFF
--- a/src/xunit.execution/Sdk/Reflection/ReflectionMethodInfo.cs
+++ b/src/xunit.execution/Sdk/Reflection/ReflectionMethodInfo.cs
@@ -68,7 +68,19 @@ namespace Xunit.Sdk
         public ITypeInfo Type
         {
 #if NETSTANDARD1_1
-            get { throw new NotSupportedException(); }
+            get
+            {
+                var reflectedTypeProperty = MethodInfo.GetType().GetRuntimeProperty("ReflectedType");
+                if (reflectedTypeProperty != null)
+                {
+                    var methodInfoReflectedType = (Type)reflectedTypeProperty.GetValue(MethodInfo, null);
+                    return Reflector.Wrap(methodInfoReflectedType);
+                }
+                else
+                {
+                    throw new NotSupportedException();
+                }
+            }
 #else
             get { return Reflector.Wrap(MethodInfo.ReflectedType); }
 #endif


bradwilson:
Obviously we would want some tests here. I'm fine with the approach of dynamically looking for `ReflectedType`, though it should be done in a cached lookup since these wrappers get used a lot.

csMACnz:
Time to actually tackle the tests on this has stalled.

I've found a quick workaround using an alternate DataAttribute custom class though:

```

    internal class BugFixMemberDataAttribute : MemberDataAttributeBase
    {
        public BugFixMemberDataAttribute(string memberName, params object[] parameters)
            : base(memberName, parameters)
        {
        }

        public override IEnumerable<object[]?>? GetData(MethodInfo testMethod)
        {
            MemberType = testMethod.ReflectedType;
            return base.GetData(testMethod);
        }

        protected override object[]? ConvertDataItem(MethodInfo testMethod, object item)
        {
            if (item == null)
            {
                return null;
            }

            return (item as object[]) ?? throw new ArgumentException($"Property {MemberName} on {MemberType ?? testMethod.DeclaringType} yielded an item that is not an object[]");
        }
    }
```

YMMV

bradwilson:
I wrote the acceptance test for this. I would've pushed the changes into your branch and merged, but you wrote them directly into your `v2` branch, which I didn't want to clutter up with my stuff (since it would've required a force push). So your changes are merged here: e813b3197c154f109c838eb0580d684b47a87a36

Thanks!

Issue #723, opened 2015-12-27
Console runner does not load configuration when using config xunit.appDomain = 'denied'


TheCloudlessSky:
When `xunit.appDomain` is set to `denied` in the configuration, the app's configuration is not available for use with `ConfigurationManager` when using the console runner. However, running with TestDriven.Net, it works as expected.

## Steps to reproduce

1. Create a .NET 4.5 Class Library project (I called it XunitAppDomainConfig).
2. Install xunit and the console runner:

        > install-package xunit -IncludePrerelease
        > install-package xunit.runner.console -IncludePrerelease

3. Create `SampleTests.cs`:

    ```csharp
    public class SampleTests
    {
        [Fact]
        void should_find_correct_connection_string()
        {
            var cs = ConfigurationManager.ConnectionStrings["hello_world"];
            Assert.NotNull(cs);
            Assert.Equal("Hello World", cs.ConnectionString);
        }
    }
    ````

4. Create an `App.config` with:

    ```xml
    <?xml version="1.0" encoding="utf-8" ?>
    <configuration>
      <appSettings>
        <add key="xunit.appDomain" value="denied" />
      </appSettings>
      <connectionStrings>
        <add name="hello_world" connectionString="Hello World" />
      </connectionStrings>
    </configuration>
    ```

5. Run tests with TestDriven.NET (passes):

        ------ Test started: Assembly: XunitAppDomainConfig.dll ------

        1 passed, 0 failed, 0 skipped, took 0.38 seconds (xUnit.net 2.2.0 build 3239).

6. Run tests with Console runner (fails):

        > packages\xunit.runner.console.2.2.0-beta1-build3239\tools\xunit.console.exe XunitAppDomainConfig\bin\Debug\XunitAppDomainConfig.dll -verbose -diagnostics

        xUnit.net Console Runner (64-bit .NET 4.0.30319.34209)
          Discovering: XunitAppDomainConfig (app domain = off, method display = ClassAndMethod)
          Discovered:  XunitAppDomainConfig (running 1 test case)
          Starting:    XunitAppDomainConfig (parallel test collections = on, max threads = 4)
            XunitAppDomainConfig.SampleTests.should_find_correct_connection_string
            XunitAppDomainConfig.SampleTests.should_find_correct_connection_string [FAIL]
              Assert.NotNull() Failure
              Stack Trace:
[...]

PR #2420, opened 2021-11-18
Enable ConfigurationManager on .NET 6

### LINKED ISSUE #1653: ConnectionStrings not loading correctly
This is a strange issue. My ``System.Configuration.ConfigurationManager.ConnectionStrings`` are not loading correctly. I think they are possibly *"loading"* (so-called) from a ``machine.config`` I have someplace on the machine. How I know this is because the configuration that is there is by the key ``"LocalSqlServer"``. However, my configuration should reflect this:

```Xml
<connectionStrings>
    <add name="MasterConnectionString" connectionString="Server=localhost;Database=master;Integrated Security=true" />
</connectionStrings>
```

However, this ``name`` does not exist, returns ``null``. I've tried adding a ``<clear/>`` line, but that does not work, either.

Methinks this is possibly an ``Xunit`` runner issue not loading the assembly correctly.

Which, incidentally, this was working prior to migration to ``Xunit``, from ``NUnit``. I'm not just sure quite what the issue is.

### LINKED ISSUE #1689: app.config settings not read in netcore xunit test project
Hello all,

I have a test project created, and added the following package:

- System.Configuration.ConfigurationManager : 4.4.1
- xunit : 2.3.1
- xunit.runner : 2.3.1

Then I have added an app.config file with the following content:
```
<?xml version="1.0" encoding="utf-8" ?>
<configuration>
  <appSettings>
    <add key="configuredSecretKeyName" value="configuredKeyName" />
  </appSettings>
</configuration>
```
But when executing the following test, it seems the app.config settings are not being read?
```
[Fact]
public void TestAppSettings()
{
    var configValue = ConfigurationManager.AppSettings["configuredSecretKeyName"];
    Assert.NotNull(configValue);
}
```
Any idea?

bricelam:
This has finally been unblocked by dotnet/runtime#56748. This will enable testing application code that uses ConfigurationManager. Just like on .NET Framework, it will read from the App.config file of the test project.

Fixes #1653, fixes #1689

### DIFF
--- a/src/xunit.v3.runner.utility/AppDomain/AppDomainManager_NoAppDomain.cs
+++ b/src/xunit.v3.runner.utility/AppDomain/AppDomainManager_NoAppDomain.cs
@@ -1,4 +1,5 @@
 using System;
+using System.IO;
 using System.Reflection;
 using System.Runtime.ExceptionServices;
 using Xunit.Internal;
@@ -7,6 +8,24 @@ namespace Xunit
 {
 	class AppDomainManager_NoAppDomain : IAppDomainManager
 	{
+		public AppDomainManager_NoAppDomain(
+			string assemblyFileName,
+			string? configFileName)
+		{
+			Guard.ArgumentNotNullOrEmpty(assemblyFileName);
+
+			assemblyFileName = Path.GetFullPath(assemblyFileName);
+			Guard.FileExists(assemblyFileName);
+
+			if (configFileName == null)
+				configFileName = GetDefaultConfigFile(assemblyFileName);
+
+			if (configFileName != null)
+				configFileName = Path.GetFullPath(configFileName);
+
+			AppDomain.CurrentDomain.SetData("APP_CONFIG_FILE", configFileName);
+		}
+
 		public bool HasAppDomain => false;
 
 		public TObject? CreateObject<TObject>(
@@ -65,5 +84,15 @@ namespace Xunit
 
 		public void Dispose()
 		{ }
+
+		static string? GetDefaultConfigFile(string assemblyFile)
+		{
+			var configFilename = assemblyFile + ".config";
+
+			if (File.Exists(configFilename))
+				return configFilename;
+
+			return null;
+		}
 	}
 }
--- a/src/xunit.v3.runner.utility/AppDomain/AppDomainManagerFactory.cs
+++ b/src/xunit.v3.runner.utility/AppDomain/AppDomainManagerFactory.cs
@@ -21,7 +21,7 @@ namespace Xunit
 				return new AppDomainManager_AppDomain(assemblyFileName, configFileName, shadowCopy, shadowCopyFolder, diagnosticMessageSink);
 #endif
 
-			return new AppDomainManager_NoAppDomain();
+			return new AppDomainManager_NoAppDomain(assemblyFileName, configFileName);
 		}
 	}
 }


bricelam:
This could also fix #723 if you want to add [some magic reflection code](https://github.com/dotnet/efcore/blob/v5.0.0/src/ef/Commands/ProjectCommandBase.cs#L76-L85) to make it take effect on .NET Framework.

bradwilson:
We are deprecating `App.config` in v3, which makes this PR irrelevant from my perspective. Do you have a reason we shouldn't?

bricelam:
This isn't about xUnit.net's settings inside App.config. This is about testing application/library code that uses ConfigurationManager.

bricelam:
For example, if I have application code that calls `ConfigurationManager.ConnectionStrings["Northwind"]`, that code will fail in the tests since it will try to read from a non-existing `xunit.console.dll.config` file instead of the test project's App.config file.

bricelam:
In other words, the `ConfigurationManager` API *isn't* deprecated in .NET 6 regardless of whether you choose to deprecate the xunit.* keys inside the appSettings section.

bricelam:
This is important for existing .NET Framework (WinForms, EF6, etc.) projects that want to migrate to .NET 6.

Currently, after changing their target framework, all of their tests that rely on App.config (again, we're not talking about their xUnit.net settings here) start failing, and there is no obvious reason why or how to fix it because they "just worked" on .NET Framework because xUnit.net was setting AppDomainSetup.ConfigurationFile.

bradwilson:
The configuration file that's being passed into that method is the xUnit.net configuration file, not `App.config`.

bricelam:
I'm pretty sure it's the App.config of the user's unit test project that gets copied to the output directory by MSBuild.

bricelam:
Here's a repro project to demonstrate the problem: [TestProject1.zip](https://github.com/xunit/xunit/files/7725305/TestProject1.zip)

The test passes on .NET Framework and fails on .NET 6.

Uncommenting the code that mirrors the code in this PR makes it pass on both.

bradwilson:
Why is this xUnit.net's responsibility? This seems like either the responsibility of .NET, or the developer who wants to enable some feature that requires additional code. We are allowing users to define their own Main() method with v3, so this code could absolutely go there.

What other "features" that .NET doesn't properly implement will we also be responsible for fixing?

bricelam:
> Why is this xUnit.net's responsibility?

Because, to users, this is a breaking change in xUnit.net between .NET Framework and .NET 6. This code already exists in xUnit.net. It just wasn't ported it .NET Core.

> ...to enable some feature that requires additional code.

To the user, this *didn't* require any additional code on .NET Framework because xUnit.net *already* took on the responsibility.

> What other "features" that .NET doesn't properly implement will we also be responsible for fixing?

None. This is the biggest point of feedback we've seen from our users trying to port from .NET Framework. It's very hard for them to understand why it suddenly stops working when they port to .NET 6. Please consider the feedback of your own users in #1653 and #1689.

> ...allowing users to define their own Main() method...

If the user's test assembly is an exe, this problem goes away. But that doesn't really help the fundamental problem I'm trying to address here: to help users porting from .NET Framework. They still won't understand why it suddenly stops working and won't be any more likely to find this additional workaround.

cc @JeremyLikness

src/xunit.v3.runner.utility/AppDomain/AppDomainManager_NoAppDomain.cs
```diff
@@ -7,6 +8,24 @@ namespace Xunit
...
+			Guard.ArgumentNotNullOrEmpty(assemblyFileName);
+
+			assemblyFileName = Path.GetFullPath(assemblyFileName);
+			Guard.FileExists(assemblyFileName);
+
+			if (configFileName == null)
+				configFileName = GetDefaultConfigFile(assemblyFileName);
+
+			if (configFileName != null)
+				configFileName = Path.GetFullPath(configFileName);
```
bricelam, line 24: Copied from here:

https://github.com/xunit/xunit/blob/19fc844daabe9ae02f350340d49a4319ac6c2ed3/src/xunit.v3.runner.utility/AppDomain/AppDomainManager_AppDomain.cs#L26-L35

src/xunit.v3.runner.utility/AppDomain/AppDomainManager_NoAppDomain.cs
```diff
@@ -7,6 +8,24 @@ namespace Xunit
...
+
+			AppDomain.CurrentDomain.SetData("APP_CONFIG_FILE", configFileName);
```
bricelam, line 26: .NET 6 equivalent of this existing code:

https://github.com/xunit/xunit/blob/19fc844daabe9ae02f350340d49a4319ac6c2ed3/src/xunit.v3.runner.utility/AppDomain/AppDomainManager_AppDomain.cs#L70

src/xunit.v3.runner.utility/AppDomain/AppDomainManager_NoAppDomain.cs
```diff
@@ -65,5 +84,15 @@ class AppDomainManager_NoAppDomain : IAppDomainManager
...
+
+		static string? GetDefaultConfigFile(string assemblyFile)
+		{
+			var configFilename = assemblyFile + ".config";
+
+			if (File.Exists(configFilename))
+				return configFilename;
+
+			return null;
+		}
```
bricelam, line 96: Copied from here:

https://github.com/xunit/xunit/blob/19fc844daabe9ae02f350340d49a4319ac6c2ed3/src/xunit.v3.runner.utility/AppDomain/AppDomainManager_AppDomain.cs#L167-L175

src/xunit.v3.runner.utility/AppDomain/AppDomainManager_NoAppDomain.cs
bradwilson: This is here because we fundamentally broke the executable model in v1/v2 (like everybody else) by declaring that unit tests should be libraries instead of executables. This has changed in v3.

bradwilson:
To reiterate: We have no intention of supporting `App.config` at all in v3, under any scenario.

If you want to use `App.config`, and additional code is required, then you'll be responsible for that code.

bradwilson:
Out of curiosity: Why isn't VSTest fixing this? They have all the information necessary (target framework, test assembly name, etc.)

bradwilson:
Actually, nothing needs to be done for v3 at all:

*csproj*

```xml
<Project Sdk="Microsoft.NET.Sdk">

  <PropertyGroup>
    <TargetFramework>net6.0</TargetFramework>
  </PropertyGroup>

  <ItemGroup>
    <PackageReference Include="System.Configuration.ConfigurationManager" Version="6.0.0" />
    <PackageReference Include="xunit.v3" Version="0.1.1-*" />
  </ItemGroup>

</Project>
```

*App.config*

```xml
<?xml version="1.0" encoding="utf-8" ?>
<configuration>
  <appSettings>
    <add key="not-xunit" value="test-value" />
  </appSettings>
</configuration>
```

*cs*

```csharp
using System;
using System.Configuration;
using Xunit;

namespace TestProject1
{
    public class UnitTest1
    {
        [Fact]
        public void Test1()
        {
            Console.WriteLine(ConfigurationManager.AppSettings["not-xunit"]);
            Assert.Equal("test-value", ConfigurationManager.AppSettings["not-xunit"]);
        }
    }
}
```

*Result*

image

bricelam:
Do what you want. I'm done advocating for users on this.

bradwilson:
Allow me to put some clarity into a single post, because I think there has been some confusion about what's trying to be supported here and what's needed to make it work.

## Supporting v2 test projects

Running a v2 test project can be done in two ways: with the v2 `xunit.runner.utility.*.dll` family, and with the v3 `xunit.v3.runner.utility.*.dll` family.

### Running via v2 runner utility

The v2 runner utility targets `net35;net452;netstandard1.1;netstandard1.5;netcoreapp1.0;uap10.0`. Of these, the one of interest is `netcoreapp1.0`, which is what is used today to run .NET 6.0 projects.

The required API (`AppDomain.CurrentDomain.SetData`) is not available because `AppDomain` does not exist in .NET Core 1.0. It wasn't added until .NET Core 2.0; it appears that SetData was part of that addition, but it's unknown to me/untested whether the API is implemented in such a way that would actually work on .NET 6.0.

All that said, adding additional target frameworks to v2 runner utility is not something to be undertaken lightly. We broke a lot of things trying to add support for `netstandard2.0` back in the 2.4.0 days, and rolling that back is what caused 2.4.1 to come into being. **I consider this a showstopper.**

### Running via v3 runner utility

The v3 runner utility targets `net472;netstandard2.0`. Of these, the one of interest is `netstandard2.0`, which is what would be used to run v2 .NET Core tests with a v3 runner. The `AppDomain.CurrentDomain.SetData` API is available here. The code you've added (once fixed to use the correct configuration file) would in theory be sufficient.

However, v2 .NET Core tests are run via the 3rd party `xunit.runner.visualstudio` library today (the bridge between xUnit.net and VSTest). The current version of `xunit.runner.visualstudio` is linked against the v2 runner utility library.

We could ship an experimental version of this runner, linked against the v3 runner APIs, to validate that v2 tests are then able to take advantage of the .NET 6.0 configuration file feature. However, the new version xUnit.net VSTest plugin would by definition be reducing the supported target frameworks that it could run, and has not been tested yet to understand what currently supported platforms might disappear as a result. A couple examples:

- xUnit.net v2 has a minimum .NET Framework version of 4.5.2. Would the VSTest plugin being compiled against `net472` mean that v2 tests must be upgraded to 4.7.2 to run?
- xUnit.net v2 has a minimum .NET Core version of 1.0. Would the VSTest plugin being compiled against `netstandard2.0` mean that v2 tests must be upgraded to .NET Core 2.0 to run?
- xUnit.net v2 supports UWP (v3 currently does not explicitly support UWP at all). Would the VSTest plugin be able to run v2 UWP tests at all? If so, would v2 tests need to upgraded to UWP 10.0.16299 to run?

Answers to these questions are needed to understand whether we would be able to ship a new version of `xunit.runner.visualstudio` against the v3 runner and expect all supported scenarios today to continue to be supported, or whether we'd need a new NuGet package (probably `xunit.v3.runner.visualstudio`) to allow users to choose whether they want v3 support or want to preserve older (admittedly deprecated) versions of .NET Framework, .NET Core, and UWP to continue to work.

## Supporting v3 test projects

v3 tests, [as previously illustrated](#issuecomment-997319603), already work without changes, because of the change in architecture. v3 projects are stand-alone executables. .NET 6.0 already supports this today without any code required by the developer or by xUnit.net.

bradwilson:
It's also worth pointing out that I believe the .NET team uses our unsupported .NET Core console runner rather than the VSTest plugin. This is no longer shipping as part of v3.

If you needed to patch this runner to support .NET 6.0, you could fork from the v2 branch and update the .NET Core console runner to add a `net6.0` target, and push your `AppDomain.Current.SetData` code into there conditionally, rather than trying to rely on something that can't be back-patched into the v2 runner utility DLLs.

PR #2434, opened 2021-12-03
Emit diagnostic message when unable to load config

### LINKED ISSUE #1655: Add diagnostic message when JSON file is malformed
Related: #1654

skimmedsquare:
* Adds proxy for ConfigReader to enable mock injection
* Modifies xunit to emit diagnostic message per assembly

Fixes #1655

```
Before you submit a PR...

  https://github.com/xunit/xunit/issuesq=is%3Aopen+is%3Aissue+label%3A%22help+wanted%22)? Yes
* Did you read the project governance @ https://xunit.net/governance ? Yes
* Does the code follow existing coding styles? (tabs, comments, no regions, etc.) Yes
* Did you write unit tests? Yes
```

### DIFF
--- a/src/xunit.v3.runner.msbuild.tests/xunitTests.cs
+++ b/src/xunit.v3.runner.msbuild.tests/xunitTests.cs
@@ -81,6 +81,42 @@ public class xunitTests
 
 			Assert.False(result);
 		}
+
+		[Fact]
+		public static void ReportsWhenUnableToReadConfig()
+		{
+			static bool configReader(
+				TestAssemblyConfiguration configuration,
+				string? assemblyFileName,
+				string? configFileName = null) =>
+					false;
+
+			var messageSink = SpyMessageSink.Capture();
+			var reporter = Substitute.For<IRunnerReporter>();
+			reporter
+				.IsEnvironmentallyEnabled
+				.Returns(true);
+			reporter
+				.CreateMessageHandler(Arg.Any<IRunnerLogger>(), Arg.Any<_IMessageSink>())
+				.Returns(messageSink);
+			var stubAssembly = Substitute.For<ITaskItem>();
+			stubAssembly
+				.GetMetadata("ConfigFile")
+				.Returns("DummyPath");
+			var assemblies = new ITaskItem[] { stubAssembly };
+			var xunit = new Testable_xunit(configReader, reporter)
+			{
+				Assemblies = assemblies,
+				NoLogo = true
+			};
+
+			xunit.Cancel();
+			xunit.Execute();
+
+			var msg = Assert.Single(messageSink.Messages);
+			var diagnosticMessage = Assert.IsType<_DiagnosticMessage>(msg);
+			Assert.Equal("Unable to read 'DummyPath'; falling back to default values.", diagnosticMessage.Message);
+		}
 	}
 
 	public class GetReporter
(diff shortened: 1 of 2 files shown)


dnfadmin:
[CLA assistant check](https://cla.dotnetfoundation.org/xunit/xunit?pullRequest=2434) <br/>Thank you for your submission, we really appreciate it. Like many open source projects, we ask that you sign our [Contributor License Agreement](https://cla.dotnetfoundation.org/xunit/xunit?pullRequest=2434) before we can accept your contribution.<br/><br/>:x: skimmedsquare [sign now](https://cla.dotnetfoundation.org/xunit/xunit?pullRequest=2434)<br/><sub>You have signed the CLA already but the status is still pending? Let us [recheck](https://cla.dotnetfoundation.org/check/xunit/xunit?pullRequest=2434) it.</sub>

skimmedsquare:
Force pushed, missed committing a final change...

bradwilson:
Thanks! I rebased this against `main` and restructure it to match the existing project style.

bradwilson:
I'm looking over this code now (after fixing the merge & style issues) and I'm not convinced that it's in the right place (or complete) yet. Some thoughts:

1. This code updates the MSBuild runner, but no other runners.

2. It's likely that this code in the wrong place anyway, because it attempts to send a diagnostic message to the runner reporter message handler, but those message handlers are never sent diagnostic messages. Instead, diagnostic messages are expected to be special handlers that are constructed by the runners themselves. There exists an issue of ordering: your code is in `Execute()` which is still loading the configuration and determining what to run, but the code that creates the diagnostic message sink exists in `ExecuteAssembly`, which is called once per test assembly (and long after the configuration has been loaded).

3. This code reports on a very generic return of `false` from `ConfigReader.Load()` with a message of `Unable to read '{filename}'; falling back to default values.` which, while is better than what we have today (silent failure), leaves the developer slightly unsure of exactly what the problem is that was encountered. For example, if I look in `ConfigReader_Json.LoadFile`, we could return `false` if the file doesn't exist, if `LoadJson` returns `false`, or if there is an exception thrown. Similarly, `LoadJson` might return `false` if the JSON top level isn't an object, or if there is an exception thrown. This leaves a lot of possibilities here for the developer to try to understand without any help from us.

I suspect that the correct design might be that we change the signature of `ConfigReader.Load` to return a nullable string, which will return `null` if the load was successful, or a string message to indicate what kind of load failure it was. Then the individual runners could be updated to report that message as they see fit.

I'd like to think on this a bit more, though, to be sure.

skimmedsquare:
Thanks for the feedback!

> This code updates the MSBuild runner, but no other runners.

Ah, makes sense. To be clear, the full list of runners / files that would need changes are `CommandLine.cs`, `TdNetRunnerHelper.cs`, and `AssemblyReader.cs`, correct?

> It's likely that this code in the wrong place anyway...

https://github.com/xunit/xunit/blob/83a532d7373d1e2462a68caba5576a9d1c8641e3/src/xunit.v3.runner.msbuild/xunit.cs#L321

This is the sink you're talking about right? Wasn't aware of the constraints related to the different message handlers. This does make a clean solution a bit odd. My initial thought now would be to maybe modify the `XunitProjectAssembly` class to have a `string` field representing config status so that these issues can be reported on at the right time.

> This code reports on a very generic return of `false` from `ConfigReader.Load()`

Fair point. At the very least it should probably disambiguate between errors that can have fairly specific error messages vs a general catch all "something went wrong". 

I'm unsure if I agree that returning a `string?` is the best choice here, that seems like a bit of a code smell compared to throwing an exception. This is based on a fairly shallow understanding of the code flow, but it seems like it would be a bit messy to need to modify multiple layers to return this string diagnostic message and possibly hinder refactoring in the future.

bradwilson:
Closing in favor of #2864. Thanks for the PR!

Issue #2402, opened 2021-10-15
Assert.Equal<T>(IEnumerable<T>, IEnumerable<T>) enumerates multiple times


ericstj:
Regression introduced here:
https://github.com/xunit/assert.xunit/commit/bb5d19df7322627f649e55e8006157cd0af31ac6#diff-5ddb02071f63a19de4084075b7e4e26c92234035da84fc639d197f2b2ce1dd4cR65

It looks to me like this should return rather than fall through to calling compare.Equals again.  It might even be further simplified as this: https://github.com/ericstj/assert.xunit/commit/c195a150063def12a8923970a1937fa1020baecb

Here's a repro for the failure
[xunitEnum.zip](https://github.com/xunit/xunit/files/7350443/xunitEnum.zip)

bradwilson:
I agree that the code should include a `return` when `aec != null` but the compare succeeded.

However, I think there's another issue with double enumeration, and that's in failure, because `EqualException` is going to re-enumerate again as well when given an enumerable (which ends up being passed to `ArgumentFormatter`). That means the only reasonable solution might be to pre-enumerate (with something like `.ToList()` and then pass the pre-enumerated list to everything, rather than the raw `IEnumerable`.

bradwilson:
That said, your fix isn't entirely appropriate, because our assertion library is available via source, which means it must compile against C# 6 without nullable annotations, so for example you can't convert to inline `out var` as that will break backward language compatibility. All assertion changes have to go through both the v2 and v3 pipelines, since I'm not differentiating between the two, and the v3 source includes a project that sets up for C# 6, so this would be caught as soon as you tried to bring in your changes (which you'd need to do in order to PR in the associated unit tests). https://github.com/xunit/xunit/tree/main/src/xunit.v3.assert.NonNullable

bradwilson:
This is now available on MyGet:

v2: [`2.4.2-pre.22`](https://myget.org/feed/xunit/package/nuget/xunit/2.4.2-pre.22)
v3: [`0.1.1-pre.28`](https://myget.org/feed/xunit/package/nuget/xunit.v3/0.1.1-pre.28)

bradwilson:
(I am going to close this, but please note that my [comment about the 2nd enumeration from `EqualException`](https://github.com/xunit/xunit/issues/2402#issuecomment-944970201) is still un-resolved.)

ericstj:
[...]

PR #2463, opened 2022-01-13
Test that Equals only enumerates once

ericstj:
Test for https://github.com/xunit/assert.xunit/pull/48

Issue https://github.com/xunit/xunit/issues/2402

Ensure that Assert.Equals on IEnumerables will only enumerate once.


bradwilson:
Thanks!

PR #2465, opened 2022-01-14
Fix "affect change" -> "effect change"

rayjolt:
"Affect change" is a common misspelling of "effect change".
See https://writingexplained.org/affect-change-or-effect-change


rayjolt:
This could probably do with a clearer commit title. Will repost in a second.

PR #2466, opened 2022-01-14
Docs: fix typo in "Why did we build XUnit 1.0" page

rayjolt:
Fix "change that we wanted to affect" to "change that we wanted to
effect". "Affect change" is a common misspelling of "effect change". [1]

[1] https://writingexplained.org/affect-change-or-effect-change


rayjolt:
On the [governance page](https://xunit.net/governance) I see that I need to sign an assignment agreement, but I couldn't find where it is. Did I overlook it?

bradwilson:
This change isn't large enough to warrant an assignment agreement, so the bot didn't bother to request one.

bradwilson:
Thanks!

Issue #2485, opened 2022-02-22
Assert.Equal / Assert.NotEqual is comparing IEnumerable of actual equality


trivalik:
If you wanna compare two objects which implement IEnumerable (or similar). The Assert.NotEqual/Assert.Equal is just comparing the inner list. The IEnumerable cast could only be used to find difference but not to decide it is equal.

Example in which Name compare should fail:
```
        public class ComponentViewModel : IEnumerable<KeyValuePair<string, object>>
        {
            private readonly IDictionary<string, object> _decorationValues = new Dictionary<string, object>();

            public string Name { get; set; }

            public IEnumerator<KeyValuePair<string, object>> GetEnumerator()
            {
                return _decorationValues.GetEnumerator();
            }

            IEnumerator IEnumerable.GetEnumerator()
            {
                return GetEnumerator();
            }

            public override bool Equals(object obj)
            {
                // is never called
                return obj != null &&
                    Name== obj.Name;
            }
        }

        [Fact]
        public void Equal_AssertEqual_NotEqual()
        {
            var first = new ComponentViewModel {Name = "1"};
            var second = new ComponentViewModel {Name = "2"};

            Assert.NotEqual(first, second);
        }
```

bartelink:
You'll get quicker and better answers if you post this on stackoverflow.com; there are many more eyes on it there.
I personally would read the code and/or the xmldocs for `class Assert` in the Xunit source code if I was trying to solve your problem...

trivalik:
As far as I know is there no documentation, as other tickets stated already. The xmldoc https://github.com/xunit/assert.xunit/blob/main/EqualityAsserts.cs#L53 shows that the magic behind is not honored in documentation.

bartelink:
Is that the overload you're calling (maybe there is an `IEnumerable`-specific one that's winning, in which case adding a `<ComponentViewModel >` might make it work?)

My point about the documentation was that the xmldoc is generally complete and accurate and/or takes PRs.

Either way my point stands that this can typically be sorted out on SO in minutes so that's generally a better avenue of approach (For instance, I use [unquote](https://github.com/SwensenSoftware/unquote/) for my assertions so don't know/care about the answer to this question), notwithstanding the fact that you are showing a great can-do attitude with failing tests attitude (which is definitely not typically what happens when people log issues that smell like usage questions to bystanders with a little knowledge like me ;) )

trivalik:
[...]

PR #2488, opened 2022-02-22
add test for IEnumerable with additional properties

trivalik:
Test for change https://github.com/xunit/xunit/issues/2485, I was not able to compile neither run tests.


bradwilson:
This will be covered by #2612 when it gets addressed. Thanks for the example test!

PR #2519, opened 2022-04-27
Display the types of unequal items in collections

mongeon:
This is for #2458.

PR #2525, opened 2022-05-13
2352 improve output for assert.equal

### LINKED ISSUE #2352: Improve output for Assert.Equal for collections of mis-matched sizes/underrun
When comparing long arrays, and in particular arrays of bytes (as may be necessary when testing binary operations, such as hashing functions, serializing, etc), xunit does not provide descriptive messages.

For example, when running this test:
```cs
Assert.Equal(new byte[] { 1, 2, 3, 4, 5, 6, 7 }, new byte[] { 1, 2, 3, 4, 5, 6, 7, 8 });
```
The output is:
```
Xunit.Sdk.EqualException
Assert.Equal() Failure
Expected: Byte[] [1, 2, 3, 4, 5, ...]
Actual:   Byte[] [1, 2, 3, 4, 5, ...]
```

and when running this test:
```cs
Assert.Equal(new byte[] { 1, 2, 3, 4, 5, 6, 7, 8 }, new byte[] { 1, 2, 3, 4, 5, 8, 7, 6 });
```
The output is the same.

These are very different failures that are being reported to the programmer as being an identical failure.

In particular, these features are missing:
1. Checking the size of the arrays
2. Reporting the position at which differences begin

An example error message that would be sufficient for each feature:
1#
```
Xunit.Sdk.EqualException
Assert.Equal() Failure
Expected: Byte[7] [1, 2, 3, 4, 5, ...]
Actual:   Byte[8] [1, 2, 3, 4, 5, ...]
```
Note the size of the arrays is reported in the array brackets - feel free to report the size in some other manner, I just thought this was fitting.

2#
```
Xunit.Sdk.EqualException
Assert.Equal() Failure at index 5
Expected: Byte[8] [..., 6, 7, 8]
Actual:   Byte[8] [..., 8, 7, 6]
```
If this index happens to be in the middle of the array, there should be ellipsis at either size of the array contents. You could show up to 5 or 10 elements to help the programmer debug the issue without having to start a debugger.

mongeon:
This PR is related to the #2519 PR (Issue #2485)  as we provide array types from there. 

Fix #2352

### DIFF
--- a/src/xunit.v3.assert.tests/Asserts/Sdk/Exceptions/EqualExceptionTests.cs
+++ b/src/xunit.v3.assert.tests/Asserts/Sdk/Exceptions/EqualExceptionTests.cs
@@ -300,5 +300,26 @@ public class EqualExceptionTests
 			Assert.NotNull(ex);
 			Assert.Equal(expectedMessage, ex.Message);
 		}
+
+		[Fact]
+		public void Array_ExpectedSameType()
+		{
+			var expectedMessage =
+				"Assert.Equal() Failure" + Environment.NewLine +
+				"                 ↓ (pos 2, type System.Int32)" + Environment.NewLine +
+				"Expected: [1, 2, 3]".Replace("'", "\"") + Environment.NewLine +
+				"Actual:   [1, 2, 3]".Replace("'", "\"") + Environment.NewLine +
+				"                 ↑ (pos 2, type System.Int64)";
+
+			var ex = Record.Exception(
+				() => Assert.Equal(
+					new object[] { 1, 2, 3 },
+					new object[] { 1, 2, 3L }
+				)
+			);
+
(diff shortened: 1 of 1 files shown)


mongeon:
> What’s the status and plan on this?
> 
> Looks like the changes currently only change the tests, but not the actual functionality/output?

@ite-klass I've changed the tests to make them fails with the new expected output of xunit tests results. I've started working on the fix in the https://github.com/xunit/assert.xunit repo, which is a submodule in this repository, but didn't succeed with the fix and got less time to work on this at the moment. If you like to get a look, please do so.

ite-klass:
I see. PR https://github.com/xunit/assert.xunit/pull/50

bradwilson:
Closing as it seems this was already covered by xunit/assert.xunit#50. Thanks!

PR #2577, opened 2022-08-23
Fix link for v3 alpha docs

danielchalmers:
The link worked if you were on https://xunit.net, but clicking it from the README on GitHub would take you to the non-existent https://github.com/xunit/xunit/blob/main/docs/v3-alpha


bradwilson:
Thanks!

PR #2579, opened 2022-08-28
Add async-await to the suggested fix in xUnit2014 analyzer rule documentation

vers-one:
While the suggested fix is correct (it replaces `Assert.Throws` with `Assert.ThrowsAsync` to fix the violation), it also renders this assert useless (the suggested test will give false positives if the type of the actual exception doesn't match the type of the expected exception).


bradwilson:
Oops! Nice catch! Thanks. :)

PR #2641, opened 2023-01-06
 TimeSpan, BigInteger, TimeOnly, DateOnly support for v2

### LINKED ISSUE #2640: More know types for serialization
I would like to add support for `TimeSpan`, `BigInteger`, `DateOnly` and `TimeOnly` for `XunitSerializationInfo.supportedSerializationTypes`.
The latter two require additional net6 TFM. Would you like to accept PR?

sungam3r:
fixes #2640


sungam3r:
Now it should work better.

bradwilson:
Note that we cannot support `BigInteger` on .NET Framework because it comes from an assembly in the GAC, though it is supported on all other platforms. `TimeSpan` is supported on all platforms. I updated the test project and test runner to be able to run tests under .NET 6 so I could add tests for the support for `DateOnly` and `TimeOnly`.

PR #2643, opened 2023-01-10
Add support for TimeSpan, BigInteger, TimeOnly, DateOnly

sungam3r:
fixes https://github.com/xunit/xunit/issues/2640 for v3

version for v2 here: https://github.com/xunit/xunit/pull/2641


sungam3r:
Someone should approve and run tests. They should fail now. I need to watch logs to fix.

sungam3r:
Now it should work better.

bradwilson:
Thanks!

The tests you provided were broken, because of some bad nullable ignores (on .NET Framework), which I fixed. In addition, I added some more tests/test data, and used a more precise/quicker serialization for TimeOnly and DateOnly (something I'll probably back-fill into the rest of the types).

I have not looked at your v2 PR yet, because I'm not convinced that there will be another shipping v2 version. I'm going to leave it open for the time being, but it will likely need to be fixed up with the same changes I put in here. I'm happy to do that if there is another shipping v2 version, so no need for you to do anything at this point.

sungam3r:
OK. Honestly, I have no plans to switch to v3, I just wanted to force some tests with BigInts, Dates, etc. to work as intended on v2. What is the degree of maturity of v3?

bradwilson:
It is stable alpha-level, but without any third party runners (in particular, no integration with Visual Studio) it's not realistically usable by anybody but me right now.

PR #2647, opened 2023-01-14
Fixed bug with generics in theory type-chooser

### LINKED ISSUE #1870: Theory type-chooser chooses wrong type in inheritence/generics combo
I've stripped down a real error bug to to the essentials (I believe):

    internal abstract class Low<T> : IEnumerable<T>
    {
        public abstract IEnumerator<T> GetEnumerator();
        IEnumerator IEnumerable.GetEnumerator() => GetEnumerator();
    }

    internal abstract class Mid<T, U> : Low<U> { }

    internal class Top<V, T> : Mid<V, T>
    {
        public override IEnumerator<T> GetEnumerator() { yield break; }
    }

    public class SomeTest
    {
        [Theory, MemberData(nameof(Data))]
        public void TestATheory<T>(IEnumerable<T> source) => Assert.True(source != null);

        public static IEnumerable<object[]> Data() => new[] { new[] { new Top<int, string>(), }};
    }

But xunit has the exception:

    System.ArgumentException : Object of type 'ClassLibrary1.Top`2[System.Int32,System.String]' cannot be converted to type 'System.Collections.Generic.IEnumerable`1[System.Int32]'.

because it has got the generic type parameters mixed up (i.e. it should be trying to convert to `IEnumerable<string>`)

koenigst:
resolves #1870

I found the following issues in the implementation:

* The type resolver did not check that the generic type was the same for the method parameter type and the passed argument.
* The type resolver only looked at the first base class when trying to find the matching type.

### DIFF
--- a/src/xunit.v3.common/Internal/ReflectionExtensions.cs
+++ b/src/xunit.v3.common/Internal/ReflectionExtensions.cs
@@ -1,6 +1,7 @@
 using System;
 using System.Collections.Concurrent;
 using System.Collections.Generic;
+using System.Diagnostics.CodeAnalysis;
 using System.Linq;
 using System.Reflection;
 using Xunit.Sdk;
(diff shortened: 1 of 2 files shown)


dnfadmin:
[CLA assistant check](https://cla.dotnetfoundation.org/xunit/xunit?pullRequest=2647) <br/>Thank you for your submission, we really appreciate it. Like many open source projects, we ask that you sign our [Contributor License Agreement](https://cla.dotnetfoundation.org/xunit/xunit?pullRequest=2647) before we can accept your contribution.<br/><br/>:x: koenigst [sign now](https://cla.dotnetfoundation.org/xunit/xunit?pullRequest=2647)<br/><sub>You have signed the CLA already but the status is still pending? Let us [recheck](https://cla.dotnetfoundation.org/check/xunit/xunit?pullRequest=2647) it.</sub>

bradwilson:
Thanks! I merged this here: https://github.com/xunit/xunit/commit/4d1f2e5d4ac9260487d0a8f35a2d045388021b33

I was reluctant to push my style changes into your PR branch since it was `main`. I assume you'll just pull the new main (unless you decide to delete your fork), but even so I didn't feel comfortable with it. 😂

PR #2657, opened 2023-01-29
#2651 Fix unicode escaping of test names.

### LINKED ISSUE #2651: Unicode escaping for method names is sometimes broken
### Discussed in https://github.com/xunit/xunit/discussions/2583

<div type='discussions-op-text'>

<sup>Originally posted by **tsawyer999** September 15, 2022</sup>
Hello,

I am trying to display the name for a test:
> Parameter "RETURN VALUE" is always added

I tried those two ways and was not able to accomplish the expected result.

>        [Fact]
>        public void Parameter_X22RETURN_VALUEX22_is_always_added()

Result: 
> Parameter "RETURN VALUEX22 is always added

>        [Fact]
>        public void Parameter_X22RETURN_VALUE_X22_is_always_added()

Result: 
>Parameter "RETURN VALUE_" is always added

Can someone give me a hand to achieve the desired result?

Thank you very much!</div>

This is a bug in the encoding routine in DisplayNameFormatter:

https://github.com/xunit/xunit/blob/2b1f75b320b54edc364785d5d7634deff71784f8/src/xunit.v3.core/Sdk/Frameworks/DisplayNameFormatter.cs#L198-L221

Lines 216 adds the literal characters that it already consumed into the output string. Unfortunately, the logic is failing because of these steps:

- It sees the `U` in `VALUEX22` and tries to interpret that as a Unicode escape (4 character)
- It then sees the `E`, which is a valid hex digit, and goes back for another
- It then sees the `X`, which is not a valid hex digit; it adds `UEX` to the output, then loops back to look for more escapes.

Unfortunately, because it consumed the `X` (and put it literally into the output), the algorithm now starts up at `22`, which of course is not an escape sequence, so it puts those into the output as literal values.

The expected encoding is `Parameter "RETURN VALUE" is always added`

The actual encoding is `Parameter "RETURN VALUEX22 is always added`

koenigst:
I implemented a look-ahead method on the `FormatContext`. This allows the implementation to consume the whole escape sequence at once or not at all. This should fix #2651.


bradwilson:
Thanks!

Issue #2621, opened 2022-11-10
`Assert.Equivalent` doesn't ignore non-public properties


ExterminatorX99:
`Assert.Equivalent` docs say it only checks public instance properties, but it actually checks public and non-public, instance and static properties.

Repro
```cs
using Xunit;

namespace Test;

public static class Program
{
    public static void Main()
    {
        var a = new A(123);
        var b = new B(123);

        Assert.Equivalent(a, b);
    }
}

public class A
{
    private static int _ignoreMeStatic { get; set; }

    internal int _ignoreMeInstance { get; set; }

    public int Number { get; set; }

    public A(int number)
    {
        Number = number;
    }
}

public class B
{
    public int Number { get; set; }

    public B(int number)
    {
        Number = number;
    }
}
```

bradwilson:
You are correct. The filter for fields is right, but the filter for properties is not:

https://github.com/xunit/assert.xunit/blob/d4e3195b3e7be0ead099819413e003c62c56eb64/Sdk/AssertHelper.cs#L32-L58

bradwilson:
Fixed in xunit/assert.xunit#58 and tested in https://github.com/xunit/xunit/commit/f1acdedcf4db6cdec53858d876e91ae4a7241ec2

PR #2661, opened 2023-02-02
#2621 Added tests for equivalence check of public instance properties.

koenigst:
Fixes xunit/xunit#2621. Submodule will be updated after xunit/assert.xunit#58 is merged.


bradwilson:
Thanks! This test was hand-merged in https://github.com/xunit/xunit/commit/f1acdedcf4db6cdec53858d876e91ae4a7241ec2 and the v2 tests will be part of a batch back-port from v3 to v2.

Issue #2534, opened 2022-06-11
ArgumentFormatter doesn't pretty-print rank-1 variable-bound arrays correctly


GrabYourPitchforks:
## Summary

`ArgumentFormatter` does not pretty-print rank-1 variable-bound arrays correctly. It prints `int[]` when it should instead print `int[*]`.

## Repro

In the file __ArgumentFormatterTests.cs__, take this theory data:

https://github.com/xunit/xunit/blob/c27a91f8cbcee37cb45699a3a81287bca225e876/src/xunit.v3.assert.tests/Asserts/Sdk/ArgumentFormatterTests.cs#L176-L189

And insert this test case into it:

```cs
{ typeof(int).MakeArrayType(1), "typeof(int[*])" },
```

Then run __build.cmd__ and you will observe the failure when the unit tests run.

```txt
xUnit.net v3 In-Process Runner v0.1.1-pre.30-dev+c27a91f8cb (64-bit .NET Core 3.1.25)
  Discovering: xunit.v3.assert.tests (method display = ClassAndMethod, method display options = None)
  Discovered:  xunit.v3.assert.tests (found 826 test cases)
  Starting:    xunit.v3.assert.tests (parallel test collections = on, max threads = unlimited)
    ArgumentFormatterTests+SimpleValues.TypeValue(type: typeof(int[]), expected: "typeof(int[*])") [FAIL]
      Assert.Equal() Failure
                           ↓ (pos 11)
      Expected: typeof(int[*])
      Actual:   typeof(int[])
                           ↑ (pos 11)
      Stack Trace:
        C:\xunit\src\xunit.v3.assert.tests\Asserts\Sdk\ArgumentFormatterTests.cs(197,0): at ArgumentFormatterTests.SimpleValues.TypeValue(Type type, String expected)
  Finished:    xunit.v3.assert.tests
=== TEST EXECUTION SUMMARY ===
   xunit.v3.assert.tests  Total: 883, Errors: 0, Failed: 1, Skipped: 0, Time: 0.253s
```

## Sample fix

The issue is in these lines in __ArgumentFormatter.cs__.

https://github.com/xunit/assert.xunit/blob/43e3e208a597bc1a4c77f2665f39f98e7401c157/Sdk/ArgumentFormatter.cs#L310-L320

They should have a call to [`Type.IsSZArray`](https://docs.microsoft.com/dotnet/api/system.type.isszarray), as shown below.

```cs
if (typeInfo.IsSZArray)
{
[...]

PR #2669, opened 2023-02-09
#2534 Fix: "ArgumentFormatter doesn't pretty-print rank-1 variable-bound arrays correctly"

Jaswir:
Fix for https://github.com/xunit/xunit/issues/2534

Is this issue still open? Saw there was a pull request a while aga.
I tested it and it works, however it doesn't have a call to IsSZArray as requested in Sample Fix and isn't implemented as an extension method to work downlevel.

Anyway I made a fix for the issue that makes a call to IsSZArray and also works downlevel incase IsSZArray is not available due to .NET Version

Let me know what you think of the fix. And if any further improvements are required.

PS: there was a separate pull request for the submodule


bradwilson:
Thanks!

PR #2684, opened 2023-03-02
Update the EqualException tests to reflect the new collection length …

sdelarosbil:
…being displayed

This is complementary to https://github.com/xunit/assert.xunit/pull/62 which addresses parts of #2352


bradwilson:
I've recently overhauled the assertion messages. Aside from focusing on improving the clarity of messages, I also spent a significant amount of time working through to ensure we don't have double enumeration and that we can safely support collections that never end (which is legal).

Because of the latter situation, I don't believe it's 100% safe to print the length of the collection, because that would require running the entire collection, which may never end.

As such, I'm going to close this PR as something we cannot safely implement. Thank you for the submission!

Issue #2342, opened 2021-07-06
Misleading error message when class fixture constructor throws an exception


supermihi:
When the constructor of a class fixture throws an exception, xunit outputs a messag
> The following constructor parameters did not have matching fixture data

which suggests that the fixture set-up is wrong. The actual exception is not visible at all in the output.
Minimal example:
```csharp
public class Fixture  {
  public Fixture() {
    throw new Exception("foo");
  }
}

public class Tests : IClassFixture<Fixture> {
  public Tests(Fixture _) {  }

  [Fact]
  public void Test() {  }
}
```
Output:
```
  Failed Test [1 ms]
  Error Message:
   System.AggregateException : One or more errors occurred. (foo) (The following constructor parameters did not have matching fixture data: Fixture _)
---- System.Exception : foo
---- The following constructor parameters did not have matching fixture data: Fixture _
  Stack Trace:
  
----- Inner Stack Trace #1 (System.Exception) -----
   at Fixture..ctor() in /home/xxx.cs:line 18
----- Inner Stack Trace #2 (Xunit.Sdk.TestClassException) -----


```

bradwilson:
Verified this is still broken in v3, so added to the roadmap.

supermihi:
thanks

orney21d:
Hi, is still this broken ??

bartelink:
@orney21d AFAICT its still on [the roadmap](https://github.com/xunit/xunit/issues/2133), and this continues to be the tracking issue.

RokasMarcinkevicius:
As this thread shows up on google, what is the best way to properly set it up?

RichardPoes:
Using the example on the [xUnit documentation](url) it is probably because you accidentally wrote
```csharp
[CollectionDefinition("Database collection")]
public class DatabaseTestClass1
```
instead of 
```csharp
[Collection("Database collection")]
public class DatabaseTestClass1
```
At least that was what I did wrong.

mateusz-dobrowolny:
@bradwilson After merge done on March 13, shouldn't this issue be set as "Done"?

bradwilson:
@mateusz-dobrowolny Yes! Thank you! 😄

PR #2686, opened 2023-03-06
Provide a suitable value for the class fixture mapping if its constru…

sdelarosbil:
…ctor throws

It allows the mapping to exist such that we aren't pushing more exceptions due to not detecting the arguments later

This one is rather complex to explain and I am open to review should the fix not be proper, but basically, let's say the fixture class's constructor throws, it prevents the test class's constructor from having an entry in its `ClassFixtureMappings`. This means that later on, during `CreateTestClassConstructorArguments`, it will seem like we didn't provided any arguments needed to the constructor of the test class because it will find the constructor, but not its instance mapping during `TryGetConstructorArgument`. This results in a misleading exception being thrown because it looks like as if the argument was missing when the instance was never in the mappings in the first place.

I am relatively unfamiliar with this project, but I feel fixing this is sort of a catch 22: we would want the mapping to exist so it can proceed without throwing further, but it technically isn't correct because the throw makes the fixture invalid. The only way I found to reliably make it work was to assign the mapping to new object. I fear it might be confusing or not proper so I would at least like someone to review if this would be a proper fix. If not, hopefully I provided enough information to figure out a better way to address this.

### DIFF
--- a/src/xunit.v3.core.tests/Acceptance/FixtureAcceptanceTests.cs
+++ b/src/xunit.v3.core.tests/Acceptance/FixtureAcceptanceTests.cs
@@ -120,6 +120,8 @@ public class FixtureAcceptanceTests
 
 		class ClassWithThrowingFixtureCtor : IClassFixture<ThrowingCtorFixture>
 		{
+			public ClassWithThrowingFixtureCtor(ThrowingCtorFixture _) { }
+
 			[Fact]
 			public void TheTest() { }
 		}
--- a/src/xunit.v3.core/Sdk/v3/Runners/TestClassRunner.cs
+++ b/src/xunit.v3.core/Sdk/v3/Runners/TestClassRunner.cs
@@ -36,7 +36,7 @@ public abstract class TestClassRunner<TContext, TTestCase>
 	protected virtual object?[] CreateTestClassConstructorArguments(TContext ctxt)
 	{
 		var isStaticClass = ctxt.Class.Type.IsAbstract && ctxt.Class.Type.IsSealed;
-		if (!isStaticClass)
+		if (!isStaticClass && !ctxt.Aggregator.HasExceptions)
 		{
 			var ctor = SelectTestClassConstructor(ctxt);
 			if (ctor != null)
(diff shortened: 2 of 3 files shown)


bradwilson:
Yeah, this is odd.

For sure, throwing two errors here is wrong, because we shouldn't be trying to map constructor arguments at all, since we know all the tests will fail. I'm not 100% sure that the right way to skip that behavior is what you're doing here; it works, but as you point out, this is something we shouldn't be doing anyways. I want to peek around inside the code to see if there is something better we can do instead.

bradwilson:
I found a way that I like a little better. I've skipped constructor argument matching when the exception aggregator already has exceptions in it, since there's no point (we won't be creating anything, since all tests will be implicitly failed). This not only stops the extra exception, it skips work that's pointless. 😄 

Aside: The new CLA bot has triggered for this PR, though I'm not sure that's necessarily appropriate given the size of the change. I have a question in to the team about why the [minimum line change setting of 8](https://github.com/xunit/.github/blob/9c9d2d4a61f774dc1cffa49164875c21981e334d/policies/cla.yml#L10) seems to be ignored by the bot in this case. If you're okay signing the CLA, then please feel free to do so; if you're not comfortable with it, then just say so here and I'll merge the PR anyways. 😁 Hopefully the team can get back to me on why this happened.

sdelarosbil:
> @sdelarosbil please read the following Contributor License Agreement(CLA). If you agree with the CLA, please reply with the following information.
> 
> ```
> @dotnet-policy-service agree [company="{your company}"]
> ```
> 
> > Options:
> > 
> > * (default - no company specified) I have sole ownership of intellectual property rights to my Submissions and I am not making Submissions in the course of work for my employer.
> > 
> > ```
> > @dotnet-policy-service agree
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > 
> > * (when company given) I am making Submissions in the course of work for my employer (or my employer has intellectual property rights in my Submissions by contract or applicable law). I have permission from my employer to make Submissions and enter into this Agreement on behalf of my employer. By signing below, the defined term “You” includes me and my employer.
> > 
> > ```
> > @dotnet-policy-service agree company="Microsoft"
> > ```
> 
> Contributor License Agreement

@dotnet-policy-service agree company="Spiria Digital"

sdelarosbil:
It seems I forgot to refference the issue that it closes, #2342

bradwilson:
Thanks!

Issue #2692, opened 2023-03-11
Compare collections using a delegate


bradwilson:
### Discussed in https://github.com/xunit/xunit/discussions/2683

<div type='discussions-op-text'>

<sup>Originally posted by **vc-74** February 27, 2023</sup>
`Assert` currently supports comparing collections, optionally passing an `IEqualityComparer` implementation.

If I'm not mistaken, only `IEqualityComparer.Equals` required in this case (IMO `IEqualityComparer` should have been called `IHashable` and inherit from `IEqualityComparer` which would have only `Equals` but I digress...).

The problem with that approach is that we have to build an `IEqualityComparer` even for simple cases, it would be much simpler to allow passing an inline comparer like:

```csharp
Assert.Equal(col1, col2, (expected, actual) =>
{
    // Comparison logic here
});
```</div>

SARAVANA1501:
https://github.com/xunit/xunit/pull/2693 Started with this, Facing some problem in submodule push, tests has been pushed, will push the assertion code soon

bradwilson:
Shipped 408d2d6bdf5b8d92a35bd0d184dacb4689006cf3

PR #2700, opened 2023-03-27
Added EqualityAssert.Equal method taking two enumerables and a delegate to compare their items

vc-74:
This PR follows [this discussion](https://github.com/xunit/xunit/discussions/2683) which lead to the creation of [this issue](https://github.com/xunit/xunit/issues/2692).

It basically consists in adding a new method to `Assert`, taking in parameter 2 enumerables and a delegate responsible for comparing the items from each enumerable 1 pair at a time.

Code sample (from the unit tests):

```csharp
var expected = new int[] { 1, 2, 3 };
var actual = new int[] { 2, 3, 4 };

Assert.Equal(expected, actual,
    (e, a) =>
    {
    	Assert.Equal(e, a - 1);
    });
```

Note that the enumerables do not need to be associated to the same item type:

```csharp
var expected = new int[] { 1, 2, 3 };
var actual = new string[] { "1", "2", "3" };

Assert.Equal(expected, actual,
    (e, a) =>
    {
        Assert.Equal(e, int.Parse(a));
    });
```

The code is C#6 compatible but would benefit from several improvements available in more recent C# releases (pattern matching, nullable reference types, short `using` syntax...)

Not a 'help wanted' issue.

Yes.

Yes, I tried to stick to the same standards as the edited files.

* Did you write unit tests?
Yes, also included in the PR.

### DIFF
--- a/src/xunit.v3.assert.tests/Asserts/EqualityAssertsTests.cs
+++ b/src/xunit.v3.assert.tests/Asserts/EqualityAssertsTests.cs
@@ -760,6 +760,150 @@ public class EqualityAssertsTests
 		}
 	}
 
+	public class Equal_WithEnumerablesAndDelegate
+	{
+		[Fact]
+		public void Success_SameType()
+		{
+			var expected = new int[] { 1, 2, 3 };
+			var actual = new int[] { 2, 3, 4 };
+
+			Assert.Equal(expected, actual,
+				(e, a) =>
+				{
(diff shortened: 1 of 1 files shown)


vc-74:
@dotnet-policy-service agree

bradwilson:
I believe that this is a duplicate of the functionality merged here: 408d2d6bdf5b8d92a35bd0d184dacb4689006cf3

Thanks for the submission!

PR #2709, opened 2023-04-28
Extending build script for working with preview .NET versions.

Maximys:
Some time ago I started [this discussion](https://github.com/dotnet/runtime/discussions/84764). Current PR fixes problem with invalid parsing .NET preview version format in AutoMapper build script.


bradwilson:
Thanks!

Issue #2435, opened 2021-12-04
Some tests fail outside of the US English culture


randomnoise:
During the execution of `.\build Test`, if a machine works with a culture that uses `,`(comma) instead of `.`(dot) as decimal separator (`tr-TR` in my case), some unit tests that include `decimal`s fail, e.g.:
```
xUnit.net v3 In-Process Runner v0.1.0-pre.485+f71b2c5db9 (64-bit .NET Framework 4.8.4420.0)
  Discovering: xunit.v3.runner.utility.tests (method display = ClassAndMethod, method display options = None)
  Discovered:  xunit.v3.runner.utility.tests (found 105 test cases)
  Starting:    xunit.v3.runner.utility.tests (parallel test collections = on, max threads = unlimited)
    Xunit2AcceptanceTests+Find+CSharp.TheoryAcceptanceTest [FAIL]
      Assert.Collection() Failure
      Collection: [_TestCaseDiscovered("e9f5452c2bbbffa22329be088dc89a9fce2747cc") name="TestClass.TestMethod(x: ???)", _TestCaseDiscovered("bd13d0e5677460f36301b5d556098c1ca05c2c92") name="TestClass.TestMethod(x: 42)", _TestCaseDiscovered("e6e94cad29f0fd42ba3eeba6155b557e3f703378") name="TestClass.TestMethod(x: 42, ???: 21,12)"]
      Error during comparison of item at index 2
      Inner exception: Assert.Contains() Failure
              Not found: TestClass.TestMethod(x: 42, ???: 21.12)
              In value:  TestClass.TestMethod(x: 42, ???: 21,12)
```

This issue was already reported in #1605.

In Windows, you can change decimal separator to `.` (dot) with `Control Panel\All Control Panel Items > Region > Additional settings... > Numbers > Decimal symbol ` and it fixes false negative unit test results.

![image](https://user-images.githubusercontent.com/8210848/144719314-2e84ed46-de7b-45bd-a8e0-b71f228db843.png)

`.\build Test` after change:
![image](https://user-images.githubusercontent.com/8210848/144719446-e91bc874-b232-43d0-9fd3-0370892f4c74.png)

Maybe this problem should be fixed in code but `CulturedFact` and `CulturedTheory` works flaky or I couldn't get them to solve this issue.

Can we at least document this solution somewhere, perhaps in [BUILDING.md](../blob/main/BUILDING.md)? This issue can be frustating for newcomers with European culture settings.

[...]

PR #2722, opened 2023-06-01
Fix culture specific build fail.

Maximys:
Reason of current PR is the same as [83a532d7373d1e2462a68caba5576a9d1c8641e3](https://github.com/xunit/xunit/commit/83a532d7373d1e2462a68caba5576a9d1c8641e3) for #2435


bradwilson:
Thanks!

PR #2746, opened 2023-07-08
Add release notes link to nuspec files

### LINKED ISSUE #2742: Add release notes link to NuGet package
It actually took me a few minutes to find the (well-written and detailed) [2.5.0 release notes](https://xunit.net/releases/2.5.0): Google's top link was to an older version, and [NuGet](https://www.nuget.org/packages/xunit) has no link to them.

It would be helpful for package consumers if the xunit NuGet page promoted the release notes. 

I believe this should be possible by adding `<PackageReleaseNotes>https://xunit.net/releases/2.5.0</PackageReleaseNotes>` here: https://github.com/xunit/xunit/blob/main/src/Directory.Build.props#L48. (One downside might be having to update that URL whenever [version.json](https://github.com/xunit/xunit/blob/2.5.0/version.json) changes, although perhaps `$(BuildVersionSimple)` from Nerdback.GitVersioning could be used to keep it always up-to-date?)

I'm happy to open a PR to add this.

bgrainger:
Fixes #2742

Added `releaseNotes` to nuspec files for projects mentioned on https://xunit.net/releases/2.5.0:

* xunit
* core
* assert
* extensibility
* runners

One downside (mostly for pre-release packages) is that this will generate a link in NuGet to a page that may not ever exist, e.g., https://xunit.net/releases/2.5.1-pre.2.

### DIFF
--- a/src/xunit.assert.nuspec
+++ b/src/xunit.assert.nuspec
@@ -10,6 +10,7 @@
     <licenseUrl>https://licenses.nuget.org/Apache-2.0</licenseUrl>
     <icon>_content/logo-128-transparent.png</icon>
     <readme>_content/README.md</readme>
+    <releaseNotes>https://xunit.net/releases/$PackageVersion$</releaseNotes>
     <description>Includes the assertion library from xUnit.net (xunit.assert.dll). Supports .NET Standard 1.1.</description>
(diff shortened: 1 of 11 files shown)


bradwilson:
Thanks!

Issue #1985, opened 2019-07-18
Use MemberData for methods with optional parameters


Diermeier:
Hello,

using a [Theory] test with [MemberData] which reference a method works fine. But if that method has optional parameters then the full signature call is required in [MemberData]-Attribute.

The below code results in a non-testable situation:

```csharp
    public class MyTestClass
    {
        public static IEnumerable<object[]> MyTestCases(string name, int maxSzenarios = 3)
        {
            for (int i = 1; i <= maxSzenarios; i++)
            {
                yield return new object[] { name, i };
            }
        }

        [Theory]
        [MemberData(nameof(MyTestCases), "MyFirst")]
        [MemberData(nameof(MyTestCases), "MySecond")]
        public void MyTestMethod(string name, int szenario)
        {
            // Do anything ...
        }
    }
```

This works:

```csharp
    public class MyTestClass
    {
        public static IEnumerable<object[]> MyTestCases(string name, int maxSzenarios = 3)
        {
            for (int i = 1; i <= maxSzenarios; i++)
            {
                yield return new object[] { name, i };
            }
        }

        [Theory]
        [MemberData(nameof(MyTestCases), "MyFirst", 2)]
        [MemberData(nameof(MyTestCases), "MySecond", 3)]
        public void MyTestMethod(string name, int szenario)
        {
            // Do anything ...
        }
    }
```

Instead optional parameter I can use classic method overloading for MyTestCases. That would be OK for me. In that case a more specific error message would help.

mateusz-dobrowolny:
Hi @bradwilson, @jnewkirk 

I took a look at this issue and I wrote PoC for the fix.
I'd like someone to take a look at this, and let me know if it is a right direction.
If yes - then I'll proceed with unit test.

I have it on my branch, but I have no rights to upload it. How is the best way to proceed? Should I copy - paste changes here?

Regards

bradwilson:
@mateusz-dobrowolny You should be able to fork this project into your account, and then push your branch there for review.

mateusz-dobrowolny:
@bradwilson  Thanks!
So I made a draft with description in the first comment.
https://github.com/xunit/xunit/pull/2756

bradwilson:
Thanks! I'll take a look soon.

bradwilson:
[...]

PR #2756, opened 2023-07-17
poc for fix issue 1985

mateusz-dobrowolny:
Before you submit a PR...

* Did you write unit tests?

I've checked the issue 1985 and thought about two approaches:
- provide support for methods with optional parameters
- provide only more detailed exception message, as reporter mentioned.

I realized that there may be a lot of cases:
- method with multiple optional parameters
- method with multiple optional parameters, and only some of them are called
- multiple methods with similar set of parameters, some of them may be optional...
- etc.
And this may lead to very complicated algorithm, which will try to make the best guess what method to use, and in many cases this decision can be impossible to be done.

So I decided it is not worth the effort to write a lot of code, instead - as reporter suggested - it should be enough to mention that such methods (with optional parameter) are not supported.

So - I put the exception if no method with exactly corresponding singature was found.

In attached "test class" - there are 2 methods, none of them is proper (both contain optional parameters), so when you run the project (randomly selected), this test will fail.
However, if you uncomment the "proper" member data method, it will pass, as the corresponding method has been found.

P.S. As a side note: I've found that `FirstOrDefault` in line 182 could be Single, as we expect only Single method with given parameters set. However, 11 unit tests failed when changing it to Single, so it may be eventually another small PR.

### DIFF
--- a/src/xunit.v3.core/MemberDataAttributeBase.cs
+++ b/src/xunit.v3.core/MemberDataAttributeBase.cs
@@ -176,15 +176,22 @@ public abstract class MemberDataAttributeBase : DataAttribute
 		var parameterTypes = Parameters == null ? new Type[0] : Parameters.Select(p => p?.GetType()).ToArray();
 		for (var reflectionType = type; reflectionType != null; reflectionType = reflectionType.BaseType)
 		{
-			methodInfo =
-				reflectionType
-					.GetRuntimeMethods()
-					.FirstOrDefault(m => m.Name == MemberName && ParameterTypesCompatible(m.GetParameters(), parameterTypes));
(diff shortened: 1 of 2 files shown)


mateusz-dobrowolny:
@dotnet-policy-service agree

mateusz-dobrowolny:
@bradwilson No hurry with the review. However, as I mentioned in #1985, at this moment it is a PoC and I'd like to know if the general direction of fix is OK.

bradwilson:
Ported to the `v2` branch and committed here: https://github.com/xunit/xunit/commit/b64618e857bb923e176bffa27960645bf8744e7d

I moved your POC test into our acceptance test format (and simplified it a bit).

Thanks!

Issue #2162, opened 2020-10-02
ClassData outputs do not get disposed if they implement IDisposable


jgowdy:
If my TestDataGenerators decorated with [ClassData] implement IDisposable, I would hope that they get Disposed after being used.

I have attached the following test project.

[XUnitTestProject1.zip](https://github.com/xunit/xunit/files/5318755/XUnitTestProject1.zip)

I have also created a pull request that attempts to fix the issue.

https://github.com/xunit/xunit/pull/2161

bradwilson:
Note that we also need to make sure data rows (and class data) that implement `IAsyncDisposable` get properly async disposed, as well as validating that the default implementations of `GetData()` support `IAsyncLifetime` creation.

JamesTerwilliger:
@bradwilson What about something like this?
https://github.com/xunit/xunit/pull/2801

bradwilson:
Fixed by #2801.

Implementation of this feature requires that if the class pointed to be `[ClassData]` implements either `IDisposable` or `IAsyncDisposable`, that we disable data pre-enumeration. That's because otherwise we have no place to do the disposal of the data class itself (the only thing returned from discovery is test cases, and if you have more than one data row, you'll have more than one test case... so no one test case can own the lifetime of the class object).

The most visible result of this will be that tests using such data sources will not show up as "one test per data row" in Test Explorer, but rather as a single test for the test method (and running that test will run all data rows). This is identical to the behavior where we have non-serializable test data.

Available in v3 `0.1.1-pre.307`

PR #2801, opened 2023-10-25
First cut at allowing ClassDataAttribute marked classes to be disposed

JamesTerwilliger:
Attempting to address https://github.com/xunit/xunit/issues/2162
@bradwilson

Before you submit a PR...

* Did you write unit tests?

### DIFF
--- a/src/xunit.v3.core/Sdk/v3/Runners/XunitDelayEnumeratedTheoryTestCaseRunner.cs
+++ b/src/xunit.v3.core/Sdk/v3/Runners/XunitDelayEnumeratedTheoryTestCaseRunner.cs
@@ -235,7 +235,7 @@ public class XunitDelayEnumeratedTheoryTestCaseRunner : XunitTestCaseRunnerBase<
 		{
 			if (trackedObject is IAsyncDisposable asyncDisposable)
 				elapsedTime += await ExecutionTimer.MeasureAsync(() => ctxt.CleanupAggregator.RunAsync(asyncDisposable.DisposeAsync));
-			if (trackedObject is IDisposable disposable)
+			else if (trackedObject is IDisposable disposable)
 				elapsedTime += ExecutionTimer.Measure(() => ctxt.CleanupAggregator.Run(disposable.Dispose));
 		}
 
--- a/src/xunit.v3.core/Sdk/v3/Runners/XunitDelayEnumeratedTheoryTestCaseRunner.cs
+++ b/src/xunit.v3.core/Sdk/v3/Runners/XunitDelayEnumeratedTheoryTestCaseRunner.cs
@@ -119,8 +119,7 @@ public class XunitDelayEnumeratedTheoryTestCaseRunner : XunitTestCaseRunnerBase<
 				foreach (var dataRow in data)
 				{
 					var dataRowData = dataRow.GetData();
-					foreach (var dataRowItem in dataRowData)
-						ctxt.DisposalTracker.Add(dataRowItem);
+					ctxt.DisposalTracker.AddRange(dataRowData);
 
 					_ITypeInfo[]? resolvedTypes = null;
 					var methodToRun = ctxt.TestMethod;
--- a/src/xunit.v3.core/Sdk/v3/Runners/XunitDelayEnumeratedTheoryTestCaseRunner.cs
+++ b/src/xunit.v3.core/Sdk/v3/Runners/XunitDelayEnumeratedTheoryTestCaseRunner.cs
@@ -99,7 +99,7 @@ public class XunitDelayEnumeratedTheoryTestCaseRunner : XunitTestCaseRunnerBase<
 					continue;
 				}
 
-				var data = await discoverer.GetData(dataAttribute, ctxt.TestCase.TestMethod.Method);
+				var data = await discoverer.GetData(dataAttribute, ctxt.TestCase.TestMethod.Method, ctxt.DisposalTracker);
 				if (data is null)
 				{
 					ctxt.Aggregator.Add(
--- a/src/common.tests/TestDoubles/Mocks.Attributes.cs
+++ b/src/common.tests/TestDoubles/Mocks.Attributes.cs
@@ -98,7 +98,7 @@ public static partial class Mocks
 		var dataAttribute = Substitute.For<DataAttribute>();
 		dataAttribute.Explicit = @explicit.GetValueOrDefault();
 		dataAttribute.Skip = skip;
-		dataAttribute.GetData(null!).ReturnsForAnyArgs(dataRows);
+		dataAttribute.GetData(null!, null!).ReturnsForAnyArgs(dataRows);
 
 		var result = Substitute.For<_IReflectionAttributeInfo, InterfaceProxy<_IReflectionAttributeInfo>>();
 		result.Attribute.Returns(dataAttribute);
@@ -111,7 +111,7 @@ public static partial class Mocks
 	public static _IReflectionAttributeInfo DataAttribute(params ITheoryDataRow[] data)
 	{
 		var dataAttribute = Substitute.For<DataAttribute>();
-		dataAttribute.GetData(null!).ReturnsForAnyArgs(data.CastOrToReadOnlyCollection());
+		dataAttribute.GetData(null!, null!).ReturnsForAnyArgs(data.CastOrToReadOnlyCollection());
 
 		var result = Substitute.For<_IReflectionAttributeInfo, InterfaceProxy<_IReflectionAttributeInfo>>();
 		result.Attribute.Returns(dataAttribute);
--- a/src/xunit.v3.common/Utility/DisposalTracker.cs
+++ b/src/xunit.v3.common/Utility/DisposalTracker.cs
@@ -74,7 +74,8 @@ public class DisposalTracker : IAsyncDisposable
 	void AddInternal(object? @object)
 	{
 		if (@object is not null)
-			trackedObjects.Push(@object);
+			if (@object is IDisposable || @object is IAsyncDisposable)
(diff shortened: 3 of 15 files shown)


JamesTerwilliger:
@dotnet-policy-service agree

JamesTerwilliger:
@dotnet-policy-service agree

bradwilson:
Looking at this implementation, we now have a split between the disposal tracker usage between the data itself (which is itself split between `TheoryDiscoverer` and `XunitDelayEnumeratedTheoryTestCaseRunner`) and now the classes being added to the tracker by `ClassDataAttribute`. I think if we're going to pass along `DisposalTracker` into the data discoverers (which they delegate to the attributes), then we should expect them to own the disposal of everything: data + anything extra like classes that get instantiated.

And as a side note, this split is unnecessary:

```csharp
if (classInstance is IDisposable)
	disposalTracker.Add(classInstance);
```

```csharp
if (classInstance is IAsyncDisposable)
	disposalTracker.Add(classInstance);
```

[`DisposalTracker` accepts raw `object`](https://github.com/xunit/xunit/blob/6f8fc7ef23acb767a0fba63dedf9cad77e7e5a2e/src/xunit.v3.common/Utility/DisposalTracker.cs#L46) and will deal with [whether or not the thing is disposable or not (and how)](https://github.com/xunit/xunit/blob/6f8fc7ef23acb767a0fba63dedf9cad77e7e5a2e/src/xunit.v3.common/Utility/DisposalTracker.cs#L135-L138). Don't be concerned about optimizing the storage in `DisposalTracker` because I'd prefer to fix it in `DisposalTracker` and not introduce conditional logic all across the rest of the code base.

JamesTerwilliger:
@bradwilson fair points all, will post an update later today or tomorrow. I did see the split between `TheoryDiscoverer` and `XunitDelayEnumeratedTheoryTestCaseRunner` and tried to accommodate, but I'll try data discoverer ownership and see what happens.

bradwilson:
Thanks for doing an update. If you opt not to, I can also pick up from where you started and make the changes.

JamesTerwilliger:
@bradwilson Having a chance to look at this a little more deeply, we have a situation where `XunitDelayEnumeratedTheoryTestCaseRunner` already wants to own the disposal of everything in the DisposalTracker in `RunTestsAsync`.

JamesTerwilliger:
I'd love to see this one through if it's OK, just want to make sure I understand the design parameters.

JamesTerwilliger:
An alternative would be for the GetData methods on DataDiscoverer and DataAttribute to return any objects that need disposing as an out parameter or a tuple member, though I'm not sure that makes it any more elegant... but perhaps it simplifies the DisposalTracker management?

bradwilson:
After bouncing around a bunch of ideas, the high level problem is this:

* On a per-test basis, we have a set of data that needs to be disposed (namely, the single row of data that's passed as arguments to the test method). This is currently owned by `TestMethodTestCase` in its own private `DisposalTracker`, which is cleaned up when we dispose the test case.
* On a per-discoverer basis, we may have an additional set of data that needs to be disposed (in the case of `ClassAttribute` being discovered by `DataDiscoverer`, the instance of the test class that was created to retrieve the data).

And I realize there are two different scenarios:

* The class object can be disposed after all the data has been discovered
* The class object cannot be disposed until all the data has been used to run the tests

I don't think there's any reasonable way for us to accommodate the second scenario if the developer wants us to be able to report one test case per data row (the mechanism that allows the user to run just a single data row in Test Explorer in Visual Studio). There's nowhere for us to store the instance of the class object where it could later be disposed after all the tests have been run. The granularity of the thing we pass back during discovery is the test case; there's no higher level concept during discovery.

So I think a stake has to be put into the ground, and one of two paths needs to be chosen:

1. If class objects need to stick around while their data is still alive, then we need to update `ClassAttribute` so that it always says "my data cannot be pre-enumerated" if the class in question is either `IDisposable` or `IAsyncDisposable`.
2. If class objects are assumed to be disposed after the data is discovered, and don't necessarily live while the data they returned lives, then we can just update `ClassAttribute` to dispose of the class after it has discovered all the data.

We don't know from the overly simplistic example from #2162, but my gut tells me that the _safest_ assumption would be the first one. It's easy to imagine a scenario where a data class does something like opens a connection to a remote service and pulls data from it, and in some way that data is only valid in memory while the connection to the remote service is still alive. If we dispose the class after discovering the data, we may be invalidating the data.

So after thinking it through, I think that's the way this feature has to exist. So if we're passing a `DisposalTracker` into `IDataDiscoverer.GetData`, here are the tasks as I see them:

1. Update `DisposableTracker.AddInternal` to only store objects which actually implement `IDisposable` or `IAsyncDisposable` (this is important for task 4 and 7).
2. Update `IDataDiscoverer.GetData` to accept a `DisposalTracker`
3. Update `DataAttribute.GetData` to accept a `DisposalTracker`
4. Update `TheoryDiscoverer` to create and pass along a `DisposalTracker` to each of the data discoverers. If a data discoverer adds anything to the tracker, then `TheoryDiscoverer` needs to fall back to returning an instance of `XunitDelayEnumeratedTheoryTestCase`. This is basically a safety valve to prevent us from returning individual test cases and not having a way to dispose of that tracker. (Note that we should immediately dispose of all the data that was returned, in addition to the tracker, so that we don't forget to dispose things just because we couldn't use them.)
5. Create a new `ClassDataDiscoverer`, derived from `DataDiscoverer`, which overrides `SupportsDiscoveryEnumeration` to check to see if the class in question implements `IDisposable` or `IAsyncDisposable` and returns `false` if so (to prevent the enumeration from happening during discovery). This will also cause `TheoryDiscoverer` to fall back to `XunitDelayEnumeratedTheoryTestCase` (but in a more optimal way that doesn't involve us discovering a bunch of data that we would have to immediately throw away).
6. Make sure to decorate `ClassDataAttribute` with the new discoverer.
7. Update `ClassDataAttribute.GetData()` to stash the instance of the class it creates into the tracker. It does not need to worry about whether it's disposable or not, because task 1 will already filter it out and task 4 won't be tripped up with a non-disposable class inside the tracker.
8. Now all cases where the class (or any non-per-test-row data) needs to be disposed will end up inside `XunitDelayEnumeratedTheoryTestCaseRunner` (either because they declined pre-enumeration, or they tried to add a disposable object to the tracker during discovery). During discovery in `AfterTestCaseStartingAsync` it will pass along `ctxt.DisposalTracker` to the data discoverers so they can add things to be disposed. The place where it's disposing everything is still the correct place, because that will happen after all the tests have been run.

I believe this will adequately implement the feature. It comes with the caveat that a disposable class data source means no pre-enumeration, but honestly I can't think of any reason you'd need to have your class data source be disposable unless the data you returned was somehow dependent on something that also needs to be cleaned up after all the data is consumed; otherwise, the class could just create resources, discover the data, dispose the resources it needed, and then return the data. Adding a lifetime-related disposal requirement pretty much always should mean "don't dispose me until the data is used".

Thoughts?

JamesTerwilliger:
I think it's a great analysis of the situation for sure, and I think the proposed solution is the safest one. The PR currently does items 2 and 3 and about 10% of 5, and doing the rest doesn't seem like a great deal of work. I can give it a shot.

I think you're right about the class itself being disposable implying that the data rows themselves have at least a reasonable chance of requiring a live connection. The canonical anti-example I keep coming up with is a class that pulls test input rows from a database. In that case, the class itself would still not be disposable (or at least could be engineered in a way that did not need it to be disposable) because the GetEnumerator method itself could/would handle the connection and cleanup. The data rows themselves would not at all require a live connection after enumeration.

But then again, I can also imagine someone trying to play with the IDisposable nature of the feature this would add and say to themselves that they can just make the class disposable, add the connection to the constructor, and the connection destruction to the Dispose() method.

bradwilson:
Yep, I agree that doing self-disposal of the database connection during enumeration makes the most sense if people want to keep individually running rows, and it'll be easy for me to tell people to do that when they come to complain that they can't run individual rows. 😁

JamesTerwilliger:
Roger dodger, on it

JamesTerwilliger:
OK, I think I got all the points covered. I'd add a test case but for some reason the v3 tests are not showing up in the local runner on my machine.

bradwilson:
Yeah, v3 doesn't have support for VSTest yet, so you won't see anything in Visual Studio (or VS Code). However, each project is runnable as they're stand-alone executables in v3 now, so you can just set your startup project to whatever set of tests you want to run and then Ctrl+F5 to run or F5 to debug.

bradwilson:
And of course `./build test` (from a VS Command Prompt) should run all tests.

src/xunit.v3.core/Sdk/v3/Runners/XunitDelayEnumeratedTheoryTestCaseRunner.cs
```diff
@@ -235,7 +234,7 @@ protected override async ValueTask<RunSummary> RunTestsAsync(XunitDelayEnumerate
 		{
 			if (trackedObject is IAsyncDisposable asyncDisposable)
 				elapsedTime += await ExecutionTimer.MeasureAsync(() => ctxt.CleanupAggregator.RunAsync(asyncDisposable.DisposeAsync));
-			if (trackedObject is IDisposable disposable)
+			else if (trackedObject is IDisposable disposable)
```
bradwilson, line 237: Good catch! This was definitely a bug.

bradwilson:
I'll add the test cases. I'm going to be a little bit of formatting as well.

bradwilson:
Thanks!

JamesTerwilliger:
@bradwilson Thank you so much! It's an incredible code base.

Issue #2090, opened 2020-04-14
Empty XML with results from DelegatingXmlCreationSink


sobkowicz:
I have problem with generated XML after running tests. I expect valid XML with tests results which I get from `DelegatingXmlCreationSink` but instead sometimes this XML is empty and results appear after a short time (I think 10-20 ms).

I use custom AssemblyRunner that use `IFrontController`, `DelegatingExecutionSummarySink` and `DelegatingXmlCreationSink` to run tests and get generated XML with results.
Details you can find in sample project (only key features to reproduce the problem).

**Sample app:**
https://github.com/sobkowicz/xunit-sink-xml-app

**Steps to reproduce:**
Download and run sample app without debuging. Empty XML appears only sometimes so I added runnning tests in loop, you can increase loop count. I added saving empty XMLs to stream and then to a file. About 1000 loops is enough to get empty XML in most cases.

**Expected XML:**
```
<assembly name="ExternalTests.dll"
          environment="64-bit .NET Standard [collection-per-class, parallel (8 threads)]"
          test-framework="xUnit.net 2.4.1.0"
          run-date="2020-04-14"
          run-time="12:03:00"
          total="1"
          passed="1"
          failed="0"
          skipped="0"
          time="0.001"
          errors="0">
    <errors />
    <collection total="1"
                passed="1"
                failed="0"
                skipped="0"
                name="Test collection for ExternalTests.ExampleTests"
                time="0.000">
        <test name="ExternalTests.ExampleTests.EmptyAndPassingTest"
              type="ExternalTests.ExampleTests"
              method="EmptyAndPassingTest"
              time="0.0000036"
              result="Pass" />
    </collection>
</assembly>
```

**XML in some cases:**
```
<assembly name="ExternalTests.dll"
          environment="64-bit .NET Standard [collection-per-class, parallel (8 threads)]"
          test-framework="xUnit.net 2.4.1.0"
          run-date="2020-04-14"
          run-time="12:03:00">
    <errors />
</assembly>
```

xunit version: 2.4.1
xunit.runner.utility: 2.4.1
xunit.extensibility.core: 2.4.1
.NET Core version: 2.1
OS: Windows 10

bradwilson:
Oof, given how simple the repro is, this feels like it might be a race condition in the reporting. :frowning:

kdthomas2121:
Is there any update to this? Experiencing the same 25% of the time

bjorn-stromberg-codic:
[...]

Issue #1534, opened 2017-10-26
Results XML is occasionally incomplete


dougbu:
1. Clone the https://github.com/aspnet/AspNetWebStack repo
2. `cd AspNetWebStack`
3. `.\build.cmd`
4. Look for warnings at the end of the build, from the `PrintTestRunSummary` target
5. If none are output, goto 3

#### Expected
The above loop should be pretty much endless, never seeing `PrintTestRunSummary` warnings. (See this [relatively-new block](https://github.com/aspnet/AspNetWebStack/blob/7674394480666e66830689d07002e3a2e26ebc32/tools/WebStack.tasks.targets#L181-L192) to output those warnings.)

#### Actual
Usually only need to loop a few times before warnings appear. A varying results file is affected when warnings do occur. Size of the results file and duration of the test run doesn't seem to matter. Every so often, two or three results files are incomplete.

The test run for the incomplete results does not indicate any issues i.e. the problem is visible only later, when `PrintTestRunSummary` examines the results file. Test execution summary is also normal e.g. 
```
    Finished:    Microsoft.Web.Helpers.Test
  === TEST EXECUTION SUMMARY ===
     Microsoft.Web.Helpers.Test  Total: 162, Errors: 0, Failed: 0, Skipped: 0, Time: 11.500s
```

### Notes
This issue is inconsistent but may occur more when the system is loaded. For example, testing in parallel (the default) makes the problem occur more often. Serial testing (`/p:TestInParallel=false` on the `.\build.cmd` command line) makes the build almost-but-not-quite completely reliable on an otherwise lightly-used system.

Have seen the warnings in `CodeAnalysis` (the default), `Debug` and `Release` builds. Might occur more often with `Debug` i.e. `.\build.cmd UnitTest /p:Configuration=Debug`.

This repo uses the MSBuild runner in command-line builds. But, wouldn't be too surprising if the Console runner was also affected.

This problem is _not_ related to #1192. The results file is valid, not truncated.

Appears like `DelegatingXmlCreationSink.HandleTestAssemblyStarting(...)` is called but `HandleTestAssemblyFinished(...)` is not.

#### Partial results
``` xml
<?xml version="1.0" encoding="utf-8"?>
<assemblies timestamp="10/26/2017 09:04:34">
[...]

PR #2810, opened 2023-11-04
Changing DelegatingSummarySink and DelegatingXmlCreationSink to invok…

jonlegendre:
Changing `DelegatingSummarySink` and `DelegatingXmlCreationSink` to invoke` innerSink.OnMessage` after local handling of messages.

fixing #2090 and maybe #1534

Note I tried to only invoke the local `HandleTestAssemblyFinished` after `innerSink.OnMessage` per the comment in #2090 
> We could move the inner sink invocation down as you suggest, with the caveat that we call HandleTestAssemblyFinished after the inner sink has had a chance to run.

However, this did not pass unit tests nor fix the issue.

The current changes I have made do pass all unit tests and also pass the test from https://github.com/sobkowicz/xunit-sink-xml-app.

I don't love the new unit test I added, and I am uncertain if you will like the location of the new class made to facilitate the test. Honestly I am not sure this test adds value, but it does demonstrate that the innerSink.OnMessage is invoked successfully as part of the process.

If this is accepted I will make a similar PR for V2.

### DIFF
--- a/src/xunit.v3.runner.common.tests/Sinks/DelegatingSinks/DelegatingXmlCreationSinkTests.cs
+++ b/src/xunit.v3.runner.common.tests/Sinks/DelegatingSinks/DelegatingXmlCreationSinkTests.cs
@@ -3,7 +3,6 @@ using System.Collections.Generic;
 using System.Globalization;
 using System.IO;
 using System.Linq;
-using System.Threading;
 using System.Xml.Linq;
 using System.Xml.XPath;
 using NSubstitute;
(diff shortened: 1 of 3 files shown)


bradwilson:
There is an issue with the test setup, specifically the implementation of `ExecutionSinkThatFillsExecutionSummaryWhenOnMessageInvoked.OnMessage`.

The problem with this setup is that it doesn't correctly mimic the behavior of DelegatingSummarySink. It will record an execution summary regardless of what message is sent to it, and in your test you send two messages. So you've accidentally side-stepped the race condition by creating the execution summary (at the wrong time) because of the `_ErrorMessage`.

If instead of the fake type, we use this code to set up a real instance of DelegatingSummarySink:

```csharp
var project = new XunitProject();
var assembly = new XunitProjectAssembly(project);
var discoveryOptions = _TestFrameworkOptions.ForDiscovery();
var executionOptions = _TestFrameworkOptions.ForExecution();
var innerSink = Substitute.For<_IMessageSink>();
IExecutionSink localSink = new DelegatingSummarySink(assembly, discoveryOptions, executionOptions, AppDomainOption.Disabled, false, innerSink);
```

Doing that instead causes the test to fail like I would've expected, because of the ordering problem:

```
DelegatingXmlCreationSinkTests.AddsAssemblyFinishedInformationToXmlWhenSummaryIsFilledByOnMessage[en-US] [FAIL]
Assert.Equal() Failure: Strings differ
           ↓ (pos 0)
Expected: "2112"
Actual:   "0"
           ↑ (pos 0)
```

bradwilson:
I'm going to see if I can update the tests on `main` and `v2` to use a real instance of the `DelegatingSummarySink` so that they can properly illustrate the behavior.

bradwilson:
I pushed the unit test change into `main`, merged `main` into your branch, and deleted the test you added (since it's no longer necessary, as the test above is now already using the real XmlSummarySink). The test now fails appropriately.

I think this fundamentally illustrates the ordering problem: do the right thing for the race condition by moving the inner sink invocation to the bottom, and it breaks `DelegatingXmlCreationSink`.

I'm still thinking about ways to solve this that don't involve breaking changes and/or duplicated types.

jonlegendre:
closing for now. I am thinking on it as well.

Issue #2799, opened 2023-10-23
Support static member discovery in default memberdata data discoverer


boerdereinar:
The default data discover currently only supports test discovery in classes and their base types; however, interfaces can also contain static members which can contain test data.

```cs
public abstract class Repro<TSelf> where TSelf : ITest
{
	[Theory]
	[MemberData(nameof(TSelf.TestData))]
	public void Test(int value)
	{
	}
}

public sealed class ReproImpl : Repro<ReproImpl>, ITest
{
}

public interface ITest
{
	static virtual TheoryData<int> TestData => new() { 1, 2, 3 };
}
```

bradwilson:
This will also likely need to include an update to the analyzers to recognize this new pattern (version limited by whatever version we choose to ship this feature into).

JamesTerwilliger:
@bradwilson I can poke at this one some more and add tests, but I gave a first try on this one here: https://github.com/xunit/xunit/pull/2861.

bradwilson:
Available in `0.1.1-pre.372` https://xunit.net/docs/using-ci-builds

PR #2861, opened 2024-01-08
Allow MemberData methods and properties to be discovered on interfaces as well as classes

JamesTerwilliger:
Unit tests hopefully to follow

This PR allows properties and methods referenced by a MemberData attribute to be discovered on interfaces as well as derived classes.

Intended to address https://github.com/xunit/xunit/issues/2799


JamesTerwilliger:
@bradwilson It's odd, I tried writing a test for this, but I couldn't get the test runner to pick up the test I wrote in this PR.

JamesTerwilliger:
(The test added in the PR currently is not the only approach I tried, I did several things all gated on NET70 or higher and all failed to pick up the test somehow. What's in the PR is just the commonality between all of what I tried.)

bradwilson:
I've updated the test and push the change.

bradwilson:
I assume this makes it ready to go? Or did you have something else in mind?

JamesTerwilliger:
@bradwilson If the test passes, it's ready to go!

bradwilson:
Thanks!

PR #2864, opened 2024-01-16
Emit warning messages to the runner for config files when relevant

### LINKED ISSUE #1655: Add diagnostic message when JSON file is malformed

sdelarosbil:
v3 support #1655 (and supercedes #2434)

As discussed in the issue, I did the changes for the v3 api and then implemented them to the IRunnerLogger whenever the runner supported it. This also includes amending the existing unit tests. I had to change the 2 CommandLine class due to the need of having to parse, store the warnings and then later log them, not sure if this is a problem, but we can discuss if it is.

The warnings I put are the following:
- The file was sent explicitly, but it wasn't a .json / .config file or doesn't exist (no warning if no filename was sent)
- The file exists, but the root object isn't a json one
- The file exists, but it failed to read
- The file exists, was read, but the deserialisation failed

Let me know if this is exhaustive, I tried to cover all the runners that had to be refactored from this. I tested in inproc runner and it seemed to work.

### DIFF
--- a/src/xunit.v3.runner.common/Configuration/ConfigReader_Json.cs
+++ b/src/xunit.v3.runner.common/Configuration/ConfigReader_Json.cs
@@ -3,6 +3,7 @@ using System.Collections.Generic;
 using System.Globalization;
 using System.IO;
 using System.Text.Json;
+using Xunit.Internal;
 using Xunit.v3;
 
 namespace Xunit.Runner.Common;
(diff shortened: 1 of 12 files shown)


bradwilson:
Looking through where this lands, there's a lot of allocation of `List<string>` that is probably unnecessary. I'm going to play around with what you've got here and see if I can find something I like that feels less over-allocate-y. 😄

bradwilson:
I definitely like the warnings better as an optional `List<string>` rather than an `out`, as it cleans up other usage and eliminates the over-allocation and copying.

bradwilson:
> The file was sent explicitly, but it wasn't a .json / .config file or doesn't exist (no warning if no filename was sent)

I took this out, because of the way `ConfigReader` works: it depends on the individual readers (`ConfigReader_Json` and `ConfigReader_Configuration`) to silently ignore files that aren't appropriate, so this just goes back to a silent `return false`.

I also added some logic at the end of `ConfigReader.Load` to say "if you wanted warnings, and you provided a config file, and yet everybody returned false with no warnings, then you're an unsupported file type".

> Let me know if this is exhaustive

This list seems good. I may come back later and add some additional warnings related to JSON values that are getting ignored, but I'm not 100% sure on this one yet. There's no realistic way to do this on the Configuration side, but I consider `.config` files to be deprecated (they're not even supported for v3 tests, only supported for v1 and v2 backward compatibility for the runners).

bradwilson:
Thanks!

PR #2875, opened 2024-01-27
Indicate that you must run `build.ps1` on Windows.

JayBazuzi:


bradwilson:
Funny, we were both on a consolidation and simplification path at the same time. :) Please check out the new scripts and let me know if you think there are more changes to be made.

bradwilson:
(The updates should allow you to run `./build` regardless of whether you're using bash or PowerShell, regardless of OS, because PowerShell will automatically pick up the .ps1 version before it tries to run the bash version, so this doc update should not be needed with the updated scripts that are cross-platform for both bash and PowerShell.)

JayBazuzi:
Awesome.

Issue #2629, opened 2022-12-04
Add support for CollectionAttribute<T> and CollectionDefinitionAttribute<T>


jeffward01:
I reviewed the code and I do not think this feature exists, but I just wanted to ask anyways.

# Question

Is it possible to use `[CollectionDefinition("myCollectionFixtureName")]` with generic attributes?

[Here is the current documentation ](https://xunit.net/docs/shared-context)for `ICollectionFixture<>`

### Current 

```csharp

[CollectionDefinition("Database collection")]
public class DatabaseCollection : ICollectionFixture<DatabaseFixture>
{
    // This class has no code, and is never created. Its purpose is simply
    // to be the place to apply [CollectionDefinition] and all the
    // ICollectionFixture<> interfaces.
}



[Collection("Database collection")]
public class DatabaseTestClass1
{
    DatabaseFixture fixture;

    public DatabaseTestClass1(DatabaseFixture fixture)
    {
        this.fixture = fixture;
    }
}

````

### Expected C# 11 functionality

```csharp

[CollectionDefinition<DatabaseCollection>]
public class DatabaseCollection : ICollectionFixture<DatabaseFixture>
{
    // This class has no code, and is never created. Its purpose is simply
    // to be the place to apply [CollectionDefinition] and all the
    // ICollectionFixture<> interfaces.
}



[Collection<DatabaseCollection>]
public class DatabaseTestClass1
{
    DatabaseFixture fixture;

    public DatabaseTestClass1(DatabaseFixture fixture)
    {
        this.fixture = fixture;
    }
}

```

-----

## Summary
* Is this possible?
*  Is this something that I can open an issue for and myself _(alas... if i have time)_ or someone else can  implement?

jeffward01:
I also opened a discussion here that has not gotten much traction: https://github.com/xunit/xunit/discussions/2627

benmccallum:
Just found myself wondering this too, if I ever get time I could explore this.

benmccallum:
I think it'd be good to rename this issue @jeffward01 to something easier to find like "Use C#11's generic attributes for xunit attributes" as this would be something that'd be useful for a quite a few attributes I imagine, e.g. `[ClassData<T>]`

bradwilson:
[...]

PR #2878, opened 2024-01-31
Allow CollectionAttribute to reference a definition class directly

JamesTerwilliger:
Attempted fix for https://github.com/xunit/xunit/issues/2629

This PR allows CollectionAttribute some flexibility in referencing a collection definition/fixture. It gives two new options:
- Take a class type reference rather than a string as a collection identifier, and
- Even "better"/"newer", allow `CollectionAttribute<TCollectionDefinition>` as a generic way to also reference the fixture class

More tests to come


src/xunit.v3.core/CollectionDefinitionAttribute.cs
```diff
@@ -13,6 +13,15 @@ namespace Xunit;
 [AttributeUsage(AttributeTargets.Class, AllowMultiple = false, Inherited = false)]
 public sealed class CollectionDefinitionAttribute : Attribute
 {
+	/// <summary>
+	/// Initializes a new instance of the <see cref="CollectionDefinitionAttribute" /> class.
+	/// Only use this constructor when collection references by test classes use the generic
+	/// <see cref="CollectionAttribute{TCollectionDefinition}"/> attribute or refer to the
+	/// fixture class using typeof(CollectionDefinition).
+	/// </summary>
+	public CollectionDefinitionAttribute()
```
bradwilson, line 22: I think with the understanding that `[Collection(type)]` and `[Collection<T>]` meaning we don't need `[CollectionDefinition]`, we probably don't want to add this constructor. We can leave this class only for definitions discovered by name.

src/xunit.v3.common/Reflection/ReflectionAttributeInfo.cs
```diff
@@ -90,6 +90,13 @@ public IReadOnlyCollection<_IAttributeInfo> GetCustomAttributes(string assemblyQ
 					list ??= new List<ReflectionAttributeInfo>();
 					list.Add(new ReflectionAttributeInfo(attr));
 				}
+				else if (attributeType.IsGenericTypeDefinition
```
bradwilson, line 93: This new behavior probably needs to get documented into `_IAttributeInfo`.

src/xunit.v3.core/CollectionBehavior.cs
```diff
@@ -7,14 +7,16 @@ public enum CollectionBehavior
...
 	/// </summary>
 	CollectionPerAssembly,
 
 	/// <summary>
 	/// By default, generates a collection per test class for any test classes that are not
-	/// decorated with <see cref="CollectionAttribute"/>.
+	/// decorated with <see cref="CollectionAttribute"/> or
+	/// <see cref="CollectionAttribute{TCollectionDefinition}"/>.
 	/// </summary>
 	CollectionPerClass
```
bradwilson, line 21: Reading this reminds me that we probably need to make sure that the following two classes end up in the same collection:

```csharp
public class TestClass1 { ... }

[Collection<TestClass1>]
public class TestClass2 { ... }
```

which means the collection ID for default classes during `CollectionPerClass` also needs to use `UniqueIDGenerator`.

And in thinking about this, it probably needs to be the case that we want the name to continue to include the class name, and just use the unique ID as a discriminator. So today the default collection in `CollectionPerClass` is called something like `"Test collection for <Type.FullName>"`, whereas now both the default and `[Collection<T>]` should probably have a name of `"Test collection for <Type.FullName> (<UniqueID>)"` for the uniqueness benefit.

bradwilson: There are tests that might be affected, probably in `CollectionAcceptanceTests`, `CollectionPerClassTestCollectionFactoryTests`, and `Xunit3AcceptanceTests`.

src/xunit.v3.core/Sdk/v3/Utility/TestCollectionFactoryHelper.cs
```diff
@@ -26,7 +27,8 @@ public static class TestCollectionFactoryHelper
 				.Select(type => new { Type = type, Attribute = type.GetCustomAttributes(typeof(CollectionDefinitionAttribute).AssemblyQualifiedName!).FirstOrDefault() })
 				.Where(list => list.Attribute is not null)
 				.GroupBy(
-					list => list.Attribute!.GetConstructorArguments().Cast<string>().Single(),
+					list => list.Attribute!.GetConstructorArguments().Cast<string>().SingleOrDefault()
```
bradwilson, line 30: You won't need this once we remove the empty constructor, though the documentation should be updated to indicate that it will only discover test collections that are intended to by discovered by name rather than by type.

bradwilson (commented):
I haven't really looked at the tests yet, as I wanted to see the impact on the production code first, and a few more things popped out to me as I did that. Thoughts?

src/xunit.v3.core/CollectionDefinitionAttribute.cs
JamesTerwilliger: I would agree, except `CollectionDefinitionAttrtibute` has the parallelism property that has no other way of being specified. Without the default constructor, one would always have to specify a name to also specify the other parameter.

JamesTerwilliger: Said differently, if `CollectionDefinitionAttribute` had no fields other than name/type, I'd 100% agree, but there is one itty-bitty teeny-weeny boolean value still there that has no other way to be set.

src/xunit.v3.core/CollectionBehavior.cs
JamesTerwilliger: Just trying to make sure I understand. In the current situation, if I do this:

`
[CollectionDefinition("bacon")]
public class TestClass1 { ... }

[Collection("bacon")]
public class TestClass2 { ... }
`

Are these classes considered part of the same collection today? If so, then completely agreed. But I wasn't sure what the behavior is right now if a test class is adorned with `CollectionDefinitionAttribute` rather than putting that attribute on a fixture.

bradwilson: Actually ignore this for now. I might want to play around with some things to try to decide whether it's actually what I really want or not. It may be one of those ideas that sounds good for 5 minutes, but bad after you spend half a day with it. 😂

JamesTerwilliger: Gold 5 standing by

src/xunit.v3.core/CollectionDefinitionAttribute.cs
bradwilson: Good point. It can become optional then.

src/xunit.v3.common/Reflection/ReflectionAttributeInfo.cs
JamesTerwilliger: Done!

src/xunit.v3.core/CollectionBehavior.cs
bradwilson: Resolving as "ignore me" on this one. 😄

bradwilson:
I'm going to add tests for the `ReflectionAttributeInfo` changes, as well as porting the changes into `ReflectionAssemblyInfo`, `ReflectionMethodInfo`, `ReflectionParameterInfo`, and `ReflectionTypeInfo`.

JamesTerwilliger:
Awesome! I'll look into making changes to the analyzer to compensate once this goes in.

bradwilson:
This is becoming...a large change to get this all working the way I want. I think I might end up doing it in the context of the `main` branch (since it's a substantial change by itself) and then I'll merge it in once I'm done.

JamesTerwilliger:
Understood.

bradwilson:
This is the commit: https://github.com/xunit/xunit/commit/02175c74514fa2c8ecfae2f0621db41f90af3089

> Showing 38 changed files with 935 additions and 305 deletions. 

Yeah. 😂 

I'll get it merged and resolved, since I stepped on you a little bit, and see how everything is playing out after that.

JamesTerwilliger:
> Showing 38 changed files with 935 additions and 305 deletions.

MotherOfGod.jpg

bradwilson:
I've got a flaky test situation going on. I'm going to push the merge but I'm still looking through things to see if I can figure out where the problem is.

JamesTerwilliger:
By any chance is the flaky test MessageBusTests.QueueReturnsFalseForFailIfStopOnFailTrue?

bradwilson:
It's definitely somehow related to caching, and I didn't just introduce it now, because running back a couple steps into your history was broken.

I had to make some surgical changes because it wouldn't fail just run by itself (being cache related) so I updated _TestFailed to print its messages and tweaked the failing tests a little so I could see them, and this is what I'm seeing:

image

There are three tests that I see that fail, and which one fails is a bit random because of the randomization of test runs.

The classes under test look like this:

```csharp
[Collection(typeof(CollectionWithClassFixtureCounter))]
class ClassWithCountedFixture : IClassFixture<CountedFixture>
{
	public ClassWithCountedFixture(CountedFixture fixture)
	{
		// 1 == test collection, 2 == test class
		Assert.Equal(2, fixture.Identity);
	}

	[Fact]
	public void TheTest() { }
}
```

```
class CountedFixture
{
	static int counter = 0;

	public CountedFixture() => Identity = ++counter;

	public readonly int Identity;
}
```

I suspect it's because you copied the original test two times, but they're all sharing the same static state. Testing that theory now...

bradwilson:
Yep, that's it. I just need to move a copy of the CountedFixture up into each place where it's used so they each have their own static state and not one shared static state for all 3. :)

JamesTerwilliger:
Oof. Sorry. Thanks for catching that. That would explain why the tests were passing in isolation when I ran them.

It's odd, I had no trouble getting this test to get picked up by the test runner, but I'm not able to get one to show up on a different change I'm making that's also blocked on .Net 7 or higher.

I'll take a crack at an analyzer for this issue next once this gets merged.

bradwilson:
As an aside, I was able to use generic attributes in my tests for all of .NET Framework 4.7.2. and .NET 6. I don't think .NET 8 is necessary for those. All I needed to make sure was that I had access to C# 11. Example:

https://github.com/xunit/xunit/blob/main/src/xunit.v3.core.tests/Sdk/Reflection/ReflectionAssemblyInfoTests.cs

JamesTerwilliger:
Ooh, nice! The other item I was looking at is static interface virtual methods, which I think does require .Net 7 or higher sadly.

bradwilson:
Yeah, there's no need for me to roll back the .NET 8 test support, it'll eventually be needed. I just don't think we need the guards on these tests.

bradwilson:
Well, I ran into a problem.

The tests for `Collection<T>` work fine on .NET 6 as expected. However, they fail on .NET Framework, presumably because it cannot find generic attributes at all, even when using C# 11 to define them. That means ideally we should not offer `Collection<T>` to .NET Framework users; however, it's defined in `xunit.v3.core` which targets `netstandard2.0`, not `net472` and/or `net6.0`.

I'm super reluctant to diverge the code base of `xunit.v3.core`, so I'm going to sit on this for a bit and ponder (and check in the few changes I've already made).

JamesTerwilliger:
Oof. Yeah, that's a problem, and I don't have any immediate ideas either.

bradwilson:
One option is to keep `[Collection(typeof(T))]` and remove `[Collection<T>]`. That doesn't fit the exact request, but it does provide some benefit over just using strings (including the ability to use a collection definition that exists in a different assembly, something which we don't support today because of the cost of scanning the assembly for collection definition classes).

JamesTerwilliger:
> One option is to keep `[Collection(typeof(T))]` and remove `[Collection<T>]`. That doesn't fit the exact request, but it does provide some benefit over just using strings (including the ability to use a collection definition that exists in a different assembly, something which we don't support today because of the cost of scanning the assembly for collection definition classes).

At the very least, we could split the PR in two and check in the `typeof(T)` now while figuring out the other half.

JamesTerwilliger:
Does making `Collection<T>` derive from `Collection(typeof(T))` simplify anything? At the moment, `CollectionAttribute` is sealed, which is w

bradwilson:
Unfortunately that means unsealing `CollectionAttribute`, which isn't an option. We seal attributes that we find via reflection because our reflection abstractions may be running where there are no binaries available (for example, to do test discovery via source code with something like Roslyn or Resharper). If the attribute is sealed, then we know for certain how it's constructed and what's available; if it's not sealed, then we don't know what other people may have done in their derivations.

I'll poke around with reflection in .NET Framework and see if I can understand what's going on, but I'm not super hopeful that I'm going to find a magic way out.

bradwilson:
Although, if we unsealed it and gave it an internal constructor, that would stop people from deriving from it. Let me try that and see what happens.

bradwilson:
Hmm, interesting.

```csharp
using System;
using Xunit;

namespace Empty;

class Attribute1<T> : Attribute { }

[Attribute1<TestClass>]
public class TestClass
{
	[Fact]
	public void TestMethod()
	{
		var attributes = typeof(TestClass).CustomAttributes;

		Assert.Empty(attributes);
	}
}
```

Results in:

.NET Framework:

```
Assert.Empty() Failure: Collection was not empty
Collection: [[Empty.Attribute1`1[[System.__Canon, mscorlib, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089]]()]]
```

.NET 6:

```
Assert.Empty() Failure: Collection was not empty
Collection: [[Empty.Attribute1`1[[Empty.TestClass, empty, Version=1.0.0.0, Culture=neutral, PublicKeyToken=null]]()]]
```

So it's there, but the type is mangled somehow.

JamesTerwilliger:
Reason number 420 why I'll be glad when Framework is finally put out to pasture, that's absolutely maddening. :-(

bradwilson:
[This](https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/proposals/csharp-11.0/generic-attributes) tends to imply that (a) runtime support was required to make generic attributes work, and (b) such runtime support was only added to .NET Core, and not .NET Framework.

image

Not a solid confirmation, but pretty strongly suggestive.

bradwilson:
Oddly enough it works correctly if you use a builtin type (though that of course isn't helpful to us).

.NET Framework:
```Collection: [[Empty.Attribute1`1[[System.Int32, mscorlib, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089]]()]]```

.NET 6:
```Collection: [[Empty.Attribute1`1[[System.Int32, System.Private.CoreLib, Version=6.0.0.0, Culture=neutral, PublicKeyToken=7cec85d7bea7798e]]()]]```

bradwilson:
I'm going to see if I can make this work with reference assemblies. I might be able to multi-target the build of `xunit.v3.core` but only ship the `netstandard2.0` binary in the NuGet package (but include both reference libraries, where the .NET Framework version does not include the generic attributes).

bradwilson:
Local package creation and testing looked good. Intellisense and references looked correct; `CollectionAttribute<T>` was not visible in .NET Framework, and none of the XML docs reference it either. The built binary includes the sole `lib/netstandard2.0` binary when compiling for either .NET Framework 4.7.2 or .NET 6. So... I think maybe this is gonna be good? 🤞🏼

bradwilson:
I think this is ready for your review if you'd like, @JamesTerwilliger. I'll merge if you're satisfied.

JamesTerwilliger:
@bradwilson I'm looking now, but it looks as if the PR builds are failing still (non-windows only?)

bradwilson:
Fixing the non-Windows issue right now.

JamesTerwilliger:
Found and fixed one spelling error, code itself looks good (well, good is relative given all of the ID mangling magic that needed to happen in a couple of places, but the logic looks sound) @bradwilson #ShipIt

bradwilson:
Once the packages are pushed to feedz.io I'll do one last test to ensure things are working as expected. Don't think I should see any surprises here but you never know. 😂

bradwilson:
Woot! Doing its thing!

image

JamesTerwilliger:
@bradwilson Nicely done, sir!

Issue #987, opened 2016-09-26
InvalidOperationException thrown in string.Join running MemberData tests


hughbe:
The following throws an `InvalidOperationException` when running tests, with the stack trace attached
![stacktrace](https://cloud.githubusercontent.com/assets/1275900/18843304/193c3f1c-8410-11e6-8267-ca46b9647454.png)

```
using System.Json;
public static IEnumerable<object[]> TestData()
{
    yield return new object[] { new JsonValue[] { new JsonPrimitive(true) } };
}

[Theory]
[MemberData(nameof(TestData))]
public void Test(JsonValue[] items)
{
    JsonArray array = new JsonArray(items);
}
```

This is a bug in xunit I think, as the error doesn't seem to be coming from my code - there is no mention of `string.Join` anywhere.
The reason why an `InvalidOperationException` is thrown is that `JsonValue` throws this exception when `GetEnumerator` is called.

/cc @stephentoub  FYI

bradwilson:
Are you sure about that? The very top of the stack trace is your code (`JsonValue.cs`).

hughbe:
Right, the exception is thrown from the `GetEnumerator` method in `JsonValue.cs`.

But my issue is that this code is being called from `xunit`, specifically some kind of `string.Join` call

bradwilson:
Can you point to the implementation of `JsonValue`?

hughbe:
https://github.com/dotnet/corefx/blob/master/src/System.Json/src/System/Json/JsonValue.cs

bradwilson:
Curious: Why are you implementing `IEnumerable`, only to throw `InvalidOperationException`?

stephentoub:
> But my issue is that this code is being called from xunit, specifically some kind of string.Join call

Presumably this is by design, and xunit is attempting to serialize any IEnumerable as part of the name of a theory?  It might be nice if it caught exceptions from enumeration and ignored them, assuming this is part of name generation or something similarly non-fatal.  That said, this doesn't seem like a bug to me.

> Curious: Why are you implementing IEnumerable, only to throw InvalidOperationException?

I ported this code, and I'm wondering the same thing ;)  I need to go dig up an old copy of the Silverlight library to see what it was doing here.  This is a straight port of Mono's implementation, which we brought to corefx for compat with code that's using it there.

hughbe:
[...]

PR #2885, opened 2024-02-06
Visualizing execution history docs

### LINKED ISSUE #2484: Test method EventSource for easier diagnostics

adamsitnik:
Docs for #987

Fixes #2484


bradwilson:
Thanks! I'm going to fix up the broken images and run them through PNGOUT for minimized size, plus tweak the writing a little bit.

We can get this merged after both the v2 and v3 code has been merged, at which point I will update the docs to indicate what the minimum versions for this in v2 and v3 are.

bradwilson:
What do you want to do about the docs here @adamsitnik since it seems like Concurrency Visualizer is sketchy? Should we just kill the docs? Or write up something else...?

bradwilson:
I'm going to close this PR on the assumption that this is not useful now that Concurrency Visualizer has confirmed they are not usable in this scenario

Issue #1902, opened 2019-02-08
Documentation: Full API reference


zspitz:
There should be a full API reference on the document site, auto-generated in some way.

I imagine each type should have a page, listing all its members; and it should be possible to hyperlink to each member.

The following issues need to be resolved:

- [ ] Should the documentation be generated from one of the platform-specific assembly sets? Or should Roslyn be used to parse the source code?
- [ ] What existing documentation generator can be used for this?
- [ ] What URL structure is appropriate for the API, starting from the docs site?

ErikSchierboom:
I agree that this would be a huge improvement! xUnit is great, but this is one area where I feel NUnit has the edge.

zspitz:
@ErikSchierboom Do you have any experience producing an API reference? Do you have any thoughts about how this should be done?

bradwilson:
Ideally I'd like:

- Auto-generation of the HTML from the XML doc comments
- Ability to have access to multiple versions

Personally I'm okay with the process of kicking off the generation/pushing to the site is something I do by hand, since I don't release new versions very often. I would not do it for CI builds, only builds I ship to NuGet.

ErikSchierboom:
I don't have any experience with generating docs, but I did find this: https://dotnet.github.io/docfx/

stakx:
DocFX is relatively easy to set up. As an example, here's the website it generates for the xunit/xunit repo if you don't apply any customizations (styling, additional content): https://stakx.github.io/xunit/.

In order to get that result, you'd do the following:

1. Install DocFX, e.g. using `choco install docfx`.

2. Go to your local xunit working directory and run `docfx init -q`.

3. Go into the generated `docfx_project` folder (which you can just rename to `docs`) and edit the `docfx.json` configuration file found there:

   <details>
   <summary>Click here to expand the contents of the `docfx.json` file.</summary>

   ```json
   {
   "metadata": [
      {
      "src": [
         {
         "cwd": "..",
         "files": [
               "src/**.csproj"
         ],
         "exclude": [
               "test/**.csproj"
         ]
         }
      ],
      "dest": "api",
      "disableGitFeatures": false,
      "disableDefaultFilter": false
      }
   ],
   "build": {
      "content": [
      {
         "files": [
         "api/**.yml",
         "api/index.md"
         ]
      },
      {
         "files": [
         "articles/**.md",
         "articles/**/toc.yml",
         "toc.yml",
         "*.md"
         ]
      }
      ],
      "resource": [
      {
[...]

PR #2916, opened 2024-04-20
Full API docs for current version classes

AlanMcBee:
To resolve issue #1902 
Adding DocFX actions to CI workflow after BuildAll target
- Builds metadata by running code through Roslyn
- Builds HTML from metadata
- Builds PDF (this can easily be removed later)

The generated files in the `_site` folder are uploaded as an artifact named "docs"

Notes:
- The only formatting was to get the xUnit.net logo to appear in the DocFX modern template
- Still needed: Integrating this with existing documentation on gh-pages branch


AlanMcBee:
@dotnet-policy-service agree

AlanMcBee:
To see a current example of the generated docs, see the artifact named "Docs" at https://github.com/AlanMcBee/xunit/actions/runs/8761001690

bradwilson:
Thanks for this! I'll take a look soon. 😄

bradwilson:
Sorry for the delay, real life has a way of intervening and stealing away time. 😂 

First thing I'm doing is moving all the doc generation into the build system so I can easily build the docs from the command line without needing GitHub Actions to do it for me. 😄 I'm also working on cutting down what gets generated based on expected usage from 3rd party developers.

I will probably have some questions in a bit, but at the moment I'm just poking around docfx to understand how it all works.

bradwilson:
I've been contemplating what the right plan is for this, given that I only want docs for the releases that go to NuGet. Here's what I'm mulling over, would appreciate feedback.

1. Convert the existing xunit.net site from Jekyll to DocFx. This should simplify integration between the existing content and the new content with a single build step (that can be done in the GitHub Actions step that builds & deploys the site).
2. Add as a CI build step to `xunit/xunit` (`main`/`v2` branches) to generate the **_metadata_** and include that in the artifacts.
3. As a ship step, copy the metadata into a version folder somewhere in the site tree so that DocFx can generate the API **_HTML_** pages appropriately, and then have an API docs landing page (at least for now?) that would be a list of all available documentation versions.

Thoughts?

AlanMcBee:
I have substantial notes/feedback on the discussion thread #2906, but here's the shorter response:

1. Agreed. See my notes in the discussion thread.
2. I'm assuming that idea (to build from main/v2) is intended to get at the earlier versions that you mentioned earlier that you want included in the documentation. I understand clearly that you want all the versions documented. I'm prioritizing that, but in the face of a complete absence of any API info on xunit.net, I thought this would be a good first step. I wanted to get this PR through first just to make sure I was contributing appropriately on the repo. 
3. I hadn't really expected the docs that are generated here to be published. I just wanted to get the xmldocs converted into HTML via DocFX. I suppose it wouldn't be too hard to get it folded into the gh-pages branch and published to xunit.net, but that's not what I expected yet.

The discussion thread has much more of my thoughts.

bradwilson:
Closing this as the new site exists at https://api.xunit.net with source at https://github.com/xunit/api.xunit.net

Issue #2974, opened 2024-07-08
Assert.Equivalent behaves incorrectly with Guid values


k-wojcik:
Similar problem to issue https://github.com/xunit/xunit/issues/2913. Guid values are incorrectly compared.

**To Reproduce**
```csharp
Assert.Equivalent(new Guid("b727762b-a1c1-49a0-b045-59ba97b17b61"), new Guid("963ff9f5-cb83-480e-85ea-7e8950a01f00"));
```
**Actual behavior** 
no exception has been thrown

**Expected behavior** 
throws `EquivalentException`

PR #2975, opened 2024-07-08
Assert.Equivalent behaves incorrectly with Guid values

k-wojcik:
Fixes https://github.com/xunit/xunit/issues/2974


k-wojcik:
@dotnet-policy-service agree

bradwilson:
Thanks! I've ported the test to v2 as well and pushed it. 😄

PR #2976, opened 2024-07-10
Fix tests failing depending on OS culture

MaxKot:
`Assert.NotInRange` tests verifying error messages were updated to match `Assert.InRange` tests. `CommandLine.MultiplierValue` test was updated because `-maxthreads` parsing is intended to be culture-independent, now test should ensure it is.


bradwilson:
Thanks!

PR #2984, opened 2024-07-29
Update/fix FSharp SupportedOSAttribute project template

kojo12228:
In response to https://elk.zone/mastodon.social/@bradwilson/112865921926948375:

- Use Map.tryFind rather than IDictionary method
- Fix setting mutable value
- Avoid constructing Exception objects by using `invalidArg` and `failwith`

For lightly more idiomatic FSharp I might have replaced the `for` loop with a tail recursive function, but exceptions are being used in control flow I don't think it makes much difference.

### DIFF
--- a/src/xunit.v3.templates/templates/project/xunit3-extension/fs/SupportedOSAttribute.fs
+++ b/src/xunit.v3.templates/templates/project/xunit3-extension/fs/SupportedOSAttribute.fs
@@ -37,11 +37,14 @@ type SupportedOSAttribute([<ParamArray>]supportedOSes: SupportedOS[]) =
 
         for supportedOS in supportedOSes do
             let osPlatform =
-                match osMappings.TryGetValue supportedOS with
-                | true, value -> value
-                | _ -> raise (new ArgumentException($"Supported OS value '{supportedOS}' is not a known OS", nameof(supportedOSes)))
+                match Map.tryFind supportedOS osMappings with
+                | Some value -> value
+                | None -> raise (new ArgumentException($"Supported OS value '{supportedOS}' is not a known OS", nameof(supportedOSes)))
 
-            if RuntimeInformation.IsOSPlatform osPlatform then matched = true |> ignore else () 
+            if RuntimeInformation.IsOSPlatform osPlatform then
+                matched <- true
+            else
+                () 
 
             if not matched then
                 // We use the dynamic skip exception message pattern to turn this into a skipped test
--- a/src/xunit.v3.templates/templates/project/xunit3-extension/fs/SupportedOSAttribute.fs
+++ b/src/xunit.v3.templates/templates/project/xunit3-extension/fs/SupportedOSAttribute.fs
@@ -39,7 +39,7 @@ type SupportedOSAttribute([<ParamArray>]supportedOSes: SupportedOS[]) =
             let osPlatform =
                 match Map.tryFind supportedOS osMappings with
                 | Some value -> value
-                | None -> raise (new ArgumentException($"Supported OS value '{supportedOS}' is not a known OS", nameof(supportedOSes)))
+                | None -> invalidArg (nameof supportedOSes) $"Supported OS value '{supportedOS}' is not a known OS"
 
             if RuntimeInformation.IsOSPlatform osPlatform then
                 matched <- true
@@ -49,7 +49,7 @@ type SupportedOSAttribute([<ParamArray>]supportedOSes: SupportedOS[]) =
             if not matched then
                 // We use the dynamic skip exception message pattern to turn this into a skipped test
                 // when it's not running on one of the targeted OSes
-                raise (new Exception($"$XunitDynamicSkip$This test is not supported on {RuntimeInformation.OSDescription}"))
+                failwith $"$XunitDynamicSkip$This test is not supported on {RuntimeInformation.OSDescription}"
             else ()
 
         Unchecked.defaultof<ValueTask>


bradwilson:
Thanks! 😄

src/xunit.v3.templates/templates/project/xunit3-extension/fs/SupportedOSAttribute.fs
```diff
@@ -37,16 +37,19 @@ type SupportedOSAttribute([<ParamArray>]supportedOSes: SupportedOS[]) =
 
         for supportedOS in supportedOSes do
             let osPlatform =
-                match osMappings.TryGetValue supportedOS with
-                | true, value -> value
-                | _ -> raise (new ArgumentException($"Supported OS value '{supportedOS}' is not a known OS", nameof(supportedOSes)))
+                match Map.tryFind supportedOS osMappings with
```
bartelink, line 40: very rare to see this unpiped, i.e. this is more common:
```
match osMappings |> Map.tryFind supportedOS with
```

src/xunit.v3.templates/templates/project/xunit3-extension/fs/SupportedOSAttribute.fs
```diff
@@ -37,16 +37,19 @@ type SupportedOSAttribute([<ParamArray>]supportedOSes: SupportedOS[]) =
...
+                | Some value -> value
+                | None -> invalidArg (nameof supportedOSes) $"Supported OS value '{supportedOS}' is not a known OS"
```
bartelink, line 42: for argumentexception, the `name` should refer to the argument with the bad value, i.e.

| None -> invalidArg (nameof supportedOS) $"Supported OS value '{supportedOS}' is not a known OS"

src/xunit.v3.templates/templates/project/xunit3-extension/fs/SupportedOSAttribute.fs
```diff
@@ -37,16 +37,19 @@ type SupportedOSAttribute([<ParamArray>]supportedOSes: SupportedOS[]) =
...
 
-            if RuntimeInformation.IsOSPlatform osPlatform then matched = true |> ignore else () 
+            if RuntimeInformation.IsOSPlatform osPlatform then
+                matched <- true
+            else
+                () 
 
             if not matched then
```
bartelink, line 49: this check needs to be unindented to make any kind of sense - as it stands it will throw on the first iteration unless L45 runs. My guess it needs to become:
```
        let mutable matched = false

        for supportedOS in supportedOSes do
            match osMappings |> Map.tryFind supportedOS with
            | None -> invalidArg (nameof supportedOSes) $"Supported OS value '{supportedOS}' is not a known OS"
            | Some osPlatform ->
            
            if RuntimeInformation.IsOSPlatform osPlatform then
                matched <- true

          if not matched then
              // We use the dynamic skip exception message pattern to turn this into a skipped test
              // when it's not running on one of the targeted OSes
              failwith $"$XunitDynamicSkip$This test is not supported on {RuntimeInformation.OSDescription}"
```
NOTE I have unindented the `if not matched` line

If the goal is not intuitive understandability to someone coming from C#, I'd write:

```
   let targetPlatform (x: SupportedOS) =
        match osMappings |> Map.tryFind x with
        | None -> invalidArg (nameof supportedOSes) $"Supported OS value '{supportedOS}' is not a known OS"
        | Some target -> target
   let supportedHere = supportedOses |> Seq.map targetPlatform |> Seq.exists RuntimeInformation.IsOSPlatform 
   if not supportedHere then
       // We use the dynamic skip exception message pattern to turn this into a skipped test
       // when it's not running on one of the targeted OSes
       failwith $"$XunitDynamicSkip$This test is not supported on {RuntimeInformation.OSDescription}"
```

bradwilson: I noticed the indentation error after merging this PR and have already fixed it.

bradwilson: You can't move validation to the constructor of an attribute because of the way attributes are constructed. It's an oddity of the way attributes work.

bradwilson: I like this iteration, and in my head I can see the C# LINQ version that this represents.

bradwilson: Thanks, I've integrated your feedback: https://github.com/xunit/xunit/commit/69176a445eef78fc0098b8a16579aefb257cd651

bartelink: Thanks!

>You can't move validation to the constructor of an attribute because of the way attributes are constructed. It's an oddity of the way attributes work.

I kinda knew that; I was more trying to convey that I'd want to have the logic about the exception from parsing be separated out for clarity. Something like:

```
type SupportedOs with
    member x.ToOsPlatform() =
        match x with
        | SupportedOS.FreeBSD ->   OSPlatform.Create("FreeBSD")
        | SupportedOS.Linux ->     OSPlatform.Linux
        | SupportedOS.macOS ->     OSPlatform.OSX
        | SupportedOS.Windows ->   OSPlatform.Windows
        // TOCONSIDER isnt there another ctor that takes the value too, and then use a generic message?
        | x ->                     raise <| System.ArgumentOutOfRangeException(nameof x, $"Supported OS value '{x}' is not a known OS")

type SupportedOSAttribute([<ParamArray>]supportedOSes: SupportedOS[]) =
    inherit BeforeAfterTestAttribute()
    let supportedPlatforms = supportedOses |> Seq.map _.ToOsPlatform 
    override _.Before(_: MethodInfo, _: IXunitTest) =
        let canRunHere = supportedPlatforms |> Seq.exists RuntimeInformation.IsOSPlatform
        if not canRunHere then
           // We use the dynamic skip exception message pattern to turn this into a skipped test
           // when it's not running on one of the targeted OSes
           failwith $"$XunitDynamicSkip$This test is not supported on {RuntimeInformation.OSDescription}"

        Unchecked.defaultof<ValueTask>
```

Or can wrap it into the enum type:
```
type SupportedOS =
    | FreeBSD = 1
    | Linux = 2
    | macOS = 3
    | Windows = 4
    member x.ToOsPlatform() =
        match x with
        | SupportedOS.FreeBSD ->   OSPlatform.Create("FreeBSD")
        | SupportedOS.Linux ->     OSPlatform.Linux
        | SupportedOS.macOS ->     OSPlatform.OSX
        | SupportedOS.Windows ->   OSPlatform.Windows
        | x ->                     raise <| System.ArgumentOutOfRangeException(nameof x, $"Supported OS value '{x}' is not a known OS")
```
which would change 
> `let supportedPlatforms = supportedOses |> Seq.map SupportedOs.toOsPlatform`

to `let supportedPlatforms = supportedOses |> Seq.map _.ToOsPlatform`

Anyway, this is all subjective so will leave it up to you - there's definitely value in the three versions following the same outline and I'm already happy as it is ;)

kojo12228: @bradwilson Thanks for the comments. I did unindent the change and failed to push the commit before running off to work!

src/xunit.v3.templates/templates/project/xunit3-extension/fs/SupportedOSAttribute.fs
kojo12228: It's pretty common to see an unpiped `Map.tryFind` when used in a match statement: https://github.com/search?q=Map.tryFind+language%3AF%23&type=code&l=F%23

bartelink: It's a fair cop - I was using a _little_ poetic license to be fair!
(I am actually normally pretty anti egregious usage of piping, unless it truly makes the code more legible - I found it unintuitive as it stood, but that may be as a result of codebases I've worked in over the last while tending to using piping in such cases happening to arbitrarily standardise on that)
I wonder if Fantomas hasa rule/suggestion/preference on this [within the context of a `match`[?)

PR #3012, opened 2024-08-26
Fix broken template source link in xunit.v3.templates.nuspec

timabell:
The repo link on https://www.nuget.org/packages/xunit.v3.templates/0.2.0-pre.69 is broken (404), this patch changes the link to the current source location

I'm not going to be signing any CLA, so you may have to recreate this patch without this PR.


timabell:
@dotnet-policy-service decline

bradwilson:
Yeah, that link shouldn't have changed, but I don't think the intention of that NuGet field is to point to an arbitrary URL. The git URL is still just `https://github.com/xunit/xunit`.

bradwilson:
Given the confusion from the CLA bot (which should not be triggering), I'm just going to close this and fix it myself. Thanks for letting me know it was broken!

Issue #1742, opened 2018-06-14
InlineData and implicit conversions


FantasticFiasco:
## Expected Behavior

1. The ReSharper test runner and the VSTest runner should not hang.
1. `dotnet test` should pass if `dotnet xunit` pass

## Actual Behavior

1. When running the test defined below the runners hang. One has to restart Visual Studio or JetBrains Rider in order to run the test again.
1. `dotnet xunit` will successfully run the test while `dotnet test` will report the following error, which actually doesn't describe the real problem, i.e. that the implicit conversion for some reason cannot be handled.

```
[xUnit.net 00:00:00.5745340] Exception discovering tests from vstest-error: System.ArgumentException: There is at least one object in this array that cannot be serialized
Parameter name: array
   at Xunit.Serialization.XunitSerializationInfo.ArraySerializer..ctor(Array array) in C:\Dev\xunit\xunit\src\common\XunitSerializationInfo.cs:line 417
   at Xunit.Sdk.SerializationHelper.Serialize(Object value) in
```

## Steps to Reproduce the Problem

1. Create a new xUnit project using the command `dotnet new xunit`
2. Paste the following code in the created project:
```
public class UnitTest1
{
    [Theory]
    [InlineData("someId")]
    public void Test1(Id id)
    {
    }
}
    
public class Id
{
    public Id(string value)
    {
        Value = value;
    }

    public string Value { get; }
        
    public static implicit operator Id(string id)
    {
        return id != null ? new Id(id) : null;
    }
}
```
3. Run the test using the ReSharper test runner or VSTest runner. You will see that the runner hangs.
4. Run the test using `dotnet test`. You will see the cryptic error message
5. Run the test using `dotnet xunit`. You will see that the test pass

## Versions

- Microsoft.NET.Test.Sdk: 15.7.0
- xunit: 2.3.1
- xunit.runner.visualstudio: 2.3.1
- dotnet-xunit: 2.3.1

JoshuaLight:
Hey, guys!

I've found similar behaviour in some tests of mine, so I decided to investigate this.

## Description

It looks like on *xUnit* **2.4.0.4010** this works well, at least Resharper isn't hanging. The error is the same as reported by @FantasticFiasco.

But if I take **2.3.1** — everything will freeze. And I actually couldn't be able to start any other tests before shutting down `dotnet` process through *Task Manager*.

I've also found that there are other cases when ReSharper hangs, for example:
```c#
[Theory]
[...]

PR #3067, opened 2024-11-27
#1742 Improved the error reporting when the DefaultTestCaseDescriptorProvider fails.

koenigst:
Calling `UniqueID` in the `DefaultTestCaseDescriptorProvider` caused an exception when there were test method arguments that could not be serialized. The exception was silently dropped. This PR tries to improve error reporting in this case.

This PR targets `v2` because the same bug does not happen in v3 (the exception happens earlier and the fallback is theory enumeration during execution).

### DIFF
--- a/src/xunit.runner.utility/Descriptor/DefaultTestCaseDescriptorProvider.cs
+++ b/src/xunit.runner.utility/Descriptor/DefaultTestCaseDescriptorProvider.cs
@@ -1,5 +1,6 @@
 using System;
 using System.Collections.Generic;
+using System.Diagnostics;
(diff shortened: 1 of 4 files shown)


bradwilson:
I'm going to leave this PR open, if you don't mind, because I'm not ready to evaluate it since I'm not sure there will be another v2 release. I'm not necessarily convinced that skipping is the right behavior here, and I'd also like to understand how a test case showed up with non-serializable data in the first place this far down the pipeline.

bradwilson:
Closing, as there are no more v2 releases planned. Thanks for the PR!

PR #3142, opened 2025-01-21
In .NET 7+, call `IParsable<T>.Parse` via a wrapper method

### LINKED ISSUE #3141: Error when implementing `IParsable<T>` via a wrapper interface
If `IParsable<T>` is implemented through another interface as shown below, the following error occurs:

> Could not find Parse method for IParsable<ParsableClass>

```csharp
interface IParsableWrapper<T> : IFormattable, IParsable<T> where T : IParsable<T>
{
	public string Value { get; }

	static T IParsable<T>.Parse(string s, IFormatProvider? provider)
	{
		if (!T.TryParse(s, provider, out var result))
			throw new InvalidOperationException();
		return result;
	}

	string IFormattable.ToString(string? format, IFormatProvider? formatProvider) =>
		Value;
}
class ParsableClass(string value) : IFormattable, IParsableWrapper<ParsableClass>
{
	public string Value => value;

	public static bool TryParse([NotNullWhen(true)] string? s, IFormatProvider? provider, [MaybeNullWhen(false)] out ParsableClass result)
	{
		if (s is null)
		{
			result = null;
			return false;
		}

		result = new(s);
		return true;
	}
}
```

kzrnm:
Fix #3141 

Added .net7.0 to TargetFrameworks in xunit.v3.common and created a wrapper method to call `IParsable<T>.Parse`, which seems to resolve the issue.


bradwilson:
We cannot add new target frameworks. This bug must be fixed another way.

Issue #3159, opened 2025-02-07
Consider lowering the `System.Memory` dependency in `xunit.v3.assert` from 4.6.0 to 4.5.5


eiriktsarpalis:
I recently tried upgrading one of personal projects to use xunit.v3 but encountered issues because the new version requires `System.Memory` version 4.6.0:

https://github.com/xunit/xunit/blob/7ad077d8526cb36497a764422ea9b16ed75813d4/src/Versions.props#L19

This is creating problems in repos that implement source generators and have been configured to use central package management: the VS host process is still on `System.Memory` v4.5.5 which is causing the source generators to crash when run on the IDE. I can of course work around the issue by setting a `VersionOverride` on the impacted source generator projects, however this is creating assembly binding issues in netfx test projects that reference the source generator.

While not an absolute blocker, it's still an annoyance for source gen projects looking to migrate to xunit.v3.

eiriktsarpalis:
Secondarily, it would be nice if the `System.Collections.Immutable` dependency got lowered to 7.0.0. While it's not creating issues for anybody working with the latest versions of VS 2022, older still supported versions are on 7.0.0 still:

https://www.nuget.org/packages/Microsoft.CodeAnalysis.Common/4.8.0#dependencies-body-tab

bradwilson:
FWIW, we are about to flip to .NET 8 minimum requirements, so these version upgrades were chosen in anticipation of that change. Even if we were to issue a new release with lowered dependencies, the change would be short lived.

I don't have much experience with source generators, so can you explain to me how the dependencies of our assertion library impact source generation?

eiriktsarpalis:
I use central package management cross-repo, which means adding xunit.v3 bumps the System.Memory cross-repo to 4.6.0. Source generator projects have a transitive dependency on System.Memory via Roslyn, but taking a dependency above 4.5.5 renders them inoperable when you try to host them on the latest VS (the VS host process uses netfx and System.Memory 4.5.5). There are a number of ways I could work around the issue, but all of them introduce hassle significant enough for me to put off a migration to v3.

> FWIW, we are about to flip to .NET 8 minimum requirements, so these version upgrades were chosen in anticipation of that change. Even if we were to issue a new release with lowered dependencies, the change would be short lived.

[...]

PR #3160, opened 2025-02-07
lower xunit-assert dependencies

eiriktsarpalis:
For reference, this attempts to address #3159.


bradwilson:
Merged along with some other unrelated changes. Thanks for the PR!

Issue #3178, opened 2025-02-15
Fixtures can not request other fixtures


neistow:
Currently, the fixture requiring another fixture doesn't work, throwing an exception.

```
Xunit.Sdk.TestPipelineException
Class fixture type 'Playground.Tests.FixtureB' had one or more unresolved constructor arguments: FixtureA fixtureA
   at Xunit.v3.FixtureMappingManager.GetFixture(Type fixtureType) in /_/src/xunit.v3.core/Utility/FixtureMappingManager.cs:line 164
   at Xunit.v3.FixtureMappingManager.InitializeAsync(IReadOnlyCollection`1 fixtureTypes) in /_/src/xunit.v3.core/Utility/FixtureMappingManager.cs:line 224
```

```cs
public class FixtureA
{
    public string A { get; private set; }
    
    public FixtureA()
    {
        A = "Can";
    }
}

public class FixtureB : IClassFixture<FixtureA>
{
    public string B { get; private set; }
    
    public FixtureB(FixtureA fixtureA)
    {
        B = fixtureA.A + " we make this work?";
    }
}

public class MyTests : IClassFixture<FixtureB>
{
    private readonly FixtureB _fixtureB;

    public MyTests(FixtureB fixtureB)
    {
        _fixtureB = fixtureB;
    }

    [Fact]
    public void Test()
    {
        Assert.NotEmpty(_fixtureB.B);
    }
}
```

I'm trying to share and compose some functionality used across tests. The only sketchy way I found of doing this is using `TestContext.Current.GetFixture`, but this heavily relies on interface implementation order and just doesn't feel *right*.

```cs
public class FixtureA
{
    public string A { get; private set; }

    public FixtureA()
    {
        A = "Can";
    }
}

public class FixtureB : IAsyncLifetime
{
    public string B { get; private set; } = null!;

    public FixtureB()
    {
    }

    public async ValueTask InitializeAsync()
    {
        var fixtureA = await TestContext.Current.GetFixture<FixtureA>();
        ArgumentNullException.ThrowIfNull(fixtureA);

        B = fixtureA.A + " we make this work?";
    }

    public ValueTask DisposeAsync() => ValueTask.CompletedTask;
}

// will work ONLY if IClassFixture<FixtureA> is implemented before IClassFixture<FixtureB>
public class MyTests : IClassFixture<FixtureA>, IClassFixture<FixtureB>
{
    private readonly FixtureB _fixtureB;

    public MyTests(FixtureB fixtureB)
[...]

Issue #1738, opened 2018-06-11
Add support for dependency injection


gitfool:
We've discussed this previously in #687 but perhaps it could be reconsidered for 3.0? 😉 

I'll open with a comment I made at the time that's still relevant today:
> We like xUnit to the degree that it made sense to leverage its familiar framework and runners to also drive our integration tests. Both Devs & QA love the consistency. As with unit tests we can drive them from within Visual Studio, with Resharper, and the commandline. This has been invaluable, but is obviously where we heavily use DI.

I'd like support for DI to be officially baked in and based on past hacking that involves the following:

- Assembly fixtures
  - e.g. [kzu/xunit.assemblyfixture](https://github.com/kzu/xunit.assemblyfixture)
- Global fixture(s)
  - i.e. across all loaded assemblies (when multiple specified)
- Predefined lifetime scopes
  - per test (transient), class, collection, assembly, global (singleton)
  - ... as per fixtures but with first class support these could replace fixtures
- Container bootstrapping
  - xunit type registration; e.g. `ITestOutputHelper` per test
  - user type registration hooks

I'm most familiar with [Autofac](https://www.fuget.org/packages/Autofac), but this task would presumably have to be container agnostic, although [Microsoft.Extensions.DependencyInjection](https://www.fuget.org/packages/Microsoft.Extensions.DependencyInjection) might be an option now.

tai-yi:
As a dotnetcore developer, I prefer  `Microsoft.Extensions.DependencyInjection`.

bartelink:
I know I could just thumbs down, but as a dotnetcore developer, I do not prefer `Microsoft.Extensions.DependencyInjection` and would hate to see a dependency on it. (I trust others, including Brad, will have similar opinions)

tai-yi:
Developers have different love for DI container.
It's reasonable to reduce dependency, so the better way is setup an interface such as below code.
Then we can write different XunitStartupPlugins.

``` c#
    public interface IXunitStartup<T, TScope>
        where TScope : IDisposable
    {
        //Root DI container
        T ServiceProvider { get; }

        TScope CreateScope(T sp);

        TServive GetService<TServive>();

        TServive GetService<TServive>(TScope scope);

        IEnumerable<TService> GetServices<TService>();

        IEnumerable<TService> GetServices<TService>(TScope scope);

    }
    public class XunitStartupAttribute : Attribute { }

    [XunitStartup]
[...]

PR #3181, opened 2025-02-16
#3178 Composable fixtures

neistow:
This is POC implementation for issue #3178.

Fixtures can now implement the `IClassFixture` interface specifying other fixtures to depend on. This will allow logic sharing between fixtures without resorting to inheritance or manual fixture instantiation.

Known issues:

- [Analyzer 1041](https://xunit.net/xunit.analyzers/rules/xUnit1041) needs to be updated to scan all composable fixture hierarchy and allow omitting explicit `IClassFixture` implementation for dependent fixtures when injecting those in the test constructor.
- Composable fixtures can have circular `IClassFixture` interface implementations if one of them doesn't specify the other in the constructor.

Open questions:

- Do we want to use the `IClassFixture` interface for this feature? Or perhaps create a separate interface for this purpose?

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -408,4 +408,5 @@ FodyWeavers.xsd
 *.msp
 
 # JetBrains Rider
+.idea
 *.sln.iml
--- a/src/xunit.v3.core.tests/Acceptance/FixtureAcceptanceTests.cs
+++ b/src/xunit.v3.core.tests/Acceptance/FixtureAcceptanceTests.cs
@@ -316,6 +316,83 @@ public class FixtureAcceptanceTests
 
 			public IMessageSink MessageSink { get; }
 		}
+
+		[Fact]
+		public async ValueTask ClassFixtureCanAcceptComposableFixture()
+		{
+			var messages = await RunAsync(typeof(ClassWithComposableFixtures));
+
+			Assert.Single(messages.OfType<ITestPassed>());
+		}
+
+		class ClassWithComposableFixtures : IClassFixture<ComposableFixtureC>
+		{
+			readonly ComposableFixtureC fixtureC;
+
+			public ClassWithComposableFixtures(ComposableFixtureC fixtureC)
+			{
+				this.fixtureC = fixtureC;
+			}
+
+			[Fact]
+			public void ComposableFixtureWasInjected() => Assert.NotNull(fixtureC);
+		}
+
+		class ComposableFixtureA;
+
+		class ComposableFixtureB : IClassFixture<ComposableFixtureA>
+		{
+			public ComposableFixtureB(ComposableFixtureA fixture)
+			{
+			}
+		}
+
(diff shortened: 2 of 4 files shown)


neistow:
@dotnet-policy-service agree

neistow:
I Ran `BuildAll` locally, and everything succeeded, should be good now.

bradwilson:
> * Do we want to use the `IClassFixture` interface for this feature?

I'm torn on this, for a few reasons.

1. It's not just class fixtures that would want this feature.

1. It seems to be a misuse of the way we designed that interface, which is to decorate the test class, not the fixture data class. Worse, what happens if you throw `ICollectionFixture<T>` on a dependency?

1. Related to the above, how can you mix the lifetimes of your dependencies depending on usage? Would it not be possible to have a class fixture want its dependency resolved as a class fixture in one scenario, but a collection fixture in another scenario?

1. Related to the two above, what if we do away with decorating the dependencies? Today you can take a dependency from levels above you (e.g., class fixtures can take collection or assembly fixture dependencies). Imagine a class fixture which takes a dependency, assuming it's coming from a level above (like a collection fixture), but it's used on a test class which is not in a collection which is decorated with the fixture interface? Today that code fails (because the dependency is missing), but if we go down the path of non-decorated dependencies that aren't found elsewhere just assumed to be at the same level (so a class fixture dependency is by default assumed to be another class fixture, unless it's found somewhere at a level above), we end up in a situation where code that previously alerted you to misuse no longer alerts you, and may instead cause unexpected failures (due to instances not being shared at the right level).

So, I'm not yet 100% sure what this feature should look like, but I think I've talked myself out of (a) reusing `IXyzFixture<T>`, or (b) going decoration-less and just assuming anything is a fixture at the same level. I don't know what this leaves: should it be a parameter attribute on the constructor which says "it's okay if you don't know this, you can make it at my level"?

This should also be part of the discussion, IMO, at least inasmuch as some developers would like for us to support property injection in addition to constructor injection:

* https://github.com/xunit/xunit/issues/1738

bradwilson:
Just to let you know... I've looked at this several times since the PR was opened and I continue to be torn on the best design. I'm still hoping a background thread in my brain will pop something out. This is still on my list of things to consider.

neistow:
[This](https://tunit.dev/docs/examples/complex-test-infrastructure#real-world-example-full-stack-integration-testing) is the kind of composition that fixtures in xUnit are missing

metacoder-feff:
@neistow @bradwilson

Check out this [add-on for composable fixtures](https://github.com/metacoder-feff/FEFF.TestFixtures). It does not require making any changes to xUnit.
[The Discussion](https://github.com/xunit/xunit/discussions/3513)
[This patch improves the integration](https://github.com/xunit/xunit/issues/3528#issuecomment-4337318115)

PR #3187, opened 2025-02-23
Cleanup xunit.v3.core.targets

Youssef1313:
I don't think hooking into `_IncludeGenerateAutoRegisteredExtensionsIntoCompilation` is necessary. Also cleaning multiple language checks and use `DefaultLanguageSourceExtension` instead.


bradwilson:
This has all changed in the `rel/v3-2.0.0` branch which is where I believe the next release comes from, so I don't think this is relevant any more.

bradwilson:
Or at least the majority of this :)

Youssef1313:
@bradwilson It looks like some changes are still relevant for the branch you mentioned

https://github.com/xunit/xunit/blob/27604f0c47552409256f9569e6ff7a8fcae379d4/src/xunit.v3.core/Package/buildTransitive/xunit.v3.core.targets#L77-L83

I'll open another PR against rel/v3-2.0.0

bradwilson:
I just did the merge from 2.0.0 into main, so this should be able to cleaned up and still done against main. Sorry for the inconvenience.

Thanks!

bradwilson:
(hopefully this message catches you before you start work against the branch I just deleted 😁)

Youssef1313:
Thanks @bradwilson. I made few more changes. I assume you have test coverage for F# that verify my changes?

bradwilson:
I don't have unit test coverage for F#/VB, whenever I touch these things I tend to build the packages locally and then test the templates locally, which is what I'll do before merging. I haven't come up with a good way to automate the template testing.

bradwilson:
The previous version was working, I expect because of my dependency. The switch to CompileAfter makes sense, though, to be less fragile.

bradwilson:
Yep, everything tested fine for all three languages:

image

I suppose I could just do something like this in the CI build script. 🤔

bradwilson:
Thanks!

Youssef1313:
> The previous version was working, I expect because of the dependency that was just removed. The switch to CompileAfter makes sense, though, to be less fragile.

Yup. We recently did a similar change on MTP side https://github.com/microsoft/testfx/pull/5095.

PR #3226, opened 2025-03-18
Reference .NET built-in packages conditionally

carlreinke:
The "System.Collections.Immutable" and "System.Memory" packages are built-in to .NET 8, so referencing them when targeting .NET 8 just gives users unnecessary transitive package references.


bradwilson:
Is the intention here to change the package? Because none of this PR changes the packages.

carlreinke:
Oops.  I've been letting the SDK generate my .nuspecs for so long I forgot there was any other way... 😓

The packages should be good now.

bradwilson:
Don't worry about the failing CI builds, those are unrelated (and just fixed in `main`).

Thanks!

PR #3231, opened 2025-03-21
Docs should use  `--filter-query` instead of `--query-filter`

benmccallum:
This must be a recent development, as comments on issues also use the `--query-filter`, but in the logs of a failed run it clearly shows all the available switches as `--filter-xyz`, so I imagine this was just normalized at some point.

Will do a scan elsewhere too


bradwilson:
Thanks, this is definitely a typo on my part.

However, the English language shouldn't change: the noun is a "filter query". So I reverted the wording change.

Issue #3243, opened 2025-03-28
Calls to `IXunitSerializer` are inconsistent with Array values between `SerializationHelper.IsSerializable` and `SerializationHelper.Serialize`.


Chryssie:
In `SerializationHelper.Serialize(object? value, System.Type? type)` when the `type` is an array then the element type is unwrapped (Line 535) and passed to the `IXunitSerializer.IsSerializable(Type type, object? value, [NotNullWhen(true)]out string? failureMessage)` method (Line 561).

https://github.com/xunit/xunit/blob/3c10d5b009e6142c722364a04d251a29c678d25f/src/xunit.v3.common/Serialization/SerializationHelper.cs#L534-L562

However, from the the `SerializationHelper.IsSerializable(object? value, System.Type? type)` when the `type` is an array it will recursively call the `SerializationHelper.IsSerializable(object? value, System.Type? type)` method with the element type with a `null` value to ensure that the array element type is serializable.

https://github.com/xunit/xunit/blob/3c10d5b009e6142c722364a04d251a29c678d25f/src/xunit.v3.common/Serialization/SerializationHelper.cs#L498-L507

The specific inconsistency is that in the first scenario if you, for example, registered a serializer for `KeyValuePair<int, object>` and then call either of the aformentioned `SerializationHelper` methods with an instance of the type `KeyValuePair<int, object>[]` then `IXunitSerializer.IsSerializable(Type type, object? value, [NotNullWhen(true)]out string? failureMessage)` would be called with `type = typeof(KeyValuePair<int, object>)` in both cases, but the first case is called with `value.GetType() == typeof(KeyValuePair<int, object>[])` whereas the second method is called with `value == null`.

bradwilson:
I'm trying to follow the issue, can you provide a code example of the `KeyValuePair<int, object>` that you're trying to serialize that is causing the issue?

Thanks!

Chryssie:
The specific scenario I was using involves serialization of a generic `KeyValuePair<,>` type so it had a lot of code. However, I can reproduce the issue with a simpler, non-generic, case which I have include below.

`.csproj`
```xml
<Project Sdk="Microsoft.NET.Sdk">

  <PropertyGroup>
    <TargetFramework>net8.0</TargetFramework>
    <ImplicitUsings>enable</ImplicitUsings>
    <Nullable>enable</Nullable>

    <IsPackable>false</IsPackable>
    <IsTestProject>true</IsTestProject>
[...]

PR #3254, opened 2025-04-05
Ensure types for calls to `IXunitSerializer.IsSerializable` are consistent.

Chryssie:
Resolves: #3243


bradwilson:
This is already working as documented. I don't believe this PR is appropriate.

PR #3272, opened 2025-04-23
Provide source information via CallerFilePath and CallerLineNumber

Youssef1313:
- This is a quick PoC that has breaking changes.
- It doesn't fully build (there are tests that need to be updated to account for the breaking changes).
- Not yet tested.

@bradwilson This PR is my *personal* view on how source information should be collected. What do you think? Is it something you can consider for xunit.v3 3.x?

### DIFF
--- a/src/xunit.v3.core/Framework/FactDiscoverer.cs
+++ b/src/xunit.v3.core/Framework/FactDiscoverer.cs
@@ -40,6 +40,8 @@ public class FactDiscoverer : IXunitTestCaseDiscoverer
 			details.SkipUnless,
 			details.SkipWhen,
 			testMethod.Traits.ToReadWrite(StringComparer.OrdinalIgnoreCase),
+			sourceFilePath: factAttribute.SourceFilePath,
+			sourceLineNumber: factAttribute.SourceLineNumber,
 			timeout: details.Timeout
 		);
 	}
@@ -84,11 +86,15 @@ public class FactDiscoverer : IXunitTestCaseDiscoverer
 
 		var details = TestIntrospectionHelper.GetTestCaseDetails(discoveryOptions, testMethod, factAttribute);
 
-		return new(
+		var errorTestCase = new ExecutionErrorTestCase(
 			details.ResolvedTestMethod,
 			details.TestCaseDisplayName,
 			details.UniqueID,
 			message
 		);
+
+		errorTestCase.SourceFilePath = factAttribute.SourceFilePath;
+		errorTestCase.SourceLineNumber = factAttribute.SourceLineNumber;
+		return errorTestCase;
 	}
 }
--- a/src/xunit.v3.core/Abstractions/Attributes/IFactAttribute.cs
+++ b/src/xunit.v3.core/Abstractions/Attributes/IFactAttribute.cs
@@ -1,4 +1,5 @@
 using System;
+using System.Runtime.CompilerServices;
 using Xunit.Sdk;
 
 namespace Xunit.v3;
@@ -84,4 +85,14 @@ public interface IFactAttribute
 	/// completely).
 	/// </remarks>
 	int Timeout { get; }
+
+	/// <summary>
+	/// Gets the source file path declaring the attribute. Usually provided via <see cref="CallerFilePathAttribute"/>.
+	/// </summary>
+	string? SourceFilePath { get; }
+
+	/// <summary>
+	/// Gets the source file line number declaring the attribute. Usually provided via <see cref="CallerLineNumberAttribute"/>.
+	/// </summary>
+	int SourceLineNumber { get; }
 }
--- a/src/xunit.v3.core/FactAttribute.cs
+++ b/src/xunit.v3.core/FactAttribute.cs
@@ -1,6 +1,7 @@
 #pragma warning disable CA1813 // This attribute is unsealed because it's an extensibility point
 
 using System;
+using System.Runtime.CompilerServices;
 using Xunit.v3;
 
 namespace Xunit;
@@ -13,6 +14,15 @@ namespace Xunit;
 [AttributeUsage(AttributeTargets.Method, AllowMultiple = false)]
 public class FactAttribute : Attribute, IFactAttribute
 {
+	/// <summary>
+	/// Createa a new instance of <see cref="FactAttribute"/> which provides source information.
+	/// </summary>
+	public FactAttribute([CallerFilePath] string? sourceFilePath = null, [CallerLineNumber] int sourceLineNumber = -1)
+	{
+		SourceFilePath = sourceFilePath;
+		SourceLineNumber = sourceLineNumber;
+	}
+
 	/// <inheritdoc/>
 	public string? DisplayName { get; set; }
 
@@ -36,4 +46,10 @@ public class FactAttribute : Attribute, IFactAttribute
 
 	/// <inheritdoc/>
 	public int Timeout { get; set; }
+
+	/// <inheritdoc/>
+	public string? SourceFilePath { get; }
+
+	/// <inheritdoc/>
+	public int SourceLineNumber { get; }
 }
--- a/src/xunit.v3.core/Framework/TheoryDiscoverer.cs
+++ b/src/xunit.v3.core/Framework/TheoryDiscoverer.cs
@@ -94,6 +94,8 @@ public class TheoryDiscoverer : IXunitTestCaseDiscoverer
 					details.SkipUnless,
 					details.SkipWhen,
 					testMethod.Traits.ToReadWrite(StringComparer.OrdinalIgnoreCase),
+					sourceFilePath: theoryAttribute.SourceFilePath,
+					sourceLineNumber: theoryAttribute.SourceLineNumber,
 					timeout: details.Timeout
 				)
 				: (IXunitTestCase)new XunitDelayEnumeratedTheoryTestCase(
@@ -108,6 +110,8 @@ public class TheoryDiscoverer : IXunitTestCaseDiscoverer
 					details.SkipUnless,
 					details.SkipWhen,
 					testMethod.Traits.ToReadWrite(StringComparer.OrdinalIgnoreCase),
+					sourceFilePath: theoryAttribute.SourceFilePath,
+					sourceLineNumber: theoryAttribute.SourceLineNumber,
 					timeout: details.Timeout
 				);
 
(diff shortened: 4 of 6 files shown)


src/xunit.v3.core/Framework/FactDiscoverer.cs
```diff
@@ -84,11 +86,15 @@ static ExecutionErrorTestCase ErrorTestCase(
...
-		return new(
+		var errorTestCase = new ExecutionErrorTestCase(
 			details.ResolvedTestMethod,
 			details.TestCaseDisplayName,
 			details.UniqueID,
 			message
 		);
+
+		errorTestCase.SourceFilePath = factAttribute.SourceFilePath;
+		errorTestCase.SourceLineNumber = factAttribute.SourceLineNumber;
```
Youssef1313, line 97: It's probably better to have these information returned from `GetTestCaseDetails`, and update `ExecutionErrorTestCase` constructor to accept these.

I just went with something quick as a "Poc".

bradwilson:
My question is how accurate this strategy is vs. `DiaSession`.

- If it's as good as `DiaSession`, then wouldn't this entirely negate the need for `ISourceInformationProvider` (at least for v3 tests)?
- If it's not as good as `DiaSession`, then wouldn't it make more sense to defer to `ISourceInformationProvider` and only use this as a last resort?

Youssef1313:
Alternative to this, we can look into the option of providing a helper to get source information as a separate NuGet package that you can depend on to provide the information when requested (e.g, via environment variable - which we will set for you when using F# in VS, or when using DevKit).

The implementation we will provide as a NuGet package will most likely be a copy of https://github.com/microsoft/vstest/tree/7b36464c47dbe0bd968972b3470e6e9da38209f2/src/Microsoft.TestPlatform.ObjectModel/Navigation. Or alternatively you can copy that implementation directly in xunit if you want to avoid an extra dependency.

bradwilson:
I'll look at what's involved in internally replicating the DiaSession code, since that's likely to be the most accurate (correct line, supports any language that can produce PDBs, etc.).

bradwilson:
After looking over the code, I don't think this is a good idea, because it depends on being able to find msdia140.dll, which I don't think we should ship (and maybe don't even have the right to ship).

Youssef1313:
About shipping that dll, I'm not sure. There is a mention [here](https://learn.microsoft.com/visualstudio/releases/2022/redistribution#dia-sdk) about redistribution but I'm not sure whether or not you can ship it.

Probably you can load it from VS installation itself by accessing `VSINSTALLDIR` environment variable, and from there it will be in `DIA SDK\bin`. But then we need to also find it when running DevKit and I'm not sure if it's shipped there somewhere or not.

Or, maybe you can start by supporting portable PDB only for now.

bradwilson:
I already have a dependency on Mono.Cecil, so I'm testing that now to see whether it has everything I need.

bradwilson:
It seems like Cecil will work, but only for tests in the root test assembly. That is, if the test is defined in a secondary assembly, the type/method lookup will fail. We will not chase down every assembly in the build folder looking for the type/method.

This could be combined with the caller info attributes used as a fallback when the source information provider cannot find the associated type/method. It will be less accurate, but it will still allow tests defined elsewhere (at least for C#/VB) to provide "close enough" source information in this scenario.

bradwilson:
I've been trying work out a way to add the caller info without a breaking change, but it doesn't seem possible due to the need for the changed constructors on `FactAttribute` and `TheoryAttribute`.

The Cecil changes are not breaking, so I'm going to push them up into the main tree.

Youssef1313:
> but it doesn't seem possible due to the need for the changed constructors on FactAttribute and TheoryAttribute.

Yup. There are two ways to workaround it:

1. OverloadResolutionPriorityAttribute, but that means it won't work correctly for old language versions before ORPA support was added.
2. UGLY: Ship two versions of the dll, one with the caller info constructor only, and another that has the two constructors. Hack things around with MSBuild so that the dll with caller info constructor only is the one that's passed to the compiler, but the other dll is what ends up in output directory. That way, all the constructors are present at runtime, by at compile-time the compiler has no choice other than selecting the constructor with caller info. But it's too ugly :)

bradwilson:
The Cecil-based source information is available in v3 `2.0.2-pre.16`. It should resolve your issue for any tests that live in the test assembly, but it would definitely be helpful for some verification on your side @Youssef1313 since I don't have the actual problem to repro. 😄 https://xunit.net/docs/using-ci-builds

bradwilson:
I did a performance measurement of using Cecil on the test assembly + every assembly that is a DLL in the same folder as the test assembly, and the startup cost is low enough that I think we should be able to cover the 99% case for tests without needing to go to the extent of using the caller info on the `FactAttribute`.

bradwilson:
@Youssef1313 Okay, the changes have been integrated into `2.0.2-pre.21` which include source information for MTP.

bradwilson:
I cannot demonstrate this in MTP mode (since it doesn't offer the ability to show this information, that I could tell), but you can see now that our discovery and test case starting messages include source information:

image

image

bradwilson:
And this includes F#/VB of course, since it's PDB based:

image

Youssef1313:
@bradwilson We would also expect a performance regression compared to VSTest because this bypasses the faster Roslyn-based discovery we have on Test Explorer side. So we will probably need a way to disable this only in VS for C# and VB. Is it something you want to do? (either completely on your side, or by us providing you some flag to control whether or not the Cecil-based info provider is enabled?)

bradwilson:
There is no Roslyn based discovery for xUnit.net v3 unless you're talking about something that hasn't shipped yet. (Or you mean something different by "discovery" than test discovery...)

Youssef1313:
> There is no Roslyn based discovery for xUnit.net v3

This is completely on Test Explorer side. We use `location.type` and `location.method` (which are calculated from TestMethodIdentifierProperty), and hand them to Roslyn (https://github.com/dotnet/roslyn/blob/748b0e60d1ea4f5d7f294024338c1649f58330ad/src/Features/Core/Portable/ExternalAccess/UnitTesting/API/UnitTestingSearchHelpers.cs).

Historically, this implementation was done because reading PDBs was very expensive, especially for repos of the size of dotnet/runtime and dotnet/roslyn (hundreds of thousands of tests).

Youssef1313:
> Also where do you have an unsigned copy of xunit.v3.runner.common?

Hmm, if not from `https://f.feedz.io/xunit/xunit/nuget/index.json` then I don't know. Let me clear my NuGet cache and re-restore.

bradwilson:
Everything in feedz.io gets signed.

What's weird is that you have both an unsigned and a signed copy. I just don't know where the unsigned copy came from. I've definitely tested this and it generally works on .NET Framework.

I have considered excluding the xUnit.net assemblies, since they'll never contain your tests, but right now we're signing everything, including the tests, which would make it impossible for me to actually test this feature if we exclude everything with our public key token. 😬

bradwilson:
> This is completely on Test Explorer side.

Just so we're clear, we've only added this feature for you.

How am I supposed to know when we're running in a mode where Test Explorer can provide this information vs. a mode where it can't? You're literally the only person who has this problem, so I hope you have a way for me to turn it on/off automatically.

Youssef1313:
> How am I supposed to know when we're running in a mode where Test Explorer can provide this information

On a high level:

- Visual Studio C#/VB: Test Explorer can already provide source information
- Visual Studio F#: Test Explorer can **not** provide source information.
- VS Code DevKit: Test Explorer can **not** provide source information for **all languages**.

Regarding the **how** you are supposed to know it, I'm just listing possibilities so we can agree on what to do:

- Today, we pass this information in `TESTINGPLATFORM_EXPERIMENTAL_VSTEST_RUNSETTINGS` (in `CollectSourceInformation`). This environment variable today is only present for VSTestBridge, but IDE sets it unconditionally so in theory you can still read it.
- We could have a simpler environment variable like `TestExplorerNeedsSourceInformation` which is either `1` or `0`.
- We could plug something more complex in the protocol itself and find a way to propagate it to the test framework somehow. (no concrete design in mind for this one yet)
- You could try to detect VS vs DevKit (there will probably be some env variable that can detect one or the other), and you can detect the language (e.g, you can use `$(Language)` in MSBuild to include it in `RuntimeHostConfigurationOption` which can be read at run-time with `AppContext.TryGetSwitch`).

Youssef1313:
@bradwilson For signing, here is what I see in NuGet Package Explorer

image

The dll in xunit.v3.runner.common NuGet package version 2.0.2-pre.2.4 has PublicKeyToken = null.

bradwilson:
I'll double check the build process.

bradwilson:
> * Today, we pass this information in `TESTINGPLATFORM_EXPERIMENTAL_VSTEST_RUNSETTINGS` (in `CollectSourceInformation`). This environment variable today is only present for VSTestBridge, but IDE sets it unconditionally so in theory you can still read it.

It sounds like this is only option that would work today. Everything you listed was "we could do this...". Is that accurate?

> * You could try to detect VS vs DevKit (there will probably be some env variable that can detect one or the other), and you can detect the language (e.g, you can use `$(Language)` in MSBuild to include it in `RuntimeHostConfigurationOption` which can be read at run-time with `AppContext.TryGetSwitch`).

This feels like waaaaaaaay too much internal information for me to code into the system. The "rules" that you've set down are only guaranteed to be accurate today.

bradwilson:
Given the name of `TESTINGPLATFORM_EXPERIMENTAL_VSTEST_RUNSETTINGS`, how long can people safely assume this passing mechanism will exist?

bradwilson:
How are you testing F#/VB in VS Code, given that the C# Dev Kit does not support the other two languages?

Youssef1313:
> How are you testing F#/VB in VS Code, given that the C# Dev Kit does not support the other two languages?

Yeah sorry my previous comments were confusing by mentioning "all languages". F#/VB aren't applicable to C# DevKit.

> Given the name of `TESTINGPLATFORM_EXPERIMENTAL_VSTEST_RUNSETTINGS`, how long can people safely assume this passing mechanism will exist?

I think it's unlikely to break it in Visual Studio or DevKit because it will then be a breaking change. @drognanar can you confirm please?

bradwilson:
I have verified that everything works as expected (F# now works in Visual Studio, C# works in Dev Kit), [respecting the environment variable](https://github.com/xunit/xunit/commit/8a1c6c2a8b04e2b883322365d4428f668810d8fe).

I've also addressed [the issue with the lack of the strong naming](https://github.com/xunit/xunit/commit/5812b89453010252ee9211bca57cdb48cb27c429) on the ILRepack'd version of `xunit.v3.runner.common`: 

image

(I have an [open issue](https://github.com/gluck/il-repack/issues/403) as to why "Compiler Flags" is not being preserved by ILRepack, but I assume that doesn't affect you.)

This is available for verification on your side in build `2.0.2-pre.28` @Youssef1313

PR #3277, opened 2025-04-28
Avoid duplicating file extensions in attachment file names

0xced:
This prevents file names with double extensions (e.g. screenshot.png.png) if the attachment name already contains a file extension.


bradwilson:
Thanks!

PR #3285, opened 2025-05-02
Default CollectSourceInformation to false

Youssef1313:
In DevKit, it will be forced as true.

```xml
<RunSettings>
  <RunConfiguration>
    <CollectSourceInformation>true</CollectSourceInformation>
  </RunConfiguration>
  <DataCollectionRunSettings>
    <DataCollectors />
  </DataCollectionRunSettings>
</RunSettings>
```

In VS F#, it will be forced as true as well:

```xml
<RunSettings>
  <BoostTestInternalSettings xmlns:xsd="http://www.w3.org/2001/XMLSchema" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance">
    <VSProcessId>32540</VSProcessId>
  </BoostTestInternalSettings>
  <RunConfiguration>
    <ResultsDirectory>C:\Users\ygerges\source\repos\TestProject83\TestResults</ResultsDirectory>
    <SolutionDirectory>C:\Users\ygerges\source\repos\TestProject83\</SolutionDirectory>
    <CollectSourceInformation>True</CollectSourceInformation>
  </RunConfiguration>
  <GoogleTestAdapterSettings xmlns:xsd="http://www.w3.org/2001/XMLSchema" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance">
    <SolutionSettings>
      <Settings />
    </SolutionSettings>
    <ProjectSettings />
  </GoogleTestAdapterSettings>
</RunSettings>
```

The above scenarios are the only scenarios where it's important to produce source information. Defaulting to false is good for at least command-line where it's an extra cost without any benefits to produce source info.

### DIFF
--- a/src/xunit.v3.runner.common/Utility/RunSettingsUtility.cs
+++ b/src/xunit.v3.runner.common/Utility/RunSettingsUtility.cs
@@ -29,7 +29,7 @@ internal static class RunSettingsUtility
 				}
 				catch { }
 
-				collectSourceInformation ??= true;
+				collectSourceInformation ??= false;
 			}
 
 			return collectSourceInformation.Value;


bradwilson:
You use the words "it will" twice. When?

bradwilson:
Also, how about `dotnet test` in MTP mode?

Youssef1313:
> You use the words "it will" twice. When?

It's done like this already.

bradwilson:
This is not what I observed. DevKit does not set the environment variable.

bradwilson:
vs. Visual Studio

image

Youssef1313:
That is not expected for me. I will be off for two weeks. @drognanar @nohwnd Please double check here.

Youssef1313:
Ping @drognanar

drognanar:
@Youssef1313 the MTP client is responsible for these variables. perhaps the tests in DevKit were not being run via MTP?

bradwilson:
@Youssef1313 What version of C# Dev Kit are you using?

bradwilson:
@Youssef1313 I still see "null" with C# Dev Kit version 1.19.63

image

bradwilson:
When it gets to the RTM version then I'll be able to merge this.

bradwilson:
@Youssef1313 I force updated to 1.20.14 (for some reason VS Code did not think that was newer), and yet I still see `null`.

I'm open to suggestions on how to resolve the differences between your environment and mine, but this reproduces as `null` on two difference machines for me.

bradwilson:
Is there perhaps an unstated dependency on a .NET 10 SDK pre-release to make this work?

nohwnd:
same version of extension, added sleep to my test to double check I am not running under testhost (via proces explorer), and I was, and I was missing this:

image

Youssef1313:
@bradwilson If that option isn't selected, the tests will be run with VSTest so it should work.

bradwilson:
This is now available in v3 `2.0.3-pre.26` https://xunit.net/docs/using-ci-builds

bradwilson:
(I'm still interested in an answer to "How does C# Dev Kit determine that a test project is MTP vs. VSTest?")

PR #3323, opened 2025-06-15
Dispose CTS in TestContext

Youssef1313:
@bradwilson You better know how things flow in xunit. I'm wondering if this could be causing some memory leaks. For winforms, I still see memory leaks:

image

I wonder if this CTS which is not being disposed is the issue here?

### DIFF
--- a/src/xunit.v3.core/TestContext.cs
+++ b/src/xunit.v3.core/TestContext.cs
@@ -27,6 +27,7 @@ public sealed class TestContext : ITestContext, IDisposable
 	IMessageSink? internalDiagnosticMessageSink;
 	readonly ConcurrentDictionary<string, object?>? keyValueStorage;
 	readonly CancellationTokenSource testCancellationTokenSource = new();
+ 	readonly CancellationTokenSource linkedCancellationTokenSource;
 	readonly List<string>? warnings;
 
 	TestContext(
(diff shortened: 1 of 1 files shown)


bradwilson:
Oops 😬

bradwilson:
Thanks!

bradwilson:
Available in v3 `3.0.0-pre.25` https://xunit.net/docs/using-ci-builds

Youssef1313:
@bradwilson So far I'm not able to test with 3.0.0-pre.25 in the original repro because it uses xunit.stafact which no longer work due to binary breaking changes related to caller info.

Can we have this fix backported?

**Side note:** It took me quite a while to figure out why things were not working. It was the attribute from stafact throwing MissingMethodException (because of the signature change in 3.0.0-pre.xx), and looks like the exception is swallowed somewhere (maybe https://github.com/xunit/xunit/blob/6c03ceb0ab58e27bd9b9a3186770c2970c1f704d/src/xunit.v3.common/Extensions/ReflectionExtensions.cs#L81)

bradwilson:
~Unfortunately this fix itself is a breaking change so it can't be backported into 2.x.~

Sorry, I was thinking about the other issue. This can be backported.

bradwilson:
Available in v3 `2.0.4-pre.2` https://xunit.net/docs/using-ci-builds

bradwilson:
> It took me quite a while to figure out why things were not working. It was the attribute from stafact throwing MissingMethodException (because of the signature change in 3.0.0-pre.xx), and looks like the exception is swallowed somewhere

I'd been planning to address this before 3.0.0 shipped, and it's been fed through with [this commit](https://github.com/xunit/xunit/commit/cff965e4881eb12254d98633521aa747873ec8c4?w=1). In most cases these failures should now show up as diagnostic messages.

Youssef1313:
@bradwilson I'm revisiting this again, it looks like `Dispose` is never actually called?

bradwilson:
@Youssef1313 Let me verify that we're disposing the `TestContext`, as that would be the most obvious reason why the CTS wouldn't be getting disposed.

bradwilson:
We're not, and adding disposal guards is bringing up issues, so this isn't quite the quick fix I was hoping for. Will let you know when I have another build for you to test.

bradwilson:
@Youssef1313 Build `2.0.4-pre.3` now includes the proper disposal of the test context.

Any attempt to access the `CancellationToken` property or `CancelCurrentTest` method of a disposed test context will throw (since the cancellation token sources have been disposed); the rest of the metadata retrieval is still permitted.

PR #3442, opened 2025-11-18
fixes #3441, typo in trx transform

### LINKED ISSUE #3441: The TRX transform generates invalid .trx files
in v3, when tests run, I generate a .trx file with the output. The file is generated but I cannot open it with Visual Studio and more web readers that I tried.

I realised that the issue is that the node `ReusltSummary` is:

`<ResultSummary outcome="Complete">`

and the expected value by the xml schema is:

`<ResultSummary outcome="Completed">`

I manually changed that and now it works!

beto-rodriguez:
- Fixes #3441


bradwilson:
Thanks!

PR #3461, opened 2025-12-19
Improve error message for command-line validation

Youssef1313:
When constructing command-line in a scripts (e.g, a bit involved PowerShell script), it's easier to see the exact value to ease debugging/spotting where the script is constructing the command incorrectly.


bradwilson:
Thanks! Going to merge despite the CI build failure. I'm planning to remove official Mono support because I've started having these issues with Mono + macOS, so I'm not concerned with the failure.

Issue #3475, opened 2026-01-19
Create Assert.All overload with `bool strict` which fails for empty collections when true


bradwilson:
As discussed in https://github.com/xunit/xunit/discussions/3473

ssaporito:
Hi! Just wanted to point out that this issue remained open even though the PR has already been merged. @bradwilson

PR #3476, opened 2026-01-19
Added an overload for All and AllAsync asserts to allow the caller to specify if test should fail if the collection is empty.

tiger4589:
Following the [idea](https://github.com/xunit/xunit/discussions/3473) I suggested and the [issue](https://github.com/xunit/xunit/issues/3475) created for that, I am trying to contribute to this repo for the first time.

Changes that were made are simple:

- Added an overload method for Assert.All with an extra `bool` parameter
- Added an overload method for Assert.AllAsync with an extra `bool` parameter

In both cases, the parameter is called `throwIfEmpty` which will throw an `AllException` if the parameter is set to true and the collection is empty.

This would kind of allow users to run 3 asserts into 1: `NotNull`, `NotEmpty`, `All` for any collection they're testing against.

I am open for suggestions if there's anything I've done wrong or different from code guidelines.

### DIFF
--- a/src/xunit.v3.assert.tests/Asserts/CollectionAssertsTests.cs
+++ b/src/xunit.v3.assert.tests/Asserts/CollectionAssertsTests.cs
@@ -33,6 +33,13 @@ public class CollectionAssertsTests
 		}
 
 		[Fact]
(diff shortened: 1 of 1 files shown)


bradwilson:
Thanks!

Issue #3526, opened 2026-03-29
Add ability to override environmental values on a per-test basis


bradwilson:
An example:

If assertion failure reporting makes strings impossible to compare because of length (differences are after the default length), today the only choice is to either flip a global switch (like env var `XUNIT_PRINT_MAX_STRING_LENGTH`), which may inappropriately impact other tests, or add additional debugging information into the test (like printing out values when detecting failure).

An alternative that influences just the test, that can be called during the test, would be ideal. Something like `Assert.SetPrintMaxStringLength()`, that sets an async local value so it only affects the current test.

Tenno641:
I gave this a shot.

The main thing I’m unsure about is the approach: I used an `AsyncLocal<int?>` inside `ArgumentFormatter` to keep the override scoped per test while avoiding additional dependencies or references.

I also added tests around failure messages, but they turned out to be quite fragile due to dynamic padding and the `↓ (pos XX)` marker. I did my best to align them with the existing test style.

bradwilson:
I think `AsyncLocal` is a reasonable strategy. I'm expanding to the rest of the overrides. I can't push to your git submodule, so I'll merge here locally, while retaining your commit credit.

bradwilson:
Available in v3 `4.0.0-pre.105` https://xunit.net/docs/using-ci-builds

bradwilson:
The new APIs are:

* `Assert.OverrideMaxEnumerableLength`
* `Assert.OverrideMaxObjectDepth`
* `Assert.OverrideMaxObjectMemberCount`
* `Assert.OverrideMaxStringLength`

PR #3560, opened 2026-04-20
Feature/override environmental values per test

Tenno641:
This PR implements #3526 - the ability to override environmental values on a per-test basis.

Specifically, it adds:

```csharp
Assert.SetPrintMaxStringLength(int? maxLength);
```

`0` disables truncation completely for the current test.

`null` (or not calling the method) removes the override.


bradwilson:
Thanks!

PR #3561, opened 2026-04-20
Sanitize attachment filenames against cross-platform invalid chars

### LINKED ISSUE #1721: XUnit V3: Adds attachment with colon in filename - GitHub artifacts upload fails
> [!Note]
> This issue is not about a defect in Verify per se, but an integration hurdle with GitHub.

When a test fails the received file is attached (`VerifierSettings.AddTestAttachment`) via [TestContext.Current.AddAttachment](https://api.xunit.net/v3/2.0.0/v3.2.0.0-Xunit.TestContext.AddAttachment.html#v3_2_0_0_Xunit_TestContext_AddAttachment_System_String_System_Byte___System_String_).
Verify.XunitV3 uses `$"Verify snapshot mismatch: {path}"` as file name. This contains a colon.
Because on Windows file systems, this is an illegal character for file names, GH upload artifacts doesn't accept it, and the following error is reported:

> The path for one of the files in artifact is not valid: /TestResults/_runnervmeorf1_2026-04-20_13_15_11/In/e19a31c6-3952-4ff5-a4b9-d4afc7b3754a/runnervmeorf1/Verify snapshot mismatch: _home_runner_work_VerifyXunitV3AttachmentWithColonFailsUploadToGitHub_VerifyXunitV3AttachmentWithColonFailsUploadToGitHub_FailingTests.TheFail.received.bin.bin. Contains the following character:  Colon :
>          
>Invalid characters include:  Double quote ", Colon :, Less than <, Greater than >, Vertical bar |, Asterisk *, Question mark ?, Carriage return \r, Line feed \n
>          
>The following characters are not allowed in files that are uploaded due to limitations with certain file systems such as NTFS. To maintain file system agnostic behavior, these characters are intentionally not allowed to prevent potential problems with downloads on different file systems.

## Workaround
Before GH upload artifacts, run a script to rename files that contain ':' and replace it with a different character.

## Minimal Repro
https://github.com/BrunoJuchli/VerifyXunitV3AttachmentWithColonFailsUploadToGitHub

Action run result: https://github.com/BrunoJuchli/VerifyXunitV3AttachmentWithColonFailsUploadToGitHub/actions/runs/24668584595

## Proposed Change

- remove the colon from the file name, or
- replace the colon with a different separator for better compatibility, or
- do not prefix the filename with `Verify snapshot mismatch:` (simply use the original filename instead)

I think you @SimonCropp have a better grasp on what alternative is best to choose here.

For replacing the colon with a different separator, how about:
- `--`
- `@`
- ` at `
- ` with `

----

> [!Note]
> I've looked at the other test-runner integrations, and the issue does not seem to apply to any of the them.

SimonCropp:
MediaTypeUtility.GetSanitizedFileNameWithExtension used
  Path.GetInvalidFileNameChars(), which is OS-dependent. On Linux it only
  returns '/' and '\0', so chars that are invalid on Windows (':', '"',
  '<', '>', '|', '*', '?', '\') passed through unchanged. Attachment
  files are frequently consumed on a different OS than they were written
  on — e.g. produced on a Linux CI runner and then uploaded as GitHub
  Actions artifacts, which rejects those same Windows-invalid chars and
  fails the upload.

  Replace the OS-dependent set with the union of invalid filename chars
  across Windows/Linux/macOS plus control chars 0-31, so sanitized
  filenames are portable regardless of the producing OS.

fixes https://github.com/VerifyTests/Verify/issues/1721

### DIFF
--- a/src/xunit.v3.runner.common.tests/Utility/MediaTypeUtilityTests.cs
+++ b/src/xunit.v3.runner.common.tests/Utility/MediaTypeUtilityTests.cs
@@ -1,0 +1,47 @@
+using Xunit;
+
+public class MediaTypeUtilityTests
+{
+	[Fact]
+	public void ReplacesCrossPlatformInvalidChars()
+	{
+		// Chars invalid on Windows must be scrubbed regardless of the runtime OS, since
+		// attachments written on one OS are frequently consumed on another (e.g. uploaded
+		// as GitHub Actions artifacts, which rejects ':', '"', '<', '>', '|', '*', '?').
+		var result = MediaTypeUtility.GetSanitizedFileNameWithExtension("name:with<>|\"*?/\\bad", "text/plain");
+
+		Assert.Equal("name_with________bad.txt", result);
+	}
+
+	[Fact]
+	public void ReplacesControlChars()
+	{
+		var result = MediaTypeUtility.GetSanitizedFileNameWithExtension("a\0b\tc\nd", "text/plain");
+
+		Assert.Equal("a_b_c_d.txt", result);
+	}
+
+	[Fact]
+	public void AppendsExtensionWhenMissing()
+	{
+		var result = MediaTypeUtility.GetSanitizedFileNameWithExtension("attachment", "application/json");
+
+		Assert.Equal("attachment.json", result);
+	}
+
+	[Fact]
+	public void DoesNotAppendExtensionWhenAlreadyPresent()
+	{
+		var result = MediaTypeUtility.GetSanitizedFileNameWithExtension("attachment.json", "application/json");
+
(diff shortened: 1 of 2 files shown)


bradwilson:
Can you retarget this to `rel/4.0.0`? I'm not planning on shipping another 3.x build. Many thanks. 😄

SimonCropp:
done

bradwilson:
Of course. Mono. 😠

bradwilson:
Thanks!

bradwilson:
Available in v3 `4.0.0-pre.97` https://xunit.net/docs/using-ci-builds

SimonCropp:
@bradwilson thanks

Issue #2055, opened 2019-12-03
Feature Request: Run Tests in Parallel with Shared Context


zijianhuang:
**Background**
Currently by default tests in the same assembly but in different classes run in parallel.

To run tests in different classes not in parallel, you put them  into the same test collection.

To have multiple classes sharing the same contexts, you use a Collection Fixture, however, then the tests run one by one.

I have some integration tests that need to launch IIS or DotNet Host, or drop/create databases, via collection fixtures.

**Feature Request**

Allow to run tests in different classes in parallel sharing the same context like dropping/creating databases. 

Programming interface expected:

```csharp
public sealed class CollectionAttribute : Attribute
	{
		public CollectionAttribute(string name, boolean parallel=false);
	} 


[Collection("Database collection", true)]
public class DatabaseTestClass1
{
 //...
}

[Collection("Database collection", true)]
public class DatabaseTestClass2
{
    // ...
}
```

redoz:
I'll chime in and say it would be very nice if v3 offered more control over how tests in collections were executed (e.g. in parallel) or offered up a different means of having shared contexts and controlling execution. I understand this isn't really a concern when unit-testing, but we're using xbehave as our integration testing framework (mostly because it lets us re-use existing infrastructure that supports xunit, runners, reporting, etc.) and I've ended up hacking on my own "AssemblyFixture" support so we can handle shared context across our test-suite. But even so I'm forced to run some tests in a non-parallel collection simply to make sure it doesn't interfere with the rest of the test-suite which is a bit inefficient as those tests run serialized now.

fjmorel:
I would love something like this. We currently use a CollectionFixture to start a Mongo server for the majority of our tests, but are using randomized collection names for each test to avoid having to clean up between each test. Since they can't conflict with each other, being able to run all the tests within this collection in parallel would help a lot.

markm77:
I made a comment on the other parallelisation issue: https://github.com/xunit/xunit/issues/1986#issuecomment-669429834.

Yes, it would be really great to have the ability to run tests in parallel when using a collection fixture. We use Xunit for both unit and integration tests, and for the integration tests often instance a webhost via a collection fixture. In addition, we also often use a single `[Theory]` test with multiple `[MemberData]` cases (i.e. dynamic test case generation). I'm really hoping the implementation of this allows `[MemberData]` cases for a single `[Theory]` test to run in parallel!

pdevito3:
[...]

Issue #1986, opened 2019-07-20
Feature Request - Test case parallelization when there is no class/collection fixture


jakubsuchybio:
* Did you search the documentation @ https://xunit.github.io/ ?
Yes.
* Did you search the issues list to see if someone already reported it?
Yes #1227 is kinda mentioning this, but in different context.

I propose to add new functionality for test collection to run it's test cases in parallel when and only when the test collection doesn't have any class fixture or it is not participating in collection fixture.

Why? Well xunit by default instantiate the test class for every test case. So there is no shared state when there is no class/collection fixture. Therefore these test cases can be parallelized without breaking because of some shared state.

**There are 3 possible implementations.**
A. To have this as default behaviour. Then it would be beneficial to all, but it could be too complex to grasp for beginners. That when no class/collection fixture, test cases in the same test collection runs parallel, but when class/collection fixture is present, then they run sequential.
B. Opt-out option. So basically A but when someone needs to disable it, they can via options.
C. Opt-in option. Here the default behaviour from 2x won't change, but for more "experienced" users it could bring great benefit/speedup for tests in test collections that doesn't use class/collection fixture.

Personally A without option to disable it is not feasible at all. B would be most beneficial for everyone, but for the cost of more complexity in default parallelism behaviour. So I think C is the best implementation that should definitely be in xunit 3.0.

By my humble opinion the majority of tests are in this category (not using class/collection fixture), therefore this could be really significant performance boost.

Thanks!

andre-ss6:
I recently started with xUnit and I'm not even aware of what collections are. I have a project with many dozens tests which share absolutely no state. My expectation was that they would all be run in parallel and I was disappointed to find out they don't.

I see that in the other issue some pointed out that there would be some roadblocks with the suggested route. What about this one though? Would it be difficult to implement it this way?

bradwilson:
We are definitely not changing the default behavior. That would unnecessarily break everybody who is already depending on it.

I'm open to adding new non-default parallelization options in v3.

psanchez54:
An optional configuration, per project for default behaviour and per method or class to make them "parallelizable" would be awesome, I had 25 test methods for a login page and had to split it into 25 classes just for the tests to run faster in parallel...

alexb5dh:
How about implementing something like `CollectionPerMethod` for [CollectionBehavior](https://github.com/xunit/xunit/blob/master/src/xunit.core/CollectionBehavior.cs) and keeping default as is?

[...]

PR #3581, opened 2026-05-23
#1986 / #2055: Add Support For Inner Collection Test Parallelization

matthew-mabrey:
# Changes
- Modifies `ParallelismOptions` flag enum to support parallelizing tests within a collection. `ParallelismOptions` are now passed down through the core test runners and parallelization can be enabled at every level (Assembly, Collection, Class, Method, Test Case, and Test).
  
  `ParallelismOptions` are determined by:
    - `ITestFrameworkExecutionOptions`: Defines parallelism options for the process. Only the values `None`, `Assemblies`, and `Collections` have any effect, as inner collection test parallelization is determined by the collection attributes. `Assemblies` indicates an assembly can be parallelized with other assemblies and a lack of the `Collections` flag disables test collection parallelization for the assembly.
    - `CollectionDefinitionAttribute`: Defines parallelism options for a single collection.
    - `CollectionBehaviorAttribute`: Defines default options for collections in the assembly decorated with this attribute.

- Overrides methods in core test runner implementations to support parallelization at the test class, method, test case, and test levels
- The semaphore used to limit the maximum number of running tests (when `ParallelAlgorithm.Conservative` is used) is now passed down the test pipeline stages when inner collection test parallelization is enabled.
- Changes long-running test detection to be determined based off of `ITestStarting` / `ITestFinished` messages instead of `ITestCaseStarting` / `ITestCaseFinished` messages since test cases can be delayed waiting for the parallelization semaphore for an indefinite amount of time

# Fixes:
- Fixes the potential for a deadlock in `MaxConcurrencySyncContext` if the dispose ever runs on one of the worker threads
- Fixes a case where `AcceptanceTestV3.RunAsync` could fail to determine the test assembly when running tests via the Rider IDE

# Examples:
```
# Run every test in this collection serially and without being parallelized with other collections
# Equivalent to "DisableParallelization = true"
[CollectionDefinition(ParallelismOptions = ParallelismOptions.None)]

# Parallelize classes in the collection and run in parallel with other collections in the assembly
[CollectionDefinition(ParallelismOptions = ParallelismOptions.Collections | ParallelismOptions.Classes)]

# Run every test in this collection in parallel and run the collection in parallel with other collections in the assembly
[CollectionDefinition(ParallelismOptions = ParallelismOptions.All)]

# Run every test in this collection in parallel without this collection being parallelized with other collections
[CollectionDefinition(ParallelismOptions = ~ParallelismOptions.Collections | ParallelismOptions.All)]

```


matthew-mabrey:
@dotnet-policy-service agree

matthew-mabrey:
# Comments/Questions:

@bradwilson 
- I realize this is a departure from the changes I originally proposed [here](https://github.com/xunit/xunit/issues/2055#issuecomment-3921759251). I initially implemented it as planned, but realized it would be more flexible by implementing it in this manner. If there are fundamental issues with this design, I can always revert to the previous proposal. I figured this implementation allows more flexibility and extensibility since flags can be added in the future without needing to create and pass around `EnableTest<Level>Parallelization` booleans for each level of the test execution pipeline.

- For the collection definition/behavior attribute APIs, would you prefer we keep `DisableParallelization` and `ParallelismOptions`? I've initially kept `DisableParallelization` for backwards compatibility so developers don't have to update existing attribute definitions, but not sure if we'd prefer just removing `DisableParallelization` to simplify things / reduce potential confusion going forward

- It's a bit unclear to me how backwards compatibility is maintained for xUnit v1/v2. How should I handle it for these changes? For example, I'm replacing `ParallelizeTestCollections` with `ParallelismOptions` in `src/xunit.v3.runner.utility/SimpleRunner/AssemblyRunnerOptions.cs`. Should I keep `ParallelizeTestCollections` for v2 and have it use `ParallelismOptions` internally, or just stick with the current changes?

- I'm slightly worried about the potential for a huge number of tasks to be started in the `ParallelismOptions.All` case, but I haven't figured out a great solve for it yet and I'm not familiar enough to know how much of a performance/memory hit it'll be. Might need to do some profiling and/or think on it more.

bradwilson:
Thanks, I'll take a look at the code soon.

> * [...] I figured this implementation allows more flexibility and extensibility since flags can be added in the future without needing to create and pass around `EnableTest<Level>Parallelization` booleans for each level of the test execution pipeline.

Sounds reasonable.

> * For the collection definition/behavior attribute APIs, would you prefer we keep `DisableParallelization` and `ParallelismOptions`? I've initially kept `DisableParallelization` for backwards compatibility so developers don't have to update existing attribute definitions, but not sure if we'd prefer just removing `DisableParallelization` to simplify things / reduce potential confusion going forward

We would probably would want to mark the property as obsolete and tell the developer what the new API is, and then remove it in 5.0.

> * It's a bit unclear to me how backwards compatibility is maintained for xUnit v1/v2. How should I handle it for these changes? For example, I'm replacing `ParallelizeTestCollections` with `ParallelismOptions` in `src/xunit.v3.runner.utility/SimpleRunner/AssemblyRunnerOptions.cs`. Should I keep `ParallelizeTestCollections` for v2 and have it use `ParallelismOptions` internally, or just stick with the current changes?

I think probably the same thing as above: mark `ParallelizeTestCollections` as obsolete. We would also want to ensure XML docs for `ParallelismOptions` would make it clear which options are supported where:

* Test parallelism wasn't an option at all in v1
* Test parallelism only supports test collections in v2 and v3 prior to 4.0
* The other options are supported only on v3 4.0+

> * I'm slightly worried about the potential for a huge number of tasks to be started in the `ParallelismOptions.All` case, but I haven't figured out a great solve for it yet and I'm not familiar enough to know how much of a performance/memory hit it'll be. Might need to do some profiling and/or think on it more.

I don't really know the answer to this either, but it should only be a factor for users who choose the aggressive parallelism option. The conservative option (which is the default) should only ever start a number of tasks equal to the maximum thread count, regardless of the boundary of parallelism.

I suppose if someone creates 100k tests and asks for the aggressive parallelism option, then I guess they get to find out what that means. 😂 They could have already done this today in v2 (or v3 with the aggressive algorithm) by coding it as one test per class. Will be worth doing an experiment with some generated code to do just that...

matthew-mabrey:
> We would probably would want to mark the property as obsolete and tell the developer what the new API is, and then remove it in 5.0.

> I think probably the same thing as above: mark ParallelizeTestCollections as obsolete. We would also want to ensure XML docs for ParallelismOptions would make it clear which options are supported where:
> 
> Test parallelism wasn't an option at all in v1
> Test parallelism only supports test collections in v2 and v3 prior to 4.0
> The other options are supported only on v3 4.0+

Just pushed up these changes to address these https://github.com/xunit/xunit/pull/3581/changes/4e5cf9f0f2a275b248c6252a18cc7de5ae93a847

> I don't really know the answer to this either, but it should only be a factor for users who choose the aggressive parallelism option. The conservative option (which is the default) should only ever start a number of tasks equal to the maximum thread count, regardless of the boundary of parallelism.
> 
> I suppose if someone creates 100k tests and asks for the aggressive parallelism option, then I guess they get to find out what that means. 😂 They could have already done this today in v2 (or v3 with the aggressive algorithm) by coding it as one test per class. Will be worth doing an experiment with some generated code to do just that...

I was thinking about the case of someone using `ParallelAlgorithm.Conservative` + `ParallelismOptions.All` and just how many resources/tasks the test pipeline would use now that it is fanning out to every collection, class, method, and test case at the same time. In that case, each stage will be starting a task for every test class, method, etc and then eventually it'll get to the run test stage where they all await the same semaphore. 

I just pushed up some [more changes](https://github.com/xunit/xunit/pull/3581/changes/c39065c91f047dc4e37029f8a205391ed75425b9) with a new `TestPipelineSemaphore` that could help with this. Internally it creates a semaphore for each stage and tries to reduce the fan-out of the test pipeline parallelism by limiting each pipeline stage to `MaxParallelThreads` tasks. I'm not super committed to this implementation if you have other thoughts though.

matthew-mabrey:
> ParallelismOptions are determined by:
> 
> ITestFrameworkExecutionOptions: Defines parallelism options for the process. Only the values None, Assemblies, and Collections have any effect, as inner collection test parallelization is determined by the collection attributes. Assemblies indicates an assembly can be parallelized with other assemblies and a lack of the Collections flag disables test collection parallelization for the assembly.
> CollectionDefinitionAttribute: Defines parallelism options for a single collection.
> CollectionBehaviorAttribute: Defines default options for collections in the assembly decorated with this attribute.

I need to think on the `ITestFrameworkExecutionOptions` part of this more as well. Seems a little confusing currently and there isn't a way to override `CollectionDefinitionAttributes`/`CollectionBehaviorAttributes` via config/command line in the current implementation. Previously I had it implemented as always using the `ITestFrameworkExecutionOptions.ParallelismOptions?` above all when defined, but that leads to issues if someone just wants to set `ParallelismOptions.Assemblies` without overriding all the `CollectionDefinitionAttributes`/`CollectionBehaviorAttributes` in the assembly.

Maybe `ITestFrameworkExecutionOptions.ParallelismOptions` and `ITestFrameworkExecutionOptions.DisabledParallelismOptions` that are bitwise AND/OR'd with the collection's parallelism options?
```
(ITestFrameworkExecutionOptions.ParallelismOptions | collection.ParallelismOptions) & ~ITestFrameworkExecutionOptions.DisabledParallelismOptions
```

bradwilson:
> I just pushed up some [more changes](https://github.com/xunit/xunit/pull/3581/changes/c39065c91f047dc4e37029f8a205391ed75425b9) with a new `TestPipelineSemaphore` that could help with this. Internally it creates a semaphore for each stage and tries to reduce the fan-out of the test pipeline parallelism by limiting each pipeline stage to `MaxParallelThreads` tasks. I'm not super committed to this implementation if you have other thoughts though.

I think the most important part is to never exceed creating more than `n` tasks where `n` is the `MaxParallelThreads` value. What you're describing sounds like it violates that rule by exploding it as a cross product of `n1` collections, `n2` classes, `n3` methods, etc. (unless I'm reading what you wrote wrong).

matthew-mabrey:
> I think the most important part is to never exceed creating more than n tasks where n is the MaxParallelThreads value. What you're describing sounds like it violates that rule by exploding it as a cross product of n1 collections, n2 classes, n3 methods, etc. (unless I'm reading what you wrote wrong).

The new `TestPipelineSemaphore` still guarantees no more than `n` concurrent _`Test`_ tasks (Tasks which are executing user defined test code) are ever running at a time, but allows up to `ParallelizedTestPipelineStages * n` concurrent tasks running in total, where `ParallelizedTestPipelineStages = 5` in the current implementation. All but `n` tasks are just waiting on the semaphore and not doing substantial work though. Before I added `TestPipelineSemaphore`, the number of tasks did explode as a cross product like you mentioned. I expect `5 * n` is a negligible amount of tasks in most instances since by default `MaxParallelThreads` is set to the system processor count.

I haven't yet been able to think of a good way to both parallelize tests within collections _and_ never exceed exactly `n` tasks without major refactors to the test pipeline. Open to ideas though.

bradwilson:
Looking at this code, I have doubts about the effectiveness of the strategy.

First, I believe that you have two kinds of parallelism:

* Parallelism across assemblies
* Parallelism within an assembly

The feature request is really only addressing the latter. However, I don't think the latter situation is best defined as a "flags" kind of situation. I get that it's "flag-like" today because "Assembly" vs. "Collection" vs. "All" are collapsed into a single type in implementation, which is probably incorrect and should be fixed rather than propagated.

For intra-assembly parallelism, I think the user wants to pick the lowest place down the stack where parallelism is allowed. The primary feature request that people have is either "run everything in parallel" or "run everything without shared state in parallel". In reality, I think the best choice is to support "run everything in parallel" and make sure we offer individual test classes (and maybe test methods?) to opt out of parallelism, such that people who would've wanted "run everything without shared state in parallel" can at least turn on "run everything in parallel" and then manually opt-out when they know their shared state is not parallel-friendly. For sure, though the idea that you could turn on/off parallelism at each layer independently is definitely not a feature that we need, and it would be quite challenging to explain and conceptualize.

In the face of all that, I'm not 100% sure where to go from here with your existing work. It feels like it's headed in a direction that I'm not sure is the way I want to go, and I don't know whether it makes more sense to try to adapt it or just start over.

matthew-mabrey:
Fair enough. I agree the flags are probably over-complex for a user-facing API and don't work great with the assembly level parallelism as you mentioned.

I think I went down this path after my initial implementation because the code was starting to pass around a bunch of booleans to control parallelism and I thought it would be easier to pass it around as a single flag. Didn't necessarily expect users would often want to control parallelism independently at every layer (especially the method, test case, and test layers separately), but I thought it would be more flexible/extensible to implement it this way internally in case some future user has a unique parallelism setup they want (like a collection that runs its classes serially but the test methods within those classes in parallel).

I realize this feature wasn't explicitly asked for though so maybe not worth the extra complexity. That's my bad for not checking in with you on this change in design first.

> In the face of all that, I'm not 100% sure where to go from here with your existing work. It feels like it's headed in a direction that I'm not sure is the way I want to go, and I don't know whether it makes more sense to try to adapt it or just start over.

Do you see other fundamental issues with the changes besides the ParallelismOptions? My original changes were very similar, except it was just passing around a new `EnableTestParallelization` bool, and checking `ctxt.EnableTestParallelization` instead of `ctxt.ParallelismOptions.HasFlag()`.

> In reality, I think the best choice is to support "run everything in parallel" and make sure we offer individual test classes (and maybe test methods?) to opt out of parallelism, such that people who would've wanted "run everything without shared state in parallel" can at least turn on "run everything in parallel" and then manually opt-out when they know their shared state is not parallel-friendly. 

To better support this design, what do you think of these changes:

- At the Assembly level, add a `EnableTestParallelization` bool property to `ITestFrameworkExecutionOptions` to allow overriding for the entire assembly via config/commandline 
- On the CollectionDefinition/CollectionBehavior attributes, add a `EnableTestParallelization` bool property to enable full test parallelism within collections and keep the `DisableParallelization` property to allow collections to opt-out from running in parallel with other collections (maybe rename to `DisableCollectionParallelization` for clarity)
- Revert all `ParallelismOptions` changes in this PR and `EnableTestParallelization` will replace all the `ParallelismOptions` arguments that get passed down through the core test runner contexts in the current changes
- On Fact/Theory attributes, add a `DisableParallelization` bool property to allow individual test methods to opt-out of parallelization

I'm not sure yet how to support classes opting out of parallelization in this case since they don't have a specific attribute AFAIK. Maybe a new `[DisableParallelization]` attribute that can be placed on classes?

bradwilson:
> * On the CollectionDefinition/CollectionBehavior attributes, add a `EnableTestParallelization` bool property to enable full test parallelism within collections and keep the `DisableParallelization` property to allow collections to opt-out from running in parallel with other collections (maybe rename to `DisableCollectionParallelization` for clarity)

This is named as `Disable` instead of `Enable` because I wanted them to be `false` by default (like any `bool` value would be). Whether this is meaningful distinction or not is a decision that is many many years in the past. 😂 I don't think there's anything wrong with the way this is named, though, so changing it feels unnecessary.

> I'm not sure yet how to support classes opting out of parallelization in this case since they don't have a specific attribute AFAIK. Maybe a new `[DisableParallelization]` attribute that can be placed on classes?

After bouncing the ideas around in my head, I think I like this approach best:

* Create an assembly-level `ParallelizationAttribute` and add properties `DefaultMode` (new `ParallelMode` enum with values `None`, `Collections`, and `All`; defaults to `Collections`), and `Algorithm` (existing `ParallelAlgorithm` enum with existing default of `Conservative`). This acts as the assembly-level opt-out (`None`). This means also obsoleting `DisableTestParallelization` and `ParallelAlgorithm` on `CollectionBehaviorAttribute` (because these things are no longer strictly test collection-level concerns).
* Leave `DisableParallelization` on `CollectionDefinitionAttribute`. This acts as the test collection-level opt-out.
* Add `DisableParallelization` to a new `TestClassAttribute` (which can be a place to hang other test class-level configuration if such is needed in the future). This acts as the test class-level opt-out.
* Add `DisableParallelization` to `FactAttribute`. This acts as the test method-level opt-out.
* Add `ParallelMode` to `TestAssemblyConfiguration` (and obsolete `ParallelizeTestCollections`).
* Add `ParallelMode` to `ITestFrameworkExecutionOptions` (and obsolete `DisableParallelization`).
* Update `Xunit2Options` to accommodate the change to `ITestFrameworkExecutionOptions` (since v2 test assemblies will continue to rely on a boolean option named `DisableParallelization`); right now it's just a straight pass-through.
* Replace `-parallel` on console runner (with that is TBD).
* Obsolete `ParallelizeTestCollections` (and maybe `ParallelizeAssemblies`) on MSBuild runner, to be synced up with whatever we decide for `-parallel`'s replacement.

I'm sure this list probably isn't comprehensive yet, because the compiler and tests tend to tell me when I'm missing things in my designs. 😂 BTW, I'm not in love with the names `DefaultMode` and `ParallelMode` because "mode" is an overly generic term.

The design of the opt-out system and the parallelization implementation (including the potential `ValueTask` explosion) is still unclear to me. It's easy to achieve today because the parallelization happens in the assembly runner (parallelizing collections), and you can only opt out at the collection level. It separates the list, runs the parallel ones first, then sequentially runs the non-parallel ones.

In the new design, we need a way to opt out a collection, class, or method such that no tests in the opt out list--regardless of how it got there--is ever run in parallel against any other test. That means we can't use the same strategy that the assembly runner users; we literally need to lift everything that is opted out up to the assembly level, and wait until all the parallel things are done before we run the non-parallel things. Whatever this design is, it still needs to preserve the messaging structure (for example, if a class opts out, it's not just enough to find the list of tests; you need to ensure that the "class starting"/"class finished", "method starting"/"method finished", etc. messages still get sent at the appropriate time, and it needs to be communicated down the whole runner stack that "this thing in particular is a non-parallel" execution so that the runners down stream don't try to fan out in parallel). I honestly don't know what this has to look like yet from a design perspective. I'm still contemplating designs and will probably have to poke around with code to see what feels right. It may be that the runners become "bi-modal"; that is, we run them once in "run the parallel things in parallel" mode, and once in the "run the non-parallel things sequentially" mode.

As for the design for parallelization at every level, in my head I'm trying to come up with a design that seems less restrictive than "5 per level" but still ensures we don't deadlock ourselves. I'm not sure how fair the parallelism needs to be here, which would change the ways I might design it. I have some thoughts that aren't fully testing/formed yet here, too, so this is also still in contemplation mode for me. For sure, though, whatever design we come up with needs to change our thoughts on minimum thread pool size, because the existing calculation isn't is not gonna cut it. (And omg I just realized I used `Math.Min` when I meant `Math.Max`, which means we've been slow-starting everybody. 🤦🏼‍♂️)

So, bottom line is: I think a general usability design is done, but the implementation leaves me with a lot more questions to think about.

matthew-mabrey:
> This is named as Disable instead of Enable because I wanted them to be false by default (like any bool value would be). Whether this is meaningful distinction or not is a decision that is many many years in the past. 😂 I don't think there's anything wrong with the way this is named, though, so changing it feels unnecessary.

The `Disable` makes sense to me, I was just considering adding `Collections` to the name if we were adding a second property on the attribute called `EnableTestParallelization` (which would also default to false because we wouldn't be enabling full test parallelization within collections by default). No matter though since the approach you laid out sounds better to me!

> The design of the opt-out system and the parallelization implementation (including the potential ValueTask explosion) is still unclear to me. It's easy to achieve today because the parallelization happens in the assembly runner (parallelizing collections), and you can only opt out at the collection level. It separates the list, runs the parallel ones first, then sequentially runs the non-parallel ones.

Ah right, I didn't think about the core test runner changes needed now that the design is going from the flags set at the collection level to the opt-outs at every level. I'll try to brainstorm some options for this new approach too.

bradwilson:
I'm going to start heading down this implementation path and see what else shakes out.

bradwilson:
It took me quite a while to puzzle out a design that works. This is where I landed (finishing implementation and tests, but I'm so far it's working as expected, including seeing tests fail when I try to go to `All` mode, given that I have some tests which make non-parallelization assumptions).

The big issue is ensuring that tests that request to opt of parallelization never run in parallel against any other test that's allowed to run in parallel. To accomplish this, I ended up with:

* Three top-level parallelization modes, set via `[assembly: Parallelization(Mode = ...)]`, configuration file, or runner switches:
  * `ParallelMode.None` is equivalent to the old `[assembly: CollectionBehavior(DisableParallelization = true)]`
  * `ParallelMode.Collections` is equivalent to the default behavior
  * `ParallelMode.All` is new, and allows all tests to be run in parallel against any other test by default.
* Ability to opt out of parallelization at the following levels:
  * Assembly level, by setting `ParallelMode.None`
  * Collection level, by using `[CollectionDefinition(DisableParallelization = true)]`
  * Class level, by using `[TestClass(DisableParallelization = true)]`
  * Method level, by using `[Fact(DisableParallelization = true)]`
  * Test case level, only available in the object model (not expressed directly to the end user)
  * Test level, by using `DisableParallelization` on either the data attribute (all data rows from the data source) or via `ITheoryDataRow` (just the individual data row)

The implementation works by adding a new `ExecutionScheduler` class, which includes two primary APIs: `RunParallelTask` and `RunSequentialTask`. This uses a gate which allows either one or more parallel tasks to run in parallel or exactly one sequential task. There are three concrete implementations of the scheduler for the three parallel modes: unlimited (`-1` max threads), aggressive, and conservative.

The call to `RunParallelTask` only happens at two levels: for `ParallelMode.Collections` in the test assembly runner, it will use `RunParallelTask` for any collections which permit parallelization; and, for `ParallelMode.All` in the test runner, it will use `RunParallelTask` to run the test. These are the "leaves" of parallelism for the two parallel modes.

The call to use `RunSequentialTask` happens at every runner level, where it detects that the object has opted out of parallelism (for example, in the assembly runner, it enumerates over all the test collections, and will use `RunSequentialTask` for each test collection that's opted out of parallelism.

The parallel mode and execution scheduler are passed down through the runner context objects, and each context "lies" about the parallel mode (and claims `ParallelMode.None`) for any layer which has already opted out of the parallelization. It does this because it's already called `RunSequentialTask` at the point of opt-out, so there's no need for it to ever call `RunSequentialTask` or `RunParallelTask` again. Once you enter one gate or the other, you stay there, and don't ever try to re-enter your own gate; similarly, once you've opted out of parallelism, there's no way to opt back in at a lower layer.

This has an expected cascading impacts on the runners and the context objects: `Run` methods on the runners had to get extra parameters, as did the constructors for the context objects, to propagate through the parallel mode and execution scheduler.

bradwilson:
It definitely helps. 👍🏼

matthew-mabrey:
Ah nice, on its surface that looks good to me! Sorry I wasn't much help here. Glad you figured out a design that didn't require huge changes to the flow of the test pipeline (i.e. the bi-modal runner) since that seemed like a headache to implement :sweat_smile: 

> The implementation works by adding a new ExecutionScheduler class, which includes two primary APIs: RunParallelTask and RunSequentialTask. This uses a gate which allows either one or more parallel tasks to run in parallel or exactly one sequential task. There are three concrete implementations of the scheduler for the three parallel modes: unlimited (-1 max threads), aggressive, and conservative.

Nice, I like this implementation and API. After our last discussion I initially had a kind of similar idea of passing a custom semaphore implementation down the test stages that would essentially guarantee the same thing, but was worried about the potential Task explosion of all the sequential tests waiting their turn. Your implementation is better though since I assume we can queue a `Func` lamba instead of prematurely spawning another task for the sequential test stage.

I also think the task/lambda explosion probably won't be a problem for the vast majority of users since I assume most will be opting out their test collections/classes and won't have huge numbers of individual test methods/cases that are opted out and that we need to allocate something for in `RunSequentialTask`.

> The call to RunParallelTask only happens at two levels: for ParallelMode.Collections in the test assembly runner, it will use RunParallelTask for any collections which permit parallelization; and, for ParallelMode.All in the test runner, it will use RunParallelTask to run the test. These are the "leaves" of parallelism for the two parallel modes.

With this implementation, how does it run the tests of separate test classes and test methods in the same collection in parallel? Maybe I'm misunderstanding without the actual code, but from how I'm interpreting this it sounds like it's only branching off parallel tasks at the collection and test runner levels and the test class/method runners will still run things sequentially.

bradwilson:
> since I assume we can queue a `Func` lamba instead of prematurely spawning another task for the sequential test stage.

That is correct, the signatures are `Func`-based:

```csharp
public ValueTask<T> RunParallelTask<T>(
    Func<ValueTask<T>> taskFactory,
    CancellationToken cancellationToken)

public ValueTask<T> RunSequentialTask<T>(
    Func<ValueTask<T>> taskFactory,
    CancellationToken cancellationToken)
```

In both cases, you wait for the gate to open before invoking the factory.

> With this implementation, how does it run the tests of separate test classes and test methods in the same collection in parallel?

In `Collections` mode, the parallel task is naturally limited by the rules (max threads, etc.) and everything "below" that is done implicitly sequentially. This should be exactly how it behaves today.

In `All` mode, the parallel tasks "above" the test runner just use normal tasks launched in parallel, meaning they run on the default thread pool. This has the potential for "task explosion", but at least it is opt-in by the end user. It isn't until we get down to the test runner that the final parallel task goes into the limitation mechanism, so in reality what will have in task quantities that number as the product of test collections x test classes x test methods x test cases.

There is a bit of a counter-weight to the task explosion problem: right now I'm not taking any of the "in-between" parallel tasks through the "guaranteed background" code path that I use for `RunParallelTask`. A task doesn't become "background" until such time as it hits an `await` that would cause it to sleep, due to waiting on an incomplete task. In practice that means it's frequently going to be common that the number of outstanding started tasks will be much lower, since it'll traverse all the way down to the test runner level before tasks are run in "guaranteed background" mode, so many tasks that might otherwise be forced to wait _haven't even been started yet_. In practical terms, that makes performance much better than forcing the task explosion to happen early.

Another way to help somewhat mitigate the task explosion is to change the default collection behavior to "one per test assembly" when you're in `All` mode, since collections at that point don't act as parallelization barriers. I haven't decided if that's something I want to codify or not because it might be confusing (even if it's beneficial), and I'm not 100% sure how much people depend on the existing collection behavior from a reporting perspective.

matthew-mabrey:
Ah okay thanks for clarifying, that all seems reasonable to me :+1: 

I'm going to close out this PR for now since you've taken over, but let me know if there's anything I can help with or if you want another set of eyes to review your PR for your changes when they're ready.

bradwilson:
My plan is to add you as a co-author on the work since you got it started.

I was also going to open a PR (not something I normally do for my own contributions) to give you an opportunity to review and comment on it before I merge it.

matthew-mabrey:
Sounds good to me :+1:

bradwilson:
Also this: https://mastodon.social/@bradwilson/116886051993421566

bradwilson:
The new PR is live: https://github.com/xunit/xunit/pull/3598

Feel free to comment on it even after I merge it, since I don't want to rush any review you want to do.

PR #3596, opened 2026-06-30
#3593 Expose MTP SessionUid via TestContext.KeyValueStorage

AkbarDizaji:
Adds a well-known key (TestContext.KeyValueStorageMicrosoftTestingPlatformSessionUid) under which the Microsoft.Testing.Platform session ID is stored in TestContext.Current.KeyValueStorage when tests are run via MTP. This lets multiple test projects in a single run session correlate/share data. The value is absent when not running under MTP.

Before you submit a PR...

* Did you write unit tests?

### DIFF
--- a/src/common/MicrosoftTestingPlatform/TestPlatformTestFramework.cs
+++ b/src/common/MicrosoftTestingPlatform/TestPlatformTestFramework.cs
@@ -164,7 +164,7 @@ public class TestPlatformTestFramework :
 				throw new ArgumentException($"Unsupported discovery filter type '{filter.GetType().SafeName()}'", nameof(filter));
 
 			var messageHandler = new TestPlatformDiscoveryMessageSink(session.MessageHandler, projectAssembly.Assembly!.FullName!, sessionUid, discoveryFilter, messageBus, cancellationToken);
-			await projectRunner.Discover(projectAssembly, pipelineStartup, messageHandler);
+			await projectRunner.Discover(projectAssembly, pipelineStartup, messageHandler, testPlatformSessionUid: sessionUid.Value);
 		}, cancellationToken);
 	}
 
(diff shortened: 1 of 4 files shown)


AkbarDizaji:
@dotnet-policy-service agree

bradwilson:
Looking over this, I see two issues: one simple, and one complicated.

The simple issue is that this PR is targeting `main`, but the next release will come from `rel/4.0.0` which has changed quite significantly from `main`, so it will need to be re-targeted. That's something I can easily do myself.

The more complex issue, though, is that it ends up threading MTP knowledge into the core framework, which feels like a layering violation, since MTP is an optional component of the framework. Two examples off the top of my head are:

* Developers should not be able to try to get the session UID key if they aren't targeting MTP, which means putting the key as a const directly into `TestContext` is inappropriate. What would feel more appropriate would be an extension to `TestContext` that's only available when targeting MTP.

* The code in the core framework should not be aware of MTP, which means changing the signature of `TestContext.SetForInitialization` doesn't feel appropriate, either (nor pushing those changes into `ProjectAssemblyRunner`, which is similarly and purposefully MTP-unaware. That would mean pushing the initialization into `TestPlatformTestFramework` and the calls into `ProjectAssemblyRunner` such that the session UID gets set via the MTP assembly rather than via the core framework assembly.

I'm looking around there right now to see what the best options would be. A couple ideas I had that I want to try would include (a) updating the message handlers in `TestPlatformTestFramework` to just set the session UID the first time they see a message from the framework, which should be early enough; or (b) adding a test context initialization callback somewhere in the pipeline (maybe in `ProjectAssemblyRunner`) that would allow `TestPlatformTestFramework` to set the session UID before any work at all is done. It feels like the former would be simpler and have less impact, whereas the latter is probably a cleaner design (and allows other extensibility points to get access to the newly created test context).

bradwilson:
I opted for the second option, and committed the changes here with you listed as a co-author: acc9c38

Thanks!

PR #3598, opened 2026-07-12
Add support for full test parallelization (#1986, #2055)

bradwilson:
This is pretty large PR, but the core pieces are:

- Adding `ParallelMode` with values for `None`, `Collections`, and `All`. The default behavior is `Collections` for backward compatibility.
- Adding `[assembly: Parallelization]` to set parallel mode, max parallel threads, and the parallel algorithm (`[assembly: CollectionBehavior]` has obsoleted `DisableTestParallelization`, `MaxParallelThreads`, and `ParallelAlgorithm`).
- Configuration files and command lines have been updated to support the new parallel mode.
- Adding `[TestClass(DisableParallelization)]` to opt test classes out of parallelization.
- Adding `[Fact(DisableParallelization)]` (and friends) to opt test methods out of parallelization.
- Adding `[DataAttribute(DisableParallelization)]` to opt data sources of out of parallelization (inherited by `InlineDataAttribute`, `MemberDataAttribute`, and `ClassDataAttribute`).
- Adding `ITheoryDataRow.DisableParallelization` to opt a data row out of parallelization.
- Adding `ExecutionScheduler`, which adds support for parallel and sequential execution gates which are used when either hitting a point of parallelization (collections for `ParallelMode.Collections` or tests for `ParallelMode.All`) or a hitting a point of opting out parallelization.
- Implementation of the parallel algorithm (conservative, aggressive, or unlimited) is done in classes which derive from `ExecutionScheduler`.
- Updating `CoreTestXyzRunner` and associated context classes to plumb through both the current parallel mode and execution scheduler.
- Updated AOT registration to support code generators to express opt-out for test assemblies, test collections, test classes, and test methods, test cases, and tests.
- Updated built-in AOT generators to emit the new registration code.

The majority of the rest of the code is updating existing production code and tests to accommodate all the required changes to the runners and context classes.

/cc @matthew-mabrey


src/xunit.v3.core.aot/Configuration/RegisteredEngineConfig_aot.cs
```diff
@@ -283,6 +283,24 @@ public static void RegisterAssemblyFixtureFactory(
 		Func<bool, ValueTask<object?>> factory) =>
 			assemblyRegistration.AssemblyFixtureFactories[type] = (_, forceCreation) => factory(forceCreation);
 
+	/// <summary>
+	/// Registers that a test assembly's parallelization algorithm.
```
matthew-mabrey:
```suggestion
	/// Registers a test assembly's parallelization algorithm.
```

src/xunit.v3.core.aot/Configuration/RegisteredEngineConfig_aot.cs
```diff
@@ -283,6 +283,24 @@ public static void RegisterAssemblyFixtureFactory(
...
+	/// </summary>
+	public static void RegisterAssemblyParallelAlgorithm(ParallelAlgorithm algorithm) =>
+		assemblyRegistration.ParallelAlgorithm = algorithm;
+
+	/// <summary>
+	/// Registers that a test assembly's parallelization mode.
```
matthew-mabrey:
```suggestion
	/// Registers a test assembly's parallelization mode.
```

src/xunit.v3.core.aot/Configuration/RegisteredEngineConfig_aot.cs
```diff
@@ -283,6 +283,24 @@ public static void RegisterAssemblyFixtureFactory(
...
+	/// </summary>
+	public static void RegisterAssemblyParallelMaxThreads(int maxParallelThrads) =>
+		assemblyRegistration.ParallelMaxThreads = maxParallelThrads;
+
+	/// <summary>
+	/// Registers that a test assembly's parallelization mode.
```
matthew-mabrey:
```suggestion
	/// Registers a test assembly's parallelization mode.
```

src/xunit.v3.core/Framework/ExecutionScheduler.cs
```diff
@@ -0,0 +1,332 @@
...
+		new _Aggressive(maxParallelThreads > 0 ? maxParallelThreads : Environment.ProcessorCount);
+
+	/// <summary>
+	/// Create an execution scheduler for the conversative parallelism algorithm.
+	/// </summary>
+	/// <param name="maxParallelThreads">The maximum number of permitted parallel tasks</param>
+	/// <remarks>
+	/// Passing <c>0</c> for <paramref name="maxParallelThreads"/> will use <see cref="Environment.ProcessorCount"/>.
+	/// </remarks>
+	public static ExecutionScheduler CreateConversative(int maxParallelThreads) =>
```
matthew-mabrey:
```suggestion
	public static ExecutionScheduler CreateConservative(int maxParallelThreads) =>
```

src/xunit.v3.core/Framework/ExecutionScheduler.cs
```diff
@@ -0,0 +1,332 @@
...
+	/// <para>Passing <c>0</c> for <paramref name="maxParallelThreads"/> will use <see cref="Environment.ProcessorCount"/>.</para>
+	/// </remarks>
+	public static ExecutionScheduler Create(
+		int maxParallelThreads,
+		ParallelAlgorithm parallelAlgorithm) =>
+			(maxParallelThreads, parallelAlgorithm) switch
+			{
+				( < 0, _) => CreateUnlimited(),
+				(_, ParallelAlgorithm.Aggressive) => CreateAggressive(maxParallelThreads),
+				(_, ParallelAlgorithm.Conservative) => CreateConversative(maxParallelThreads),
```
matthew-mabrey:
```suggestion
				(_, ParallelAlgorithm.Conservative) => CreateConservative(maxParallelThreads),
```

src/xunit.v3.core/Framework/ExecutionScheduler.cs
```diff
@@ -0,0 +1,332 @@
...
+		public void Dispose()
+		{
+			int newCount;
+
+			lock (scheduler.gate)
+				newCount = --scheduler.sequentialCount;
+
+			if (newCount == 0)
+			{
+				scheduler.sequentialThreadId = null;
```
matthew-mabrey: Should `sequentialThreadId = null` here be under the scheduler gate lock as well? Doesn't seem super likely there would be a race condition where this clears the thread ID after `EnterSequentialGate` sets it, but the other uses of it are under the gate lock, so maybe still potentially possible?

src/xunit.v3.core.tests/Framework/ExecutionSchedulerTests.cs
```diff
@@ -0,0 +1,199 @@
...
+			Assert.StartsWith("Start ", startMessage);
+
+			var index = int.Parse(startMessage.Substring(6));
+			var finishMessage = messages[i + 1];
+			Assert.Equal($"Finish {index}", finishMessage);
+		}
+
+		async void worker(object? index)
+		{
+			var random = new Random();
```
matthew-mabrey: Was this left in on accident?

src/xunit.v3.core/Framework/ExecutionScheduler.cs
```diff
@@ -0,0 +1,332 @@
...
+	}
+
+	async ValueTask<IDisposable> EnterSequentialGate(CancellationToken cancellationToken)
+	{
+		while (true)
+		{
+			cancellationToken.ThrowIfCancellationRequested();
+
+			lock (gate)
+				if ((parallelCount == 0 && sequentialCount == 0) || sequentialThreadId == Environment.CurrentManagedThreadId)
```
matthew-mabrey, line 129: I could be missing something but it seems like there isn't anything stopping sequential tests from getting queued and run first (depending on ordering). For example, if the first collection in the assembly is full of tests with ParallelizationDisabled set only at the test class level.

Is that something we care about preventing? I'd assume we want all parallel tasks running first and then run the sequential. Guessing for the vast majority of cases the test pipeline will reach a parallel test first and it will be a non-issue though.

matthew-mabrey (commented):
Just a couple very minor things, but overall changes look good to me 👍

src/xunit.v3.core.tests/Framework/ExecutionSchedulerTests.cs
bradwilson: Oops. Leftover from an older version of this test which had random delays.

src/xunit.v3.core/Framework/ExecutionScheduler.cs
bradwilson: That's correct, this is entirely possible. It's not only possible for sequential to run first, but also possible to interleave sequential and parallel, depending on when things arrive "at the gate".

I'm sure some wording exists somewhere in the docs about ordering (parallel first) that will need to get updated, and the change in behavior noted in the 4.0 release notes.

src/xunit.v3.core/Framework/ExecutionScheduler.cs
bradwilson: Yep, you're right. 👍🏼

matthew-mabrey (approved):

PR #3611, opened 2026-08-19
Disable colors for the TeamCity reporter

### LINKED ISSUE #1438: TeamCity output contains `ResetColorANSI()` characters
Just upgraded to xUnit 2.3-beta-4 to 2.3-beta-5 and notice the new console colour changes are bleeding into TeamCity, that is my suspicion looking at https://github.com/xunit/xunit/compare/2.3-beta4...2.3-beta5 anyway.

image

May possibly follow this up with PR if somebody else doesn't beat me to it :+1:

HengYpinn:
Fixes #1438.

TeamCity reads stdout as service messages, so ANSI color and reset sequences just turn into noise in the build log. We already know when the TeamCity reporter has been selected; this makes no-color the default at that point for the current v3 runners.

This only fills the value when it hasn't been configured, so an explicit `NoColor = false` still wins.

I added the same regression case to both command-line implementations and ran their complete test assemblies locally:

- in-process console: 248 passed
- standalone console: 278 passed

`dotnet format whitespace --verify-no-changes` and `git diff --check` are also clean.

### DIFF
--- a/src/xunit.v3.runner.common/Parsers/CommandLineParserBase.cs
+++ b/src/xunit.v3.runner.common/Parsers/CommandLineParserBase.cs
@@ -938,6 +938,9 @@ public abstract class CommandLineParserBase
 				RunnerReporters.FirstOrDefault(r => "default".Equals(r.RunnerSwitch, StringComparison.OrdinalIgnoreCase))
 					?? new DefaultRunnerReporter();
 
+		if (Project.Configuration.NoColor is null && Project.RunnerReporter is TeamCityReporter)
+			Project.Configuration.NoColor = true;
+
 		return Project;
 	}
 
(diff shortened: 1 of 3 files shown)


HengYpinn:
I noticed #1438 was originally reported against v2, while this PR updates the current v3 runners. [TeamCity’s current documentation says ANSI coloring is supported by default](https://www.jetbrains.com/help/teamcity/build-log.html#ANSI-style+Coloring+in+Build+Log). Would you still prefer the TeamCity reporter to disable colors automatically, or should explicit -useAnsiColor preserve them? Happy to adjust the implementation either way.

bradwilson:
Given that TeamCity supports color now, this is probably unnecessary. Users can still disable color either via a command line option (`-noColor`) or an environment variable (`NO_COLOR=1`).

HengYpinn:
Thanks for confirming, I will close this PR

PR #3627, opened 2026-09-14
AOT generator: stop dropping inherited test methods

meziantou:
## Problem

Under Native AOT, `TestClassGenerator` loses some inherited test methods declared in source without reporting anything, and the run is green. Reflection-based discovery runs all of them. Shared abstract contract test suites are a common xUnit pattern, so this hides real tests.

The generator walked base classes through syntax (`TestClassGenerator.ProcessTestClass`). That failed in two cases where every class is in source:

1. **Qualified base type.** In `class FooTests : MyApp.Testing.ContractTests` (or `global::…`), the base is not a `SimpleNameSyntax`, so the generator returned early.
2. **3+ level chain across files.** The recursion called `semanticModel.GetSymbolInfo` on a node from another syntax tree. Roslyn threw, and `catch { return; }` hid the exception. The existing `CrossFileInheritance` test only covers 2 levels.

## Fix

- The generator binds the first base-list type from the class's own declaration, whatever form it is written in. That type is always in the same syntax tree, so this is safe.
- It then walks `INamedTypeSymbol.BaseType` instead of recursing through syntax, so base classes in other files work at any depth. The `try/catch` is removed.
- Partial classes are still gated the same way: base classes are processed only from the partial declaration that lists the base.
- **Only test methods declared in source are supported**, as discussed in review. The walk stops at the first base type from a referenced assembly, whose test methods are not registered. This is unchanged from `main`. Reporting that case as an error belongs in an analyzer in xunit/xunit.analyzers, which this PR does not include.

There are no public API changes. An earlier iteration made `MethodDeclarationSyntax` nullable and registered theory data for methods from referenced assemblies. Both were reverted after review, so the PR's net diff touches only `TestClassGenerator` and tests.

## Tests

`CoreGeneratorTest` gains a `GenerateSourcesWithReferencedAssembly` helper, which compiles a separate in-memory assembly and references it.

New tests in `TestClassGeneratorTests`:

- `CrossFileInheritance_MultipleLevels`: `C : B : A` across three files.
- `QualifiedBaseClassName`: `MyApp.Testing.ContractTests` and `global::MyApp.Testing.ContractTests`.
- `BaseClassInReferencedAssembly_InheritedTestMethodsAreNotRegistered`: the class's own source test is registered, and the test method inherited from the referenced assembly is not.

With `main`'s `TestClassGenerator.cs` and the new tests, the 3 source-only cases fail and the referenced-assembly test passes, since `main` doesn't register those methods either. With this PR, all 69 `xunit.v3.core.aot.generators.tests` pass, and the build has no warnings.


bradwilson:
> Methods declared in a referenced assembly have no `MethodDeclarationSyntax`

I am comfortable with the notion that AOT _only ever supports test methods from source code_ and that it should reject test methods from _binary references_. (Silently reject isn't great; an error from an analyzer should be issued, since generating errors from source generators is strongly discouraged.)

bradwilson:
Thanks!

bradwilson:
Available in v3 `4.0.2-pre.5` https://xunit.net/docs/using-ci-builds

PR #3631, opened 2026-09-14
Clamp maxParallelThreads multiplier to at least 1 thread and reject a zero multiplier

### LINKED ISSUE #3628: A `maxParallelThreads` multiplier that rounds down to 0 runs tests on all cores
`maxParallelThreads` accepts a multiplier such as `0.5x`, which is multiplied by `Environment.ProcessorCount`. The result is cast to `int`, which rounds down. If the result is `0`, xUnit treats it as "default" and uses one thread per core. So a small multiplier gives the opposite of what the user asked for: `0.2x` on a 4 cores machine => `(int)(0.2 * 4)` = `0`.

I think the fix is mostly adding `Math.Max(1, (int)(maxThreadMultiplier * Environment.ProcessorCount))` everywhere.

I'm not sure if `0x` should be invalid, be similar to `0`, or 1 thread (`Math.Max(1, multiplier)` would clamp it to 1).

meziantou:
Fixes #3628

A `maxParallelThreads` multiplier is multiplied by `Environment.ProcessorCount` and cast to `int`, which truncates. When the result was `0` (for example `0.2x` on a 4-core machine), it hit the "default" sentinel and ran tests on one thread per core, which is the opposite of what was asked for.

## Changes

Following the guidance in https://github.com/xunit/xunit/issues/3628#issuecomment-5667494952, every place that parses the multiplier now:

- Clamps a non-zero multiplier to at least one thread: `Math.Max(1, (int)(multiplier * Environment.ProcessorCount))`
- Treats a zero multiplier (`0x`, `0.0x`) as invalid, using each location's existing handling for invalid values:
  - `CommandLineOptionsProvider` (MTP `--max-threads`) and `CommandLineParserBase` (`-maxThreads`): throws the existing `ArgumentException`
  - MSBuild runner (`xunit.cs`): logs the existing "value is invalid" error and fails the task
  - `TestConfig` (MTP configuration) and `ConfigReader_Json` (`xunit.runner.json`): ignored, like other invalid values

## Tests

Added coverage for zero multipliers being rejected/ignored and for `0.0001x` resolving to `1`, in:

- `xunit.v3.mtp-v2.tests` (`CommandLineOptionsProviderTests`, `TestConfigTests`)
- `xunit.v3.runner.utility.tests` (`ConfigReader_JsonTests`)
- `xunit.v3.runner.console.tests` and `xunit.v3.runner.inproc.console.tests` (`CommandLineTests`)

Ran the affected tests locally on net8.0 (mtp-v2, runner.utility, runner.inproc.console): all pass. `xunit.v3.runner.console.tests` targets net472 only and wasn't run locally (macOS); it exercises the same `CommandLineParserBase` code as the inproc tests. The MSBuild runner has no unit tests, so that change is covered by compilation only.


src/xunit.v3.mtp-v2.tests/CommandLineOptionsProviderTests.cs
```diff
@@ -148,6 +148,8 @@ public class ArgumentSwitches : CommandLineOptionsProviderTests
 		[InlineData("Invalid value 'abc' (must be one of: 'off', 'on', 'only')", "explicit", new[] { "abc" })]
 		[InlineData("Invalid value 'abc' (must be an integer between 0 and 2147483647)", "long-running", new[] { "abc" })]
 		[InlineData("Invalid value 'abc' (must be one of: 'default', 'unlimited', a positive number, a multiplier in the form of '0.0x')", "max-threads", new[] { "abc" })]
+		[InlineData("Invalid value '0x' (must be one of: 'default', 'unlimited', a positive number, a multiplier in the form of '0.0x')", "max-threads", new[] { "0x" })]
```
bradwilson: The message here is a little strange, because the example (`0.0x`) is explicitly not permittable. I wonder if switching the example to `1.5x` or something similar makes more sense. 🤔

meziantou: Switched the example to `1.5x` in all three invalid-value messages (MTP `--max-threads`, console `-maxThreads`, and the MSBuild task), and updated the tests to match.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

bradwilson:
Thanks!

bradwilson:
Fix available in v3 `4.0.2-pre.4` https://xunit.net/docs/using-ci-builds

PR #3633, opened 2026-09-19
#3630: Wake all scheduler gate waiters so parallel tests don't serialize behind a sequential test

### LINKED ISSUE #3630: Parallelization is suboptimal in some cases where sequential tests run before parallel tests
In the following example, I would expect the execution time to be less than 4s: 
- 1s for the sequential test
- 2s for all parallel tests + 0.25s for the fixture

However, it runs for about 17s.

```c#
using Xunit.Sdk;
using Xunit.v3;

[assembly: Parallelization(Mode = ParallelMode.All)]

public class SequentialTests
{
	[Fact(DisableParallelization = true)]
	public async Task SequentialTest1()
	{
		await Task.Delay(1000, TestContext.Current.CancellationToken);
	}
}

public class ParallelTests(DelayFixture _) : IClassFixture<DelayFixture>
{
	[Fact] public async Task Test1() => await Task.Delay(2000, TestContext.Current.CancellationToken);
	[Fact] public async Task Test2() => await Task.Delay(2000, TestContext.Current.CancellationToken);
	[Fact] public async Task Test3() => await Task.Delay(2000, TestContext.Current.CancellationToken);
	[Fact] public async Task Test4() => await Task.Delay(2000, TestContext.Current.CancellationToken);
	[Fact] public async Task Test5() => await Task.Delay(2000, TestContext.Current.CancellationToken);
	[Fact] public async Task Test6() => await Task.Delay(2000, TestContext.Current.CancellationToken);
	[Fact] public async Task Test7() => await Task.Delay(2000, TestContext.Current.CancellationToken);
	[Fact] public async Task Test8() => await Task.Delay(2000, TestContext.Current.CancellationToken);
}

// Ensure parallel tests run after the sequential one
public sealed class DelayFixture : IAsyncLifetime
{
	public ValueTask InitializeAsync() => new(Task.Delay(250));
	public ValueTask DisposeAsync() => default;
}
```

meziantou:
Fixes #3630.

## Problem

With `ParallelMode.All`, parallel tests that queue up behind a `DisableParallelization` test end up running one at a time instead of together. The repro in #3630 takes ~17s instead of the expected ~4s (1s sequential + 8 x 2s parallel).

## Cause

`ExecutionScheduler`'s gate used an `AsyncAutoResetEvent`, whose `Set()` releases exactly one waiter. The release path only signals on the edge where the holder count drops to zero:

```csharp
lock (scheduler.gate)
    newCount = --scheduler.sequentialCount;

if (newCount == 0)
    scheduler.gate.Set();   // wakes ONE waiter
```

So when the sequential test finishes, exactly one of the queued parallel tests is woken. As soon as it enters, `parallelCount > 0` makes every remaining waiter satisfy the parallel predicate (`parallelCount > 0 || sequentialCount == 0`) — the state is already favorable for all of them, but nothing wakes them to notice. They stay parked until `parallelCount` falls back to zero, which releases one more, and so on.

Relatedly, because a wake-up is a consumable token, a waiter that woke and still could not enter (a sequential waiter seeing `parallelCount > 0`) swallowed a signal other waiters needed.

## Fix

The gate now broadcasts to every waiter of the current generation rather than handing a token to one:

* A waiter that cannot enter captures the current generation's task **under the same lock that guards the counters**, so no wake-up can be lost between the predicate check and the wait.
* A release swaps in a fresh generation and completes the old one after leaving the lock, waking everyone parked on it.
* The signal means only "state changed, re-check". Entry is still decided by the same predicate under the same lock, so mutual exclusion between parallel and sequential work is unchanged — e.g. if a sequential waiter wins the race after a sequential release, the parallel waiters re-check, see `sequentialCount != 0`, and park on the new generation.

The completion runs outside the lock and the generation uses `RunContinuationsAsynchronously`, so the releasing thread doesn't synchronously resume every woken waiter — each of those continuations immediately reaches for the same lock, and under `MaxConcurrencySyncContext` inline resumption would hijack the releasing thread.

Dropping the `initialState: true` of the old event loses nothing: the first entrant finds all counts at zero and never waits, so that initial token only ever bought a spurious extra spin.

`AsyncAutoResetEvent` in `xunit.v3.common` now has no callers, but it's a public type so I left it in place.


bradwilson:
Thanks! This was a good bit of detective work, and I saw this just as I was sitting down to figure it out myself. 👍🏼 

-----

As an aside, I saw something using `-reporter verbose` that was slightly worrying with your original test:

> 

I added `Console.WriteLine` statements to ensure the tests weren't actually starting when the verbose reporter said they were starting:

```csharp
using System;
using System.Threading.Tasks;
using Xunit;
using Xunit.Sdk;
using Xunit.v3;

[assembly: Parallelization(Mode = ParallelMode.All)]

public class SequentialTests
{
    [Fact(DisableParallelization = true)]
    public async Task SequentialTest1()
    {
        Console.WriteLine("Sequential - Start");
        await Task.Delay(1000, TestContext.Current.CancellationToken);
        Console.WriteLine("Sequential - Finish");
    }
}

public class ParallelTests : IClassFixture<DelayFixture>
{
    [Fact]
    public async Task Test1()
    {
        Console.WriteLine("Parallel1 - Start");
        await Task.Delay(2000, TestContext.Current.CancellationToken);
        Console.WriteLine("Parallel1 - Finish");
    }
    [Fact]
    public async Task Test2()
    {
        Console.WriteLine("Parallel2 - Start");
        await Task.Delay(2000, TestContext.Current.CancellationToken);
        Console.WriteLine("Parallel2 - Finish");
    }
    [Fact]
    public async Task Test3()
    {
        Console.WriteLine("Parallel3 - Start");
        await Task.Delay(2000, TestContext.Current.CancellationToken);
        Console.WriteLine("Parallel3 - Finish");
    }
    [Fact]
    public async Task Test4()
    {
        Console.WriteLine("Parallel4 - Start");
        await Task.Delay(2000, TestContext.Current.CancellationToken);
        Console.WriteLine("Parallel4 - Finish");
    }
    [Fact]
    public async Task Test5()
    {
        Console.WriteLine("Parallel5 - Start");
        await Task.Delay(2000, TestContext.Current.CancellationToken);
        Console.WriteLine("Parallel5 - Finish");
    }
    [Fact]
    public async Task Test6()
    {
        Console.WriteLine("Parallel6 - Start");
        await Task.Delay(2000, TestContext.Current.CancellationToken);
        Console.WriteLine("Parallel6 - Finish");
    }
    [Fact]
    public async Task Test7()
    {
        Console.WriteLine("Parallel7 - Start");
        await Task.Delay(2000, TestContext.Current.CancellationToken);
        Console.WriteLine("Parallel7 - Finish");
    }
    [Fact]
    public async Task Test8()
    {
        Console.WriteLine("Parallel8 - Start");
        await Task.Delay(2000, TestContext.Current.CancellationToken);
        Console.WriteLine("Parallel8 - Finish");
    }
}

// Ensure parallel tests run after the sequential one
public sealed class DelayFixture : IAsyncLifetime
{
    public ValueTask InitializeAsync() => new(Task.Delay(250));
    public ValueTask DisposeAsync() => default;
}
```

> 

Whew! That's a good thing, something I need to file away.

bradwilson:
Available in v3 `4.0.2-pre.7` https://xunit.net/docs/using-ci-builds

PR #3635, opened 2026-09-21
Fix ANSI escape sequences in TRX output

### LINKED ISSUE #3634: Generated TRX file is broken when captured stdout contains some escape sequences
When captured stdout contains some escape sequences.
Empty TRX file is generated **silently**. (or wrong **partial** TRX file is generated on some cases)

**Reproduce Code**

```csharp
[assembly: CaptureConsole]

namespace Project13;

public class UnitTest1
{
    [Fact]
    public void Test1()
    {
        string underlinedText = GetUnderlinedText("Underlined Text");
        Console.WriteLine(underlinedText);
    }

    private const string ESC = "\x1B";

    private static string GetUnderlinedText(string text)
        => $"{ESC}[4m{text}{ESC}[0m";
}
```

**Reproduce Steps**
1. Run program with `trx` generate options
    > dotnet run --report-xunit-trx --output Detailed
2. Confirm generated empty trx file is generated.

Note: I've tested on Windows environment.

---
This issue is occurred when using `xunit.v3` 4.x versions.
and it's not occurred when using `xunit.v3.mtp-v2` 3.2.2.

#### Other Issue
When generating TRX file on Windows environment.
Stdout messages are wrapped with `<Message>` tag per line.

On Linux/macos environment.
Entire stdout messages are wrapped with single `<Message>` tag.

Normally It's not affects trx parsing behaviors.
Though **wrong partial TRX file** generation issue is observed on Linux/macos environment only.

So it might be better to adjust behaviors by adding `LF` line splitting.
https://github.com/xunit/xunit/blob/c327428e7c75ee6999a0c0398b7116f6fcf9b65b/src/xunit.v3.runner.common/Results/Builtin/TrxResultWriterMesageHandler.cs#L329-L330

sb123sb123:
Fixes #3634

## Cause
Captured stdout is written directly into TRX XML Message elements. ANSI SGR sequences such as ESC[4m and ESC[0m contain the XML 1.0-invalid ESC character (0x1B), so saving or parsing the report can fail and leave an empty or partial TRX file.

## Changes
- Reuse the existing AnsiUtility.RemoveAnsiEscapeCodes helper before writing captured output into TRX Message elements.
- Add a regression test covering the underline sequence from the issue reproduction and asserting that the resulting TRX XML contains the plain text.

## Tests
- dotnet build src/xunit.v3.runner.common.tests/xunit.v3.runner.common.tests.csproj -c Release -f net8.0 --no-restore --nologo -m:1 -v:minimal
- dotnet test --project src/xunit.v3.runner.common.tests --configuration Release --framework net8.0 --no-build --no-launch-profile --explicit on --no-progress --output Normal --max-parallel-test-modules 1 (709 passed)
- Focused direct xUnit runner test for TrxResultWriterMessageHandlerTests.TestResult_OutputRemovesAnsiEscapeCodes (passed)
- dotnet format whitespace --verify-no-changes on both changed files and git diff --check

## Limitations
Validation covered the net8.0 test target on the remote Windows environment. The repository full Build target compiled many outputs but later stopped when the G drive ran out of space while generating unrelated NetFx/x86 artifacts; the repository build wrapper also requires msbuild.exe, which is not installed in this environment.

## AI assistance
AI assistance was used to inspect the issue and repository, analyze the cause, draft the focused implementation and regression test, and review the final diff. The final behavior, scope, tests, and disclosure were reviewed before submission.


bradwilson:
There are two issues:

- It's not clear to me that stripping ANSI is the right answer, rather than escaping it.
- This project has a [strict no-AI policy](https://xunit.net/code-of-conduct#strict-no-llm--no-ai-policy) which invalidates this PR. You may re-open a new PR which has been written entirely by a human without any LLM/AI assistance.

bradwilson:
Closing as I have a more comprehensive fix available for all reports.

